PR #116, opened 2013-08-17
Implemented iOS web app support

corihudson:
I've implemented iOS web app support for Snap! (essentially allowing Snap! to function as an iOS app without having to go through the App Store). You can try it out at [http://files.technoboy10.tk/Snap!] .


tjvr:
Can we please implement this?

CosmicWebServices:
I agree

PR #298, opened 2014-01-10
add french help translation

gagalago:
Add the possibility to translate help. And apply them on French. If you want other translations, you can see to http://info.scratch.mit.edu/Languages


hasso:
What's the status of this? I'm looking at using Snap! with my students, but translated help is absolutely must be for them (they are 10-12 years old). Among other stoppers this is most critical for me.

PR #362, opened 2014-03-23
Initial Version of Text Stamping

cycomachead:
Updated:

Text Stamping for _Snap!_:
Here's what we have that's new:
- Sprite-scoped font properties (these are defaults):
  
  ``` javascript
  {
  // CSS Style Font Options (which work for Canvas)
  'font size' : '12', // measured in 'px'
  'font family' : 'Helvetica',
  'font variant' : 'normal',
  'font weight' : 'normal',
  'font style' : 'normal',
  // 'line-height' : '1', // currently not exposed
  // Canvas Font Options
  'text align' : 'left',
  'text baseline' : 'alphabetic',
  // Custom Snap! Options
  'move with text' : false,
  'rotate with sprite' : false
  } 
  ```
- `stamp text` Block
  - 1 input, which always uses the current properties
- `pixel width of (text)` block
  - reports the width in px (Height can be obtained via font size)
- `set text option (option) to (value)` block
- the attribute of block now reports the font settings

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -6205,7 +6205,6 @@ InspectorMorph.prototype.updateCurrentSelection = function () {
 
     if (isNil(sel)) {return; }
     val = this.target[sel];
-    if (this.currentProperty === val) {return; }
     this.currentProperty = val;
     if (isNil(val)) {
         txt = 'NULL';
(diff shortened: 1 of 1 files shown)


brianharvey:
Thanks!

I've attached the two forms you sent me, and my CV.

You didn't send me a biobib form; if you do that, I'll fill it out and 
return it.

ddgarcia:
what’s this about?
dan
--
Dan Garcia, Ph.D.   777 Soda Hall, 510.517.4041   ddgarcia@cs.berkeley.edu
Senior Lecturer SOE    UC Berkeley, EECS dept    cs.berkeley.edu/~ddgarcia

PR #403, opened 2014-04-26
New Feature: Search Functionality

khotchkiss13:
I have added search functionality to snap, which allows for users to search for blocks by name. I made a input string morph and a push button morph that are located under the categories amd above the palette. It works for all standard, and custom blocks, as well as variables, push buttons, and toggles. Search works in all supported languages.  There are two issues I still have, which are that the toggles do not refresh after they are clicked, and the search button's text is not initially centered in the button.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -2158,7 +2158,7 @@ BlockMorph.prototype.hidePrimitive = function () {
     if (cat === 'lists') {cat = 'variables'; }
     ide.flushBlocksCache(cat);
     if (ide.currentCategory === 'search') {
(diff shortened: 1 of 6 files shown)


bromagosa:
Just sorting through old PRs...

Shouldn't this one be closed? We can already search in Snap<i>!</i> since a while ago...

jmoenig:
So, what we *do* want, I guess, is being able to search for a (partial) string in the project (blocks, code)...

bromagosa:
Yep, that's a separate issue though :)

PR #452, opened 2014-05-25
separating local and global vars

### LINKED ISSUE #189: Separate Global vars from Sprite Vars
In BYOB all Sprite variables used to be below a little line, so you could tell they affected just that sprite. Snap doesn't give any way of differentiating between globals and sprite variables.

marwahaha:
resolves #189


cycomachead:
Closing because this was done in 4.1.1

PR #541, opened 2014-07-30
Translate export script pic with result bubble

MMSequeira:


bromagosa:
Seems like this is already in the PT locale since a while ago (https://github.com/jmoenig/Snap/blob/master/locale/lang-pt.js#L2052).

Closing this :)

Issue #460, opened 2014-05-29
Camera + Mic Inputs for Costumes and Sounds


cycomachead:
This has been discussed before, but never filed as an issue. Maybe someone will see this and want to take on the project. :)
- You should be able to import a costume by taking a photo with a web cam
- Likewise, you should be able to use a mic to record a sound for a sprite.

bromagosa:
Cool, this is done! :)

PR #562, opened 2014-08-11
add camera blocks (partial fix for #460)

Gubolin:
The stage gets two new pen blocks and a predicate, the sprites two predicates and a reporter.
The camera picture is painted as pen trail, that means that the sprites are able to recognize colors with the sensing blocks but also that trails / stamps will be overwritten.


Gubolin:
Closing old pull requests with merge conflicts, as I do not have the time to main them.

PR #565, opened 2014-08-18
add location blocks

Gubolin:
* new reporter: language
* new reporter: location (all, country, state, district, suburb, city, road, house number, licence) (not always exact, uses Open Street Map and the geolocation API)

With these blocks it is possible to have multi-language Snap! projects. Also, you could create a paperchase project, implement StreetView, ... .

Issue #81, opened 2013-04-29
Virtual keyboard doesn't work on Android


jmoenig:

ToonTalk:
I discovered today that [this patch](https://gist.github.com/swiperthefox/48996720c0e82b3d02560188a4b5006f) for enabling IME (e.g. Chinese input) also fixed the problem with input in Snap! on Android. See pull #2322 for details.

jmoenig:
Thanks, @ToonTalk , that's a very valuable argument. I'll look into this issue next week when 5.0 is out. Supporting all languages for input is really important.

jmoenig:
the issue of entering text into input fields on Android has now been resolved as of https://github.com/jmoenig/Snap/tree/v5.0.6
Thanks, everybody, who was involved in addressing this issue!

PR #544, opened 2014-07-31
add volume blocks (fix #405)

Gubolin:
Added volume control blocks

Please test and review it.

PR #575, opened 2014-09-06
Distribute offline versions of Snap! for desktop and mobile

Gubolin:
*Work in progress*
This commit adds the configuration files for `cordova` and `node-webkit` so that Snap! can be packaged into a mobile or desktop application.
In addition to this it is possible to add blocks that have access to the operating system, so file system access or bluetooth, vibration, notifications, ... can be added with cordova plugins and nodejs.

*edit* Changes:
* fixed Android keyboard (#81)
* added "buildscripts"
* added vibrate block (sensing)
* added compass block (sensing)
* added acceleration blocks (sensing)
* added notification blocks (looks)

*edit* Todo:
* NFC - cannot test, cancelled. See (http://plugins.cordova.io/#/package/com.chariotsolutions.nfc.plugin)
* file blocks (sensing) (need testing)
* added presentation-only mode (removed)
* added location blocks (merged #565) (sensing) (removed)


desktop.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Requirements:
+# git, zip, nodejs
+# node-webkit (https://github.com/rogerwang/node-webkit), node-webkit-builder (https://github.com/mllrsohn/node-webkit-builder)
+
+builddir=$(mktemp -d)
+
+if [[ $2 == "" ]]
+then
+    git clone https://github.com/Gubolin/snap.git $builddir
```
bb010g: Shouldn't this be changed to the official Snap repo?

mobile.sh
```diff
@@ -0,0 +1,64 @@
...
+
+builddir=$(mktemp -d)
+
+cordova create $builddir edu.berkeley.snap "Snap\!"
+cd $builddir
+rm -rf www config.xml
+
+if [[ $2 == "" ]]
+then
+    git clone https://github.com/Gubolin/snap.git www
```
bb010g: Here also.

binary.sh
```diff
@@ -0,0 +1,91 @@
...
+   if [ -f "$3" ]
+    then
+        content=$(cat $3)
+    else
+        ide=true
+    fi
+fi
+
+buildsource=$(mktemp -d)
+git clone "https://github.com/Gubolin/snap.git" $buildsource
```
bb010g: Repo address?

cycomachead:
Would it be possible to have a single repo variable for all scripts?

Or, since these files are already in the repo, shouldn't it be possible to just use the existing checked out source -- that way anyone who clones can easily build their own versions, and we don't need to worry about unnecessary commits to files in the repo.

bb010g:
Repo constant would be nice (if we need it), esp. if we change from the (IMHO) clunky Snap--Build-Your-Own-Blocks name. They look to be needed here, as this is meant to be usable without a source provided already (lighter weight for distribution).

Gubolin:
> Shouldn't this be changed to the official Snap repo?

In theory, yes. Practically, no, because the mandatory config files are not in the official Snap! repository yet.

> Would it be possible to have a single repo variable for all scripts?

Ok, done. You can set `snapsource` now. Don't forget the configuration files in the repository for cordova and node-webkit, or the build will fail.

> just use the existing checked out source

This was my first approach. You can use it if you call `mobile.sh` or `desktop.sh` directly and give the optional `BUILDSOURCE`. This functionality is not included in the main part because it cluttered my working directory and I think that it's better to have a 'clean' build from upstream.

nemilya:
@gubolin thanks for mobile version! It was my dream)

What do you think about using Vagrant (https://www.vagrantup.com) for fast setup VM for Cordova/android builds?

For example VM 'snap-build-box' - pre configured Ubuntu with all cordova/nodejs/android specific libs. Like this - https://github.com/driftyco/ionic-box Vagrant config for ionic-mobile framework - include android/nodejs/cordova libs. Works great on my win32 machine - it builds apk for 15 seconds.

Steps:
1. configure libs for android, cordova etc
2. configure mobile Snap version
3. Set programm for Snap
4. Build android apk

1, 2 - on bootstrap of virtual machine (vagrant)
3, 4 - work flow

Gubolin:
I just tried it, ionic works fine already. I don't want to build an extra box, I think that would be overkill.

Get the three scripts ([main](https://github.com/Gubolin/snap/blob/mobileapp/binary.sh), [mobile](https://github.com/Gubolin/snap/blob/mobileapp/mobile.sh), [desktop](https://github.com/Gubolin/snap/blob/mobileapp/desktop.sh)) and execute as described above.
You need to install UglifyJS (`sudo npm install uglify-js -g`) and I had to `sudo chown -R vagrant:vagrant /home/vagrant/.android`. The final apks are in the directory where you ran the script.

brianharvey:
That all sounds fine for you, but ask yourself whether your high school teacher can do it.  :-)  It's not going to fly with users until they can do a one-click install.

cycomachead:
> That all sounds fine for you, but ask yourself whether your high school teacher can do it. :-) It's not going to fly with users until they can do a one-click install.

I don't think we should expect users to compile offline builds. Wouldn't we make them available via the website? The one thing which is missing here is an update function. I'm not sure what capabilities Cordova and node-webkit provide for writing to the system, but ideally the app should be able to update files from the server based on modification times.

Gubolin:
> I don't think we should expect users to compile offline builds.

I don't, too. This started as an experiment for you as developers.

> It's not going to fly with users until they can do a one-click install.

I thought it would be easiest to share the built packages. For testing purposes, I uploaded mine [on the releases page](https://github.com/Gubolin/snap/releases). Sadly, the files are rather big (90MB per desktop platform, 1/2MB for mobile; 256MB all platforms, zipped) so it takes me a while to upload them.

> The one thing which is missing here is an update function. I'm not sure what capabilities Cordova and node-webkit provide for writing to the system, but ideally the app should be able to update files from the server based on modification times.

At the moment it is quite difficult to have an auto-updating cordova application (you cannot use the Android app, anyway, so this is a problem that should be solved later). With node-webkit, it seems to be rather easy (see https://github.com/shrewdlogarithm/node-webkit-auto-update-demo).

nemilya:
Hello, All!

This is my example of snapmobile-box: https://github.com/nemilya/snapmobile-box

Gubolin:
@nemilya Nice!
Do you think the mobile version should be small or that it should include everything the 'normal' Snap*!* has, too? The `help` folder is 11MB big which is about 75%.

nemilya:
@gubolin mobile version not include IDE - just Play, help files not needed to include, I think.

I plan to publish example to Google play - and think that possibility to have a number of examples in one apk snap will be very useful - for example five demos - example of cordova integration, examples of Snap applications, example of porting Scratch.

Show list of demos on left top menu when clicked.

Then is GPS so important? It need permission - and it looks strange when application ask for my position) For advanced demo this will be good.

Also issue about music on Snap in cordova - needed media plugin.

Then for future - will be interesting to test https://crosswalk-project.org - speed have to be 10x faster. But app size ~40Mb.

Also for current cordova mobile snap version - will be interesting to have feature hot upload snap.xml to working android application. Directly from browser IDE. For development it will be very useful.

Also I have plans to make simple web service (and include it in snapmobile-box) for build apk from web-form where to upload just snap_project.xml - i.e. user not need ssh - just open in browser service and upload project's xml. For this service the amount of build time is important - and because of this I've made some changes in build script.

nemilya:
Updated https://github.com/nemilya/snapmobile-box

Now VM is correct doing `vagrant up` ~30 min for configure guest VM (snapmobile-box).

After that ~1 min for building Android's apk.

nemilya:
Updated https://github.com/nemilya/snapmobile-box

For development added build apk with URL to Snap's xml:
```
$ ./build_android.sh URL_TO_XML
```

I.e. possible to update Snap's project at Smartphone without reinstall apk (just restart app).

nemilya:
Look at this example
http://tctltntcl.ekrivoruchko.com
I don't investigate sources yet - but it interact with accelerometer of my android smartphone from within web page and chrome!

If include this possibility directly in Snap - it will be easy to play at mobile. For example builded in variables - '_mobile_acc_x' and tests Boolean '_mobile_accelerometer_available' variable.

If user create this variables - they are automated filled by Snap! And will be possible to make conditions and control of main sprite by delta of x y z accelerometers

nemilya:
Article about how device orientation works in html5
http://www.sitepoint.com/using-device-orientation-html5/

Gubolin:
Whoa, nice!!!

Crosswalk is exactly what I was looking for! The Android keyboard bug was a big showstopper. With the Crosswalk WebView this is hopefully fixed.

I saw how you added the possibility to load the XML at load, I like that. *added to `binary.sh`*

About GPS permissions: If you don't want it, don't add the cordova plugin.

What do you mean with the device orientation? There are already accelerometer reporters.

**tl;dr**
Keep up your good work! I'll try Crosswalk now.

Gubolin:
Now it works (finally). With crosswalk, there is no keyboard issue anymore.
I used `crosswalk-cordova-10.39.235.15-arm` because I got errors with the beta/canary versions.

You can download a debug apk on the [release page](https://github.com/Gubolin/snap/releases).

nemilya:
Great!

Will try later.
What about play sound?

Also will be very useful for test/demos to include 'load from cloud' feature in mobile version (shared projects).

Gubolin:
> What about play sound?

It should work out of the box.

> Also will be very useful for test/demos to include 'load from cloud' feature in mobile version (shared projects).

You just need to change `snap.html` from `config.xml` and `package.json` to `snap.html#present:UserName=foo&ProjecName=bar`. For convenience, I just updated the scripts so you can do the change via command line.

Gubolin:
> I used crosswalk-cordova-10.39.235.15-arm because I got errors with the beta/canary versions.

[npm install -g cordova-android-crosswalk](https://www.npmjs.com/package/cordova-android-crosswalk) works better, but I am too lazy to update the scripts at the moment.

q2dg:
Please, please, please, please!!!!

q2dg:
Why this issue got closed?

Gubolin:
> Why this issue got closed?

I opened the PR two years ago. To merge this, you'd have to resolve conflicts with the last year's code changes and test whether everything still works as expected. I do not have time for this because I no longer teach Snap*!* :(

Even simple fixes like #551, one-line changes like #545 or important functionality such as #544 remained untouched for two years.

If you actually use this, I can have another look at it.

q2dg:
@Gubolin Ok, I see. Thanks

Issue #433, opened 2014-05-12
Prevent Export Errors in Chrome


cycomachead:
Chrome (in it's current version) cannot handle URI's longer than ~2.09 million characters. When such a URL is given Chrome, an "Aw, Snap" page is displayed. 

If the exported XML is going to be longer than ~2.09 million characters, AND the browser is Chrome we should warn users before trying to display a new page with XML. The best remedy is to share using the Cloud.

EDIT: The other remedy is to use the `<a download>...` attribute which chrome supports and is currently a hidden option in _Snap!_

cycomachead:
Closing since this was mostly fixed but see #726

PR #588, opened 2014-09-25
Disable putting project contents in location.hash

cycomachead:
This is a **partial** resolution to #433 which causes Chrome to crash.
Placing the project contents in the URL bar will lead to crashes on Chrome
when the URL text is > 2.09 million characters. This fix does not resolve
the problem of exporting a project (Chrome will 'crash' the new tab's process),
but at least the contents of current tab (Snap!) will remain intact.

One other benefit is that browser histories for Snap! users will decrease in
size. This has a noticeable affect in Safari on OS X where large histories slow
down searching and URL-autocomplete.

Use of #open:<project...> will still work the same, and in the case of opening
a shared cloud project, #present... is still set in the URL bar. This could (in
theory) leave open the possibly of a Chrome crash if a username or project title
are longer than 2.09m characters. However, such a case seems incredibly
unlikely.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -2765,7 +2765,6 @@ IDE_Morph.prototype.rawSaveProject = function (name) {
             try {
                 localStorage['-snap-project-' + name]
                     = str = this.serializer.serialize(this.stage);
-                location.hash = '#open:' + str;
                 this.showMessage('Saved!', 1);
             } catch (err) {
                 this.showMessage('Save failed: ' + err);
@@ -2773,7 +2772,6 @@ IDE_Morph.prototype.rawSaveProject = function (name) {
         } else {
             localStorage['-snap-project-' + name]
                 = str = this.serializer.serialize(this.stage);
-            location.hash = '#open:' + str;
             this.showMessage('Saved!', 1);
         }
     }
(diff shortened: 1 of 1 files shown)


cycomachead:
@jmoenig @brianharvey We've already run into this a few times. 

There was a suggestion of including the project contents in the URL via a shift-click menu -- that would work fine and wouldn't deal with trying to include any project names or IDs or anything, so you wouldn't actually need to save the project somewhere. 

There is also long-ago posted #537 which hides the URLs only when they are going to break Chrome.
Personally, I think this is more of a stopgap and that we ought to do away with the project URLs, though I'm OK with the shift-click option.

The current project URLs have other disadvantages in that they're pretty unfriendly to look at and deal with, and they pose problems w/ browser histories and search since they can be MB's in size.

jmoenig:
yeah, I'll work on something that lets me still use the url version. It's probably going to be a - hidden - setting which the browser remembers. I'm really relying on this feature when developing Snap and imo its benefits are overwhelming. But I understand your concerns and will make it so you have to explicitly turn it on if you want it.

cycomachead:
> yeah, I'll work on something that lets me still use the url version. It's probably going to be a - hidden - setting which the browser remembers.

Yeah, I definitely do use the refresh model when dev'ing stuff, so a hidden option makes sense.
(Plus when testing projects, you're not likely to be dealing with large amounts of media.)

PR #603, opened 2014-10-04
allow only numbers to be reported as number

Gubolin:
`is 3ab a number` now returns `false`.
Other blocks still convert it to `3`.


Gubolin:
see #715

PR #605, opened 2014-10-04
leave a mark on pen down (fix #209)

### LINKED ISSUE #209: The pen down command should immediately leave a mark
Executing the down command does not, at this point, leave any pen mark on stage:
down - up

Moving 0 steps does not help: 
down - move 0 - up

In order to leave a mark, one has to move at least one step:
down - move 1 - up

Scratch's behavior is preferable, I think: lowering the pen immediately leaves a mark. This is also closer to our intuition of what a real pen would do.

Gubolin:


bromagosa:
Closed as we now draw a dot when we move 0 steps.

PR #606, opened 2014-10-04
show an error message if project cannot be loaded (fix #190)

### LINKED ISSUE #190: Snap Doesn't respond when it can't open a project.
A student just tried to open a snap project using #open:...... but Snap simply showed a white screen until I added corsproxy to the url. 

Is there a way snap could warn a user if the project can't be opened?

Gubolin:


jmoenig:
yep, this is going to be pulled next. Thanks!

PR #624, opened 2014-10-25
Set location.hash to cloud URLs on project shares

cycomachead:
This will make it much easier to get the URL for a shared project.
It also removes the # from the URL when the open project is unshared.

I've tested this and it works but it seems a little weird to reference
the main IDE as `this.siblings()[0]` though I didn't see a better method
to use.

(Also, I thought there was an open issue for this but I couldn't find one.
However, this will definitely save some confusion among students and teachers.)


cycomachead:
@jmoenig @brianharvey Any chance of this being pulled? :)

cycomachead:
Bump. :)

jmoenig:
Yep, this is now ready to be pulled and integrated with the next batch of changes. Thanks!

PR #631, opened 2014-11-03
add "any key" to key hat block (fix #387)

### LINKED ISSUE #387: Generic Key Down Hat Block
I think it'd be nice, if we have a way of "When ANY key pressed" hat block. (Probably just an additional input to the current block).

However, there would need to be some way to getting at the last key pressed. A global "last key pressed" would be an option, I suppose.

Would it be possible to have an "upvar" in a hat block? That might be more awkward than the idea sounds, but I also ~~somewhat~~ dislike global variables.

Gubolin:
I did not implement 'upvars' because I don't know how, so here you go with a global variable :-)


jmoenig:
this has been superseded in the meantime. Closing

PR #689, opened 2015-01-03
Add the Examples folder to the Repo

cycomachead:
This will be handy for others distributing Snap! offline.
I also removed an unnecessary call in the text of the label block, but it
otherwise functions the same way.


cycomachead:
Thanks! :+1:

Issue #613, opened 2014-10-13
Documentation on Snap app structure is missing


confiks:
Hi, I'm trying to dig myself into the Snap source, in an attempt to see if it is feasible to use it in an educational project I'm working on. I'm trying to use a subset of your project, most probably only the script area and the execution window. So, considering two ways to dig, I did a depth-first search on definition and execution order, documenting along the way:
- 'contributing to BYOB4.txt' tells me a lot about contributing to the project, but very little about how the project is structured, and what responsibilities individual component/modules (Morphs!) have.
- snap.html lists a lot of Javascript files that push a lot of objects on the global scope, all of them 'Morphs'. The inclusion order might explain how the different components interdepend. Morphic.js is included first, then come GUI dependencies and the GUI, after which the subject matter of Snap is included: blocks and rules and data.
- Morphic.js is beautifully documented, and explains how Morphic works, and how morphs interact with each other. It unfortunately does not tell how the Snap application that is written on top works.
- The HTML page just executes: new IDE_Morph.openIn(new WorldMorph([canvas]))
- WorldMorph really belongs to Morphic, so IDE_Morph must define the Snap application. It does so by first calling buildPanes(), which builds all panes by creating zillion children and grandchildren Morphs. Then it adds itself as a child to the WorldMorph via add() and inherited Node's addChild().

At that point, all parts of the application are set in motion, and it become hard to see how the 'control flow' flows. After studying the code for a small hour I notice that responsibilities for many features are spread out across components. I would for example really like to understand how the state of the execution window is altered by the scripts that are defined in the scripts window.
## tl;dr

Could you please write down what you would tell someone in a quick 5 minutes, about how your project is structured, how the moving parts (of Snap!) interact, and how the program defined by the blocks is executed.

Issue #425, opened 2014-05-07
documentation folder in repo


marwahaha:
@cycomachead and I think this is a really good idea for new developers. Also, contributingtoBYOB4.txt is out-of-date.

towerofnix:
(Note: fixed in 6c5b33d631253a3a2c6704c32908aa4de24c7e73.)

Issue #197, opened 2013-10-04
Clean up directory - Lang files


joefarebrother:
PLEASE can you put all the lang-xxx.js files into a new folder? This will make it SO much easier to look through the folders to find one file! It only requires one little change to the code, in gui.js:

``` js
IDE_Morph.prototype.setLanguage = function (lang, callback) {
    var translation = document.getElementById('language'),
        src = 'langs/lang-' + lang + '.js', //There that's all you need to change... just add the folder name and a slash...
        myself = this;
    SnapTranslator.unload();
    if (translation) {
        document.head.removeChild(translation);
    }
    if (lang === 'en') {
        return this.reflectLanguage('en', callback);
    }
    translation = document.createElement('script');
    translation.id = 'language';
    translation.onload = function () {
        myself.reflectLanguage(lang, callback);
    };
    document.head.appendChild(translation);
    translation.src = src;
};
```

birbilis:
btw, to not break other forks and make it harder to merge changes manually, I think you have to use some special move (or something) command of the version control system to move those files into a new folder, not just make the folder and copy files there, then delete old ones

birbilis:
ahem, should also merge this or mark as duplicate, hadn't noticed this one
Cleanup root source folder #823

PR #690, opened 2015-01-03
Cleanup documenation

cycomachead:
I'm not sure if the movement of files will be a popular thing, but I think a proper docs folder will be helpful. :)

This PR doesn't change anything about Snap<i>!</i> itself, but it is a start to making development a tiny bit more clear.

Things that I'd like to do:
- Contributing needs to be updated to add instructions on Github, though I think we should still leave the Changeset info in there since it's perfectly valid still. 
- translating needs to reference the Contributing guide since they currently duplicate some info. 
- I'm not sure what to do with the original Readme.txt file, so I just left it there.

The all caps names are for Github. It will link to the contributing file on any pull request to remind anyone of the best practices.


cycomachead:
Reference:
#613 #425

brianharvey:
Your other two today were no-brainers, but I'm going to wait for Jens to weigh in on this; I suspect it'll break something (e.g., "Download source").

cycomachead:
Yeah, that's fine. (Download Source pulls from the zip file on s.b.e, which is made manually.)

brianharvey:
Yes but in includes some of those doc files (contributing, e.g.) and I bet when you say "manually" you mean "Jens runs a shell script."  :-)

brianharvey:
As I suspected, this one is rejected by Jens, who wants to keep snapsource as flat as possible.

cycomachead:
Ok, I kind of suspected the same thing -- but is there a good reason? The current folder is getting a bit unwieldy... (I recall other issues being posted about similar things). 

However, cleaning / improving the existing documentation would still be useful.  
Would this be pulled if the files were still in their original places?

brianharvey:
If you want to change Contributing, you should discuss the details with Jens by email.  /I'm/ certainly not pulling changes to that!  :-)

cycomachead:
@jmoenig @brianharvey Can we re-visit this, but perhaps keeping things flat, if necessary?

I'll fix the merge conflicts and push a few more tweaks.

jmoenig:
I really, really want to keep things flat, meaning specifically no sub-directories, ideally none at all, because that way I can easily sort everything by whatever I need in different circumstances. What's the problem we're addressing with nested folders?

cycomachead:
> What's the problem we're addressing with nested folders?

Mostly, it gets hard to find what's actually useful, and keeps things organized. (See #197)

However, it's not really super important. It would be good to have better docs and clean up the contributing file.

cycomachead:
@brianharvey @jmoenig 

I undid the docs folder, so now this just has some cleaned up documentation, and then I can work on adding more, but I'll make updating CONTRIBUTING it's own PR.

How does it look now?
See here: https://github.com/cycomachead/snap/tree/docs

Klortho:
:heart: 
Just my 2 cents: I think these changes are big improvements -- especially the list of references to how to get in touch with the community. 

> If you want to change Contributing, you should discuss the details with Jens by email. /I'm/ certainly not pulling changes to that! :-)

You can see the (whitespace suppressed) diff [here](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/690/files?diff=split&w=1#diff-6a3371457528722a734f3c51d9238c13), and there are no substantive changes. Just a little tidying.

cycomachead:
Many of these were recently handled by Jens, there's more I want to do, but I'll handle that later.

PR #691, opened 2015-01-03
Replace alonzo.gif with the newer alonzo.png

cycomachead:


cycomachead:
@brianharvey Not sure if there's any reason to keep the only one around, but if there is I can add it back. :)

PR #697, opened 2015-01-07
New Keyboard Shortcuts for Snap!

chrisk867:
Hi, my name is Christopher Kilian and I took the Snap Decal last semester. As my final project, I made some new keyboard shortcuts that I think would be useful for Snap. I forked the repository this morning so everything should be up to date. The only changes I made were in the objects.js file between lines 4734 and 4891. I tried to emulate the style the code was written in so the code would be readable and familiar. Here are the changed I made:
- ctrl q: Make a new comment
- ctrl y: Perform the "clean up" function that lines all the blocks up in a neat row.
- ctrl u: Undrop the last block you dropped. If you drop a large chunk of connected or nested blocks, it  undrops the entire chunk.
- ctrl d: Deleted the last block you placed.
- ctrl r: Ringifies the last block you placed
- ctrl e: Unringifies the last block you placed, but only if it is ringified.
- ctrl c: Copies the last block you placed. If you placed a chunk of connected or nested blocks, it copies the entire chunk.
- ctrl l: opens up the libraries menu
- ctrl i: imports tools
- ctrl b: create a new block
- esc: Exit presentation mode. Only works if you are already in presentation mode. One of the student instructors of the decal asked me to add this in. Note: the keyboard shortcut that already existed that stopped all scripts was already bound to escape, so I changed that one to ctrl esc.

I also made sure to add in the checks to make sure that these shortcuts only work when not in presentation mode (except for esc, which only works in presentation mode). Feel free to rebind the keys in any way you see fit. If you have any questions don't hesitate to ask.

Thank you!

### DIFF
--- a/objects.js
+++ b/objects.js
@@ -4735,59 +4735,68 @@ StageMorph.prototype.fireKeyEvent = function (key) {
         return this.fireGreenFlagEvent();
     }
     if (evt === 'ctrl f') {
-        return this.parentThatIsA(IDE_Morph).currentSprite.searchBlocks();
+        if (!ide.isAppMode) {ide.currentSprite.searchBlocks(); }
+        return;
     }
     if (evt === 'ctrl n') {
-        return this.parentThatIsA(IDE_Morph).createNewProject();
+        if (!ide.isAppMode) {ide.createNewProject(); }
+        return;
     }
     if (evt === 'ctrl o') {
-        return this.parentThatIsA(IDE_Morph).openProjectsBrowser();
+        if (!ide.isAppMode) {ide.openProjectsBrowser(); }
+        return;
     }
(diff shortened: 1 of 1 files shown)


scj643:
I would love to have this implemented since I have to use snap for my AP Computer Science class and using the mouse is a real pain.

liam4:
> * ctrl d: Deleted the last block you placed.
> * ctrl r: Ringifies the last block you placed

These sound great! Some modifications might need to be made for keyboard editing, though - I'd like to be able to use these key binds on selected blocks..

DavidGasku:
Would be great to use a key modifier to duplicate blocks. Ie, holding Shift while start dragging, would initiate duplicating the dragged block.

PR #700, opened 2015-01-10
Add ask and answer blocks to the tool sprite

cycomachead:
See the screenshots for implementations but these are very simple blocks which
will make basic OOP operations a bit easier until 4.1

`ask` takes in a reporter and a string and will call that reporter on the requested
sprite.
`tell` takes in a script and a string and runs the script on the sprite.

These are sensing blocks and there are two differences from the BYOB tool sprite:
* These use string and not "objects" since that doesn't exist yet.
* I did not (yet) implement the "input list" versions because I'm not sure they will
be used as frequently, though I do think they should be in some library.


cycomachead:
Implementation:

screen shot 2015-01-02 at 8 32 39 pm

cycomachead:
@brianharvey Any thoughts?

brianharvey:
Dunno -- once we have 4.1 we're going to have different, incompatible ASK and TELL, so users may be confused.  Teachers?  How do you vote?

jmoenig:
You know, these blocks are *so easy* to build yourself...

cycomachead:
Yeah, but the point of the tool sprite isn't necessary about easy to build
blocks.

(The ignore block, the throw block, and a lot of the word manipulation
blocks are also pretty darn simple.)

The point of these blocks would be to allow students to use OOP-like
paradigms w/o first knowing how they work.
ᐧ

--
Michael Ball | michaelballphoto.com

cycomachead:
> Dunno -- once we have 4.1 we're going to have different, incompatible ASK and TELL, so users may be confused. Teachers? How do you vote?

@brianharvey I was looking at some of the discussion of 4.1 again and BYOB -- won't these blocks still behave the same way in 4.1? (Including being implemented externally?) The `OF` block would just be updated to _also_ accept an "object" in addition to a string name, right? 

When 4.1 comes, all we'd need to do is change the block specs of these to use the object icon, rather than a string input, but everything else should work the same.

Mostly asking again because I know many cases where these would be helpful for students.

jmoenig:
I'm not against merging this now. Complete 4.1 OOP is going to take more time and in the meanwhile these "zombiefication" tools are fun...

cycomachead:
Closing and superseded by #887

PR #703, opened 2015-01-16
updated Catalan translation

bromagosa:
Translated all new strings that have been added to Snap! since my last translation, plus corrected some others that were a bit unclear.

### DIFF
--- a/lang-ca.js
+++ b/lang-ca.js
@@ -1225,17 +1225,17 @@ SnapTranslator.dict.ca = {
     'sqrt':
         'arrel quadrada',
     'sin':
-        'sinus',
+        'sin',
     'cos':
-        'cosinus',
+        'cos',
     'tan':
-        'tangent',
+        'tan',
     'asin':
-        'arcsinus',
+        'asin',
     'acos':
-        'arccosinus',
+        'acos',
     'atan':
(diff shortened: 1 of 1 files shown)


lang-ca.js
```diff
@@ -1187,29 +1219,31 @@ SnapTranslator.dict.ca = {
...
     'abs':
-        'abs',
+        'valor absolut',
     'floor':
         'part entera',
     'sqrt':
         'arrel quadrada',
     'sin':
-        'sin',
+        'sinus',
```
bb010g: Translations like these seem a bit odd to me. "sin" is the universal mathematical symbol for the sine function, AFAIK. (If it isn't what's used in Catalan mathematical works, though, please say so. :))

bromagosa: You're right, I got carried away because I noticed I had left "abs" in there (and we don't usually use that abbreviation), so I changed that one and kept going. I'm correcting the rest.

Thanks for noticing! :)

jmoenig:
Hi @bromagosa , are you still working on this, or can I pull it (later today)? Tnx! -Jens

bromagosa:
Hey @jmoenig, I'm done :)

Thanks!

Issue #446, opened 2014-05-22
Language codes


tjvr:
Scratch 2.0 and Snap! use different language codes to mean the same
thing:

``` javascript
    'ja-hr': 'にほんご', // Scratch
    'ja_HIRA': 'にほんご', // Snap!
    /* ... */
    'pt-br': 'Português (Brasil)', // Scratch
    'pt_BR': 'Português do Brasil', // Snap!
    /* ... */
    'sk': 'Slovenčina', // Scratch
    'sl': 'Slovenščina',
    'si': 'Slovenščina', // Snap!
    /* ... */
    'zh-cn': '简体中文', // Scratch
    'zh': '简体中文', // Snap!
```

Scratch: […/locale/lang_list.txt](http://scratch.mit.edu/scratchr2/static/locale/lang_list.txt)
Snap!: [locale.js](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/locale.js#L115)

PR #713, opened 2015-02-06
Added language auto detection feature

karenarialin:
On line 3451, added function this.getBrowserLanguage to detect the user's browser language. On line 1759, we call getBrowserLanguage to automatically refresh and reload in the user's browser language (overrides default English language).

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -1756,11 +1756,7 @@ IDE_Morph.prototype.applySavedSettings = function () {
     }
 
     // language
-    if (language && language !== 'en') {
-        this.userLanguage = language;
-    } else {
-        this.userLanguage = null;
-    }
+    this.userLanguage = this.getBrowserLanguage();
 
     //  click
     if (click && !BlockMorph.prototype.snapSound) {
@@ -3451,6 +3447,17 @@ IDE_Morph.prototype.saveProjectsBrowser = function () {
 };
 
 // IDE_Morph localization
+
+IDE_Morph.prototype.getBrowserLanguage = function () {
+    var lang;
+    if (window.navigator.languages) {
+        lang = window.navigator.languages[0];
+    } else {
+        lang = (window.navigator.userLanguage || window.navigator.language);
+    }
+    lang = lang.substring(0, 2);
+    return lang;
+};
 
 IDE_Morph.prototype.languageMenu = function () {
     var menu = new MenuMorph(this),


jmoenig:
nice, thanks! This has been suggested by a number of non-English speaking users in the past, so we should discuss it.

To be honest, though, I don't like websites that automatically switch to a  - in my own case German - translation whenever I log on. Scratch is a good example. It keeps forgetting that I want to see the site in English and every couple of days I have to scroll all the way down to reset it to English. You can manually set Snap to a translation, and it will remember your choice whenever you open it again. Also, if you're a teacher and want to always show Snap in a particular language you can share a link to URL that forces Snap to translate to a certain language (for example, German):

    http://snap.berkeley.edu/snapsource/snap.html#lang:de

(careful, after you follow this link you'll have to reset the translation to your preferred language, hehe!)

brianharvey:
> every couple of days I have to scroll all the way down to reset it to English.

That's just a bad UI design -- most translatable sites have little flag icons at the top of the page.  And someone who encounters Snap<i>!</i> for the first time without a teacher might not realize that it's even capable of supporting a language other than English, so the browser's native setting is a better first choice.

Does Scratch translate all the help files?  That's something I wish we did.

bb010g:
> You can manually set Snap to a translation, and it will remember your choice whenever you open it again.

It would seem, then, that we could just use the browser language & store it on non-initialized runs and use the stored lang afterwards, as long as Snap is reliable in its storage, which it seems to be.

cycomachead:
> Does Scratch translate all the help files? That's something I wish we did.

If we stop using images for help files we could at least use a service to translate them. 



> It would seem, then, that we could just use the browser language & store it on non-initialized runs and use the stored lang afterwards, as long as Snap is reliable in its storage, which it seems to be.

Yeah, it seems like the first time you load Snap<i>!</i>, the app could check for a saved browser setting and if it isn't present, use the auto detection.

jmoenig:
another idea would be not to switch automatically, but to pop-up a speech bubble or something offering you to switch to what the browser assumes to be your language (if it's different from the current one). This "bubble" could simply go away as soon as you click on something else (just like a tooltip or a menu), and you wouldn't see it if you open Snap in presentation mode.

cycomachead:
Also, if we do this, we will need to make sure all the help files are properly named or map the few exceptions we have. #446 Was discussing that i believe.

jmoenig:
Well, the current help file mechanism - which we inherited from Scratch 1.4 - is clearly a kludge. The *right* way to go about this would be for each help panes to be it's own actual (fully interactive) mini-project!

khotchkiss13:
> another idea would be not to switch automatically, but to pop-up a speech bubble or something offering you to switch to what the browser assumes to be your language (if it's different from the current one).

I like this idea. 

I think that if we checked the browser language, prompted the user about changing their language if it doesn't match, and then if they select change, it would change it, or they exit the prompt and nothing changes. There could also be a "don't ask again" option so that it doesn't reappear all of the time.

Issue #707, opened 2015-01-26
Cannot ge a pen up event in iPad


haviital:
It is possible to detect only pen down event (mouse key down). When i touch the screen "mouse key down" is true and then immediately false, even if I keep finger down. So there is no way to get pen up event, which is quite essential.

 But Snap is however a great tool and can become the first really cross-device scratch-like framework because of javascript! Just keep doing the good work and improve the iPad functionality too.

jmoenig:
Thanks! I cannot reproduce this. On my iPad Snap's ```mouseDown?``` reporter shows the current state of whether a finger touches the glass or not.

haviital:
Really? A simple example is below. I have iOS 8.1 in my iPad3 3rd gen. I have tested it both on Safari and Chrome browser.

![image](https://cloud.githubusercontent.com/assets/1862346/5916526/ea9f676c-a61a-11e4-9941-5a879889a5e5.png)

jmoenig:
yes, I was testing the exact same script on my iPad, same iOS version. Whenever I put the finger on the glass the sprite says "true", when I lift it up again it says "false".

jmoenig:
Here's what happens on my iPad:

![touch](https://cloud.githubusercontent.com/assets/2596285/5916807/9b57e954-a615-11e4-91f0-2f18cedeadec.gif)

haviital:
Ok, I cannot deny that :-) I have to retest this at home with iPad today.

jmoenig:
I have to say that the reaction *is* kinda slow, though... there is a distinct pause between the finger actually touching the glass and the sprite saying so

haviital:
I double checked the problem in my iPad. If I keep the finger down for few seconds the message changes to "false", without lifting the finger. It does not happen every time. Sometimes it works as expected, but over 50% of time it works incorrectly.

The fps is also really slow in general. Something like 1 fps! It can be related to touch problems. Perhaps touch events are not handled fast enough and they return error(?).

haviital:
I noticed that using "Say" command in Forever loop makes Snap very slow (about 1 fps) on iPad. Changing sprite costume based on mouseDown state is much cheaper. 

However the original problem stays. I even took the Snap source and modified the SensorMorph example in morphic.js and made the html-test page. You can test it here: http://koti.mbnet.fi/~haviital/snap/hannu_test1.html

How to reproduce the problem:
[...]

PR #723, opened 2015-02-17
Fixed touch up event (e.g. in iPad) so that right mouse click simulation do not mess it.

haviital:
Symptoms:
In iPad, the touch up event (i.e. "mouseClickLeft" ) do not come if the finger do not move. That is because after 400 ms the touch is interpreted as a long press and the right mouse click simulation begins. The simulation overwrites the original left mouse button state. Because of this defect it is not possible to make game with touch controls for iPad.

The Fix:
- store the current mouse button value before simulating the right mouse click
- return the information on whether the context menu was really opened after the right mouse click
- if the context menu was not opened, restore the original mouse button value
- Note: the fix really affects only to touch functionality. The Mouse functionality should be as it was. 

Note:
I am a very experienced C++ coder but my Javascript knowledge was practically zero. Hopefully, the code is not too C++ like :-)

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -6909,16 +6909,18 @@ MenuMorph.prototype.popup = function (world, pos) {
         world.activeMenu.destroy();
     }
     if (this.items.length < 1 && !this.title) { // don't show empty menus
-        return;
+        return false;
     }
     world.add(this);
     world.activeMenu = this;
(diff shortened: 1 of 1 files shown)


haviital:
Relates to the issue #707

PR #744, opened 2015-03-17
Added embed.html to allow embedding projects

bromagosa:


jmoenig:
Thanks @bromagosa ! I wasn't aware just how easy it was to use iFrames :-)

So, I wonder whether we can simply tweak Snap so we can do the exact same thing with snap.html. What would we need? Looks like you allow more options (hide controls), but that's a brilliant idea, and people have been asking for more options anyway (e.g. hide some primitives etc.), so it might be a good idea to think about this.

bromagosa:
Sure! That would be ideal! Actually the only thing needed would be a parameter system that allowed you to tweak the UI. For instance:

* appMode=true/false
* start=true/false
* showControls=true/false

Also, if Snap! fired a callback/set some flag telling you when a project has finished loading, we could also get rid of the ``projectName`` hack.

Since this is something we really need for Snapi, I could take a look and modify this pull request tomorrow if you want.

jmoenig:
Wow, yes, that would be awesome, Bernat, if you were willing to investigate further! Thanks!

bromagosa:
@jmoenig this should be working as follows:

* ``#present:`` works as it did before but now accepts three different arguments
    - ``editMode`` → if true, will show the editor instead of fullscreen mode
    - ``noRun`` → if true, will _not_ fire the green flag event
    - ``hideControls`` → if true, will hide all controls

I've chosen negative logic in ``noRun`` because this way we get to keep the default behavior untouched.

I've tried it in local and it works, but I couldn't test it properly with cloud projects. Please test it before merging! :)

jmoenig:
Thanks, this is great, Bernat! I'll test it and then I'll pull it. Thanks!

jmoenig:
Oh, and then could get rid of #cloud: ... :-/

bromagosa:
Yep, #cloud: is now the same as #present:blahblah?editMode=true&noRun=true

It doesn't need to go though, maybe people are using it :)

jmoenig:
Hehe, maybe I'm the only one who's used it, we'll see :-D

cycomachead:
> Oh, and then could get rid of #cloud: ... :-/

I think a lot more people would use this if it were the default mode when opening a shared project. There are many people who would like there to be an easy behavior for sharing links which open in the editor.

jmoenig:
I haven't yet had the opportunity to test your changeset, @bromagosa , but I wonder whether we can make it, so it also disables the "do you really want to leave" message.

This is from a teacher:

A new question was posted by .... 

Remove the safety message when navigating away from published project?

My students are working on digital Snap! portfolios.  They are totally jazzed about being able to embed Snap! projects into their web pages, but annoyed that each time they navigate to a different sub-page they get the message are you sure that you want to navigate away from this page.  Can they disable this message easily?

bromagosa:
I've found it annoying too with the experimental embedded projects!

I'll do it tomorrow, what should I name the parameter? Something like _disableLeavePrompt_?

jmoenig:
it could be automatic with the "hideControls" switch, what do you think?

cycomachead:
> it could be automatic with the "hideControls" switch, what do you think?

I think that makes the most sense for now.

bromagosa:
I messed up a little bit, I forgot to pull before pushing and ended up with a gui.js file that's lost its history. Hope it's not a problem, but if it is I can open a new pull request and apply the changes there!

Anyway, this version prevents the system prompt to pop up when hideControls is true.

bromagosa:
@jmoenig, I've integrated this pull request into BeetleBlocks and it works perfectly!

You can see it in action here:

http://ericrosenbaum.github.io/BeetleBlocks/beetleblocks.html#present:Username=bromagosa&ProjectName=gear&hideControls=true

http://ericrosenbaum.github.io/BeetleBlocks/beetleblocks.html#present:Username=bromagosa&ProjectName=gear&editMode=true

http://ericrosenbaum.github.io/BeetleBlocks/beetleblocks.html#present:Username=bromagosa&ProjectName=gear&noRun=true

bb010g:
Could you squash these commits before merging?

bromagosa:
Yeah, but actually the changes are minimal... check this out: https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/744/files

gui.js
```diff
@@ -308,6 +308,14 @@ IDE_Morph.prototype.openIn = function (world) {
         }
     }
 
+    // This function returns the value of a parameter given its key
+    function getParameterByName(name) {
```
cycomachead: Not that it's a big deal, but isn't there a URL decoder in cloud.js that could be used?

bromagosa: Oh, you mean https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/cloud.js#L607? I didn't know about that, I'll change it!

bb010g:
[**@bromagosa**](https://github.com/bromagosa) commented on [Apr 9, 2015, 12:46 AM PDT](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/744#issuecomment-91141487 "2015-04-09T07:46:38Z - Replied by Github Comment Enhancer"):
> Yeah, but actually the changes are minimal... check this out: [https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/744/files](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/744/files)

Yeah. I'm not saying that there's a lot of changes and it needs a new PR or anything, just that [squashing](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) your 8 commits into 1 nice simple commit would be a lot nicer when browsing the commit history.

bromagosa:
@bb010g Thanks for the info, I didn't know I could do that in git. I tried to squash the commits, but there were commits from master being merged in there as well and I couldn't figure out how to do it properly. I'm issuing a new pull request.

bromagosa:
#768 is the new one

PR #759, opened 2015-04-02
added file fish.svg to costumes

feeneyp:
Trying to dig up public domain svg images for the costume library.  Idea is to provide some sprites that are a bit less "kiddy".  I will add more if this image suits.

Image found at http://pixabay.com/en/fish-trout-rainbow-trout-fishing-161013/


brianharvey:
Jens, do all the browsers handle SVG now?  I thought I remembered that 
that's why we didn't already have SVG costumes.

jmoenig:
Let me check....

cycomachead:
Is the intent to keep the SVG vecotrized? On import, we could always rasterize the file and that works fine, I believe.

brianharvey:
On 4/2/15 12:44 PM, Michael Ball wrote:
> Is the intent to keep the SVG vecotrized? On import, we could always rasterize the file and that works fine, I believe.

Then we might as well rasterize it in the library, no?  The whole point 
is for it to be resizable after import.

cycomachead:
> Then we might as well rasterize it in the library, no?  The whole point
> is for it to be resizable after import.

In this case sure, though if we did have a rasterizer it would work for dropped SVGs as well. :)

jmoenig:
> if we did have a rasterizer

Snap *does* have an SVG rasterizer since the beginning of time, but it now treats SVGs as ... um ... vector graphics. You can switch back to rasterizing then by shift-clicking on the settings menu head and selecting ```rasterize SVGs```

PR #762, opened 2015-04-05
Add many more sounds and ability to browse sounds directory recursively

kotl:


cycomachead:
I'm closing out old PR's. We aren't opposed to having more costumes and sounds, though! However, there should be a PR that just adds those and doesn't have other modifications.

PR #764, opened 2015-04-08
Restrict "Save to Disk" to only show in supported browsers

cycomachead:
This makes it so that only supported browsers show the save to disk
function
(currently Chrome 30x+ and FF36(?)+) This prevents unexpected behavior in
Safari and IE, and will automatically show the menu should these browsers
ever get support.


cycomachead:
Ugh there's a bunch of bad commits in here....let me fix that.

PR #767, opened 2015-04-08
Added no javascript text.

Logalicioustuff:
Since Snap! is ran on javascript, you need your browser to have javascript enabled, so if it's not enabled, let's show some nice big text saying it needs to be on. If needed or recommended, I can add this to the other html files that look like they need this as well. :-)


Logalicioustuff:
Actually, ignore this.... Let me do something ;)

PR #793, opened 2015-05-04
Update to Portuguese localization 

MMSequeira:


jmoenig:
Thanks, Manuel, for the update! I'm going to pull this along with some other translation updates that are scheduled to come up.

PR #803, opened 2015-05-18
Cmdline opts

brollb:


bb010g:
Squash the commits if you open a new PR.

cycomachead:
@brollb I don't think you actually meant to send the PR here, but is this project online anywhere to easily play with? The work looks very interesting!)=

brollb:
Yeah, I didn't mean to submit the PR. 

@cycomachead I haven't put the project online yet. I can do that sometime soon. However, it is worth noting that the project will probably be best hosted locally as there are different messaging/grouping paradigms which are configured on the start of the server (2-player/2-player turn based/n-player role based/etc).

That is, the server can be configured to group clients in pairs or require the users to register as a given name (and then create groups with unique names); it also can support turn based games in which the server restricts a player from playing twice in a row.

That being said, I would be happy to make the code available (with perhaps the basic 2 player grouping paradigm) and let you know if you are interested. 

If you have any other questions, let me know and I would be happy to answer them!

brollb:
@cycomachead I hosted an example of the project on Heroku: http://netsblox.herokuapp.com

Hopefully you like it!

PR #805, opened 2015-05-19
Fixed typo in Process definition

brollb:
Changed Process.prototype.contructor to Process.prototype.constructor


jmoenig:
Good catch, thanks!

PR #821, opened 2015-06-07
Remove harcoded url in gui.js (fix #814)

### LINKED ISSUE #814: Enhancement: remove hardcoded URLs in Snap! sources
Hi,
I wanted to host Snap! locally on my computer and add my custom blocks in 'librairies' and 'Example' folders. As it was not working, I looked at the code of gui.js file and saw that the path of librairies, examples and others files are harcoded... 
```
menu.addItem(
        'Libraries...',
        function () {
            // read a list of libraries from an external file,
            var libMenu = new MenuMorph(this, 'Import library'),
                libUrl = 'http://snap.berkeley.edu/snapsource/libraries/' +
                    'LIBRARIES';

            function loadLib(name) {
                var url = 'http://snap.berkeley.edu/snapsource/libraries/'
                        + name
                        + '.xml';
                myself.droppedText(myself.getURL(url), name);
            }
```
One of the big advantage I see in Snap! is to be able to start the application from the internet  (with http://snap.berkeley.edu/snapsource/snap.html) than on my own computer, without web access. With that, loading examples or libraires without web access become very tricky...

I tried to change all reference of 'http://snap.berkeley.edu/snapsource/' on gui.js by a variable *base_url* defined by:
```
var full_url = window.location.host + window.location.pathname ;
// Remove the last part of the full url (e.g. "snap.html")
var base_url = full_url.split('/');
base_url.pop(base_url.length-1);
base_url = base_url.join('/')+'/';
``` 
But it does not works well. 
Don't you think it should be a good idea to remove hardcoded urls ?

show0k:
It is tested in Chromium and Firefox (in a GNU/Linux OS).
It works well if sources are hosted on a web server (=served from http protocol)
If snap.html is only executed on the web browser from local path (file://path/to/snap/snap.html) there are issues: 
* In chromium [CORS](http://en.wikipedia.org/wiki/Cross-origin_resource_sharing) is not suported for "file://" URLs, so it is not possible to load examples or librairies
* In Firefox, CORS are supported for "file://" URLs, librairies can be loaded, but when it try to load examples files "TypeError: this.text.split is not a function" is raised...

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -82,6 +82,11 @@ var WardrobeMorph;
 var SoundIconMorph;
 var JukeboxMorph;
 
+// Get the full url without "snap.html"
+var baseUrl = document.URL.split('/');
+baseUrl.pop(baseUrl.length-1);
+baseUrl = baseUrl.join('/')+'/'; 
+
 // IDE_Morph ///////////////////////////////////////////////////////////
 
 // I am SNAP's top-level frame, the Editor window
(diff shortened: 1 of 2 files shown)


jmoenig:
Thanks!

PR #826, opened 2015-06-09
fixes #815

### LINKED ISSUE #815: Green flag event also triggers "when I receive any message" hat blocks
I think I may have found a little inconsistency, IMO it should either be that the green flag does trigger all ``when I receive any message`` blocks (as it is actually just like any other message) and then the ``message`` reporter reports ``__shout__go__``, or it doesn't trigger them at all.

The current situation is as I show in this project:

http://snap.berkeley.edu/snapsource/snap.html#present:Username=bromagosa&ProjectName=green%20flag%20bug

The first time you click on the green flag, everything goes as expected, the second time, both sprites speak at the same time.

bromagosa:


jmoenig:
Thanks, Bernat!

bromagosa:
My pleasure, as always :)

PR #828, opened 2015-06-10
Replace the Definition of the list→sentence block

cycomachead:
The changes the definition to use the `join` block from `join words`.
The call is now `join () ( ) ()` instead of `join words () ()`.
I noticed this has a substantial speedup when working with large lists.
See this project for some test cases:
http://snap.berkeley.edu/snapsource/snap.html#present:Username=cycomachead&ProjectName=list%20sentence%20profiler
(Method C is the method used for this change.)


brianharvey:
Thanks for finding this!  The slowdown was actually in JOIN WORDS; once I fixed that, your method A beat method C, so that's what I used.  I've pushed my version of your fix.

cycomachead:
Cool, I was going to take a look a join words next, but I guess this solves that. :)

(I'm fixing some stuff for a pull-request for a colors.)

PR #830, opened 2015-06-11
Part 1: First-Class Colors

cycomachead:
See the first commit message.

---

```
[Credit for some of the code and idea goes to Bernat's work on Snapi!]

This implements a reporter, `pen color`, which shows the current pen
color, and allows the pen to be saved to variables. (The variables can
then be used in the `set pen color to %num` block.)

When reported, "said", or used in a watcher, a color is displayed as a
small solid square.

Note -- this commit does contain the mechanisms for serialization of
color variables. (That will come next.)
```

What's next:
I think it would be great to have `color` blocks using RGB values and the Snap<i>!</i> color picker. 
It would also be nice to have a block that reports the RGB(A?) values of a color as a list.

@ddgarcia -- for just colors specifically- what else could be useful?


brianharvey:
Yes please, converters for color to and from RGBA, converters to and from Scratch color numbers (or, I guess, a list of Scratch number, shade, and ghost values).  Someone will probably want CMYK too I guess.  I'd like to have the whole feature ready for comments before we think about pulling any of it, right, @jmoenig?

cycomachead:
Yeah, there's no need to pull immediately. (I'm also making this against the `OOP` branch, so there will be some time before things hit everyone anyway.)

I think there are many people who will want all sorts of color conversions (HSL, CMYK, HSV, YCbCr, YUV...ooh how many more are there? :P), but all these could be implemented by the user w/ RGB access. 

I guess my real questions are what should the block specs be? 

* `color %clr` : a color reporter with a graphical color picker.
* `color r: %n g: %n b: %n a: %n` : accept RGBA inputs and report a color.
* `%n as RGB` : take in a Scratch number and report the RBGA values as a 4 item list.
(This will also accept a color as a drop in for the number.)
I'm also not sure what the name of the this function should be. Should it be called "color %n ..."? that's either more clear, or redundant depending on your view....

I think that's probably good. But the only thing that bothers me is that there's no clean way to go from a 4 item list to a color. This probably isn't a big deal, but I feel like it should be easy to chain the to and from RGB functions and get the same value.

Also, what should the ordering of the pen palette look like?

bb010g:
A [WIP] tag would be nice prepended to the PR title.

Those blocks look nice. The list -> block args problem would be alleviated with globbing, but that may be another discussion.

brianharvey:
Scratch colors are complicated because (if you have the arrow costume) the sprite's color is affected both by the Pen blocks (color, shade) and by the Looks blocks (brightness, ghost), but the latter two don't affect the pen trace.  (Maybe that's a bug? @jmoenig)

This may be the wrong idea, but I'm imagining that numeric representations would come in these forms:

{R, G, B}
{R, G, B, A}
ScratchNumber
{ScratchNumber, shade}
{ScratchNumber, shade, brightness, ghost, "Scratch"}

... so that we could infer the format from the length of the list.  (Btw I don't fully understand how brightness and shade interact.)

Then the COLOR block could let you drop a list or numeric expression onto the color square, which it would interpret as RGB(A) or Scratch based on its length.  And there could be another block COLOR->[dropdown with "RGB", "RGBA", "color number", "shade", etc] that outputs a number or list.

All the existing blocks would stay, and then at the bottom of the palette would be reporters

PEN COLOR (no inputs, reports current pen color as a color)
COLOR [color box] (reports the chosen color as a color)
COLOR->[pulldown] [color box] (reports number or list)

I think everything else could then be written in Snap<i>!</i>, including things like a COLOR block that takes input numbers R, G, B, A (maybe with A defaulting to 255).

Does that work?

cycomachead:
> Those blocks look nice. The list -> block args problem would be alleviated with globbing, but that may be another discussion.

Heh, I want that too. It exists for variadic blocks, so it'd be cool for others. 

> (Maybe that's a bug? @jmoenig)

I don't think it's a bug, at least not by my understanding. The graphics affects modify the look of the costume, which is a separate thing from the pen. I guess this is somewhat confusing when the costume is a turtle who's color is the pen, but really that isn't often the case. Also, AFAIK, this is the same behavior as Scratch.

As far as the details -- if we don't have brightness and ghosting then that simplifies the lists, but it still feels a bit complex to have 4 list types. (We need dictionaries / associations!)

My biggest problem, though, is that it's not clear to users that you can drop in something over the current color dropdown, so I don't see any user intuitively trying to put in a 3 value list to get a color.
Certainly, we could make the blocks in Snap<i>!</i> ourselves, but for something like RGB, I really feel like that needs to be a default. I only came to scratch when I was old(er) so I can't say how intuitive that scale is because I still don't get it -- but I've generally found RGB colors to make sense in more ways for most people.


I am happy with a pulldown based block, though I still think there should be one which has some numeric input slot. However I would rename the block: `color [box] → [pull-down]`

(Side note: It would be super-duper awesome if a pull-down could change the block parameters to the right of the pulldown.)

> I think everything else could then be written in Snap!, including things like a COLOR block that takes input numbers R, G, B, A (maybe with A defaulting to 255).

Also, that reminds me -- is the scale for A [0,255] (ints) or [0,1] (float)? Because Snap<i>!</i> isn't consistent internally for the alpha value.

bb010g:
Ooh! Ooh! Ooh! This may be a good time to add [HUSL](http://www.husl-colors.org/) support. HUSL is a nicer, saner color format that I think would fit well with the general cleanliness of Snap*!*. Perhaps another dropdown option?

brianharvey:
We could compromise and add the block with four numeric inputs to tools.

Scratch doesn't have a default turtle costume, so the issue doesn't arise in Scratch.  That doesn't stop it being a bug for us, or at least a misfeature.  Wouldn't you like to be able to draw semitransparent lines?

About the complexity:  I'm assuming we have a color library with blocks like "Scratch color ()" and "Scratch color () with shade ()" and "RGBA color R: () G: () B: () A:()" and so on.  You don't have to assemble lists yourself, but they're handy because the block that converts colors to numeric representations can only report one value, so it has to be a list.

About alpha:  Now we're outside of my area of expertise.  Ask Dan, or Jens.

About pulldowns changing the block's prototype:  Don't even ask.  Jens will get mad and kill the whole pull request.  :-)

cycomachead:
> Scratch doesn't have a default turtle costume, so the issue doesn't arise in Scratch. That doesn't stop it being a bug for us, or at least a misfeature. Wouldn't you like to be able to draw semitransparent lines?

Yes, but I don't think should be done through `ghost`. I haven't tested, but given a proper RGBA input, Snap<i>!</i> should handle the transparency just fine. 

> About the complexity: I'm assuming we have a color library with blocks like "Scratch color ()" and "Scratch color () with shade ()" and "RGBA color R: () G: () B: () A:()" and so on. You don't have to assemble lists yourself, but they're handy because the block that converts colors to numeric representations can only report one value, so it has to be a list.

I'm cool with a library, though I'm slightly concerned about discoverability, but I guess if it's named `Colors` and the help info references it, then it's not so bad. I guess, the primitives will still report lists, but perhaps that's not too bad?

bb010g:
[**@cycomachead**](https://github.com/cycomachead) commented on [Jun 11, 2015, 9:53 AM PDT](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/830#issuecomment-111205650 "2015-06-11T16:53:59Z - Replied by Github Comment Enhancer"):
> > About the complexity: I'm assuming we have a color library with blocks like "Scratch color ()" and "Scratch color () with shade ()" and "RGBA color R: () G: () B: () A:()" and so on. You don't have to assemble lists yourself, but they're handy because the block that converts colors to numeric representations can only report one value, so it has to be a list.
> 
> I'm cool with a library, though I'm slightly concerned about discoverability, but I guess if it's named `Colors` and the help info references it, then it's not so bad. I guess, the primitives will still report lists, but perhaps that's not too bad?

This would be a good place for a reader macro equivalent. :) Look at a nice Morphic.js color picker; really use a list of numbers.

cycomachead:
Ok, I think I got this all done! 
You can play with it here: http://cycomachead.github.io/snap/

Here's the blocks:
screen shot 2015-06-15 at 3 53 52 am

There is still the Snap<i>!</i> library to build, but that won't be too difficult. :)

@brianharvey @jmoenig Thoughts?

cycomachead:
Note that b73107b3f3536b699155d4c458062676f5c644af modifes the `%clr` input, and I further modified the `touching` primitives so that they also accept Colors in list format. 

IMO, it seems like all color slots should behave the same way, but I suppose it could work to special-case only the `color %clr` block to accept list inputs. Thoughts?

jmoenig:
Oh, that's really nice, Michael!

cycomachead:
I made some further improvements just now --

* added human-readable blocks as a library called "Color Reporters" (would "Utilities" be a better word?)
* allow equality comparisons of colors

cycomachead:
Brian points out I should have a block which handles web named colors.

Also, I think we might want an HSV function since that's common.

cycomachead:
@brianharvey 

Two things ---
In addition to the named colors, would a hex color block be useful?

Unless I modify the internal color structure in Snap<i>!</i> the "named color" block will need to be a `JSFunction` if it's implemented just as a utility is that OK

brianharvey:
Ideally, there would be a block into which you can type any color representation in the known universe, and the block would parse it.  So, you know, 0.255.0 would be green.

I don't really care about the boundary between primitive and defined blocks, although I don't see why a Snap! implementation couldn't just have a big list of lists, ((red 255.0.0) (blue 0.0.255) (black 0.0.0) ...) built in.

cycomachead:
@brianharvey @jmoenig 

Ok, I think this is (nearly?) done! :grinning: 

See ea59d4e81244cb0cb4714b283fe455b83e4f6906 for some notes.

The gist is that this uses native CSS parsing of colors which are arbitrary strings. It doesn't quite to "any color representation in the known universe", but I do think it's fairly close. This allows named colors and hex colors as well as `rgb()`, `hsl()` formats (and the corresponding alpha formats).

There's really only one issue that I have:
__How do users discover text input of colors?__
I think we need a new block, but I don't really want to make a new primitive, though it isn't that terrible.
This block could have the spec `color %s` or "color from text %s" with a default parameter of "blue".
This would be identical to the `color` function in utility, just have a different spec for readability.

Personally, this basic block is would be so useful I'd vote for adding it to Tools over the (new) color library.

Regarding the color library:
Are there any additional color blocks that should be added?
Right now there's the blocks below. I think that's probably good, since anyone can add their own, but I'm willing to create (a few) more. The only complex one of these is the HSV block, but really it's not that bad. 

colors


I'd love for this to be in the 4.1 beta, or even current Snap<i>!</i> since it doesn't require the rest of OOP. ;)

bb010g:
Why are all those string args in `join`s?

cycomachead:
> Why are all those string args in joins?

There's no way to put text in a `ColorSlotMorph`, which is why I mention the concern over discoverability requiring an additional color block. It's trivial to create on but I'm just not sure of the "best" name / spec for the block. The color block is really one function that can take in quite a few different types of data (Color, List, Number, Text) and do the right thing. The trick, I think, is to convey this without resorting to a dozen different blocks. 

Right now, I think a simple block `color %s` would work just fine to be in the tools library. (Optionally: additionally the Colors library?). However, this one block would support Hex Colors, named colors, and the more obscure CSS options. Really, I don't expect / nor intend to expose the CSS options to the normal Snap<i>!</i> user. (Having them is fine, I believe, since there's no reason to limit functionality. It won't break anything.)

cycomachead:
Oh, I just thought of something else:

Should "color" be a new "type" in Snap<i>!</i>? (Such that it appears in the "is a ?" block menu?)

Also, I should never say that something is done.

studej:
Can we choose `color picker` as parameter type? I would like to create for example augmented `is touching` predicate with variadic parameter of type `color` :-)

bb010g:
:+1: for a opaque color type, with scratch color, HSL, RGB, &c just being generators for that.

cycomachead:
> Can we choose color picker as parameter type (or in future)? I would like to create for example augmented is touching predicate with variadic parameter of type color :-) That's the reason why have color as new data type @cycomachead

I would say, this would be something for the future. :)

PR #835, opened 2015-06-16
Add a ? to CONTAINS spec to follow Scratch

cycomachead:
See LLK/scratch-flash#621
This modifies the translations to use the current translation, but
does NOT modify the content of each translation, since I'm sure not
every language uses "?" in the same way.


jmoenig:
Scratch added the question mark sometime into v2.0. I'm not convinced of the question mark in predicates. After all, that's what the hexagonal shape of the block already implies, and we don't add question marks to logic operators either. A question mark looks like interpunctuation to me, but - almost all of the time - results in disrupted "sentences":

  ``` wait until not mouse down? ```

is somehow awkward, as is

   ``` if touching cat? move 10 steps ```

because aren't questions at all.

cycomachead:
It was only a few months ago - but from a consistency point of view - the contains block and the comparison operators are the only booleans without a "?". 

The view of how you read things is kind of personal, since I don't think the question mark really looks that out of place - and I wouldn't mind if the comparison blocks had "?" on then.

brianharvey:
In Lisp, where this question-mark convention originated, we don't have title text spread out among (or after) input slots, as we do in Smalltalk-influenced Snap<i>!</i>.  In the Lisp (or Logo) context, there's no issue about whether the code looks like English; it would be (CONTAINS? list value) where the placement of the question mark is clearly non-English regardless of whether the block is embedded in a larger expression.

Even in Scheme, by the way, the comparison functions are exceptions and don't have question marks in their names.

I really have mixed feelings about this.  The blocks whose names start with "is" (e.g., is ___ identical to ___ ?) would look really strange to me without the question marks, but I could take or leave the question mark on TOUCHING.

But, I do think that there's a lot of virtue to the redundancy of hexagons and question marks.  Especially for kids figuring out the language on their own, without a teacher or textbook to explain the block shapes, the question mark is a big help.

The CONTAINS block in particular turns out to be hard for kids to understand -- that's my experience as a teacher.  One issue is that the inputs are in the reverse order from the way they should be in terms of the real question that a kid wants to ask, which is "is THIS one of THOSE?"  (Remember, quite early in the BYOB3 thread, some kid asked for that block?  I put it in the tools sprite because I agreed that that was more natural.)

In fact I think it would be helpful to rename the block
     DOES list CONTAIN item ?
because it's always harder (for me anyway) to understand a block that doesn't start with a name.  The infix math operators are okay when they have short inputs, but if you write, say, the quadratic formula with an infix / instead of the big horizontal bar, it's pretty hard to read; I find it easier to read the Lisp prefix version.

So (I am only finding this out as I'm typing this) I guess I've sort of convinced myself to be on Michael's side in this discussion, only with the DOES renaming I'm suggesting.

cycomachead:
Hmm, I hadn't thought of using the word "does" but I do agree that it makes the block conceptually easier to grasp from the get-go!

> One issue is that the inputs are in the reverse order from the way they should be in terms of the real question that a kid wants to ask, which is "is THIS one of THOSE?" 

This is definitely the feeling I get from a lot of students, and even other programmers. (Many environments have some version of X in Y, JS being a bit weird in this case...but that's because its JS. :P)

PR #836, opened 2015-06-16
Fix Error Handing in the DATE block

cycomachead:
For some reason the function wasnt properly catching errors and was instead rasing native
JS errors. This fixes that bug, so the function will now report nothing when it doesnt know
what to do.

In addition, moving the variable `dateMap` inside the function reduces a global variable, which
shouldnt affect anything other than keeping the code clean. :)


jmoenig:
Thanks!

PR #837, opened 2015-06-18
Optimizations

nathan:
These changes make Snap<i>!</i> perform *much* better with large scripts and projects.

I wasn't able to JSLint it since JSLint doesn't support the options listed in `contributing to BYOB4.txt` anymore. Perhaps we could switch to JSHint?


morphic.js
```diff
@@ -3046,49 +3030,87 @@ Morph.prototype.fullCopy = function () {
     Other properties are also *shallow* copied, so you must override
     to deep copy Arrays and (complex) Objects
     */
-    var dict = {}, c;
-    c = this.copyRecordingReferences(dict);
+    var map = new Map(), c;
```
cycomachead: This breaks <= IE10, and Safari 7. That's not a huge change, but not totally insignificant either, IMO.

jmoenig:
interesting! How much faster is it?
The old JSLint version is still around at http://www.old.jslint.com/

jmoenig:
I was trying to compare a bunch of big real (Berkeley student made) projects, but didn't get far because they didn't load.

BTW, @cycomachead and @brianharvey , could we please, pretty please, have a (big!) collection of real, Berkeley-student-made final projects to test against? How about sharing (internally) all those ones that were presented at Cal Days. I only have a couple of projects that were sent to me because they didn't work at some time :-)

Thanks!

nathan:
> I was trying to compare a bunch of big real (Berkeley student made) projects, but didn't get far because they didn't load.

…link?

jmoenig:
they're files sent to me by students over time. I only have a few of them, but keep asking for more, so we can have a more comprehensive test suite. I think Michael can give them to us in no time. I'll send you a few via email...

nathan:
Fixed.

Note that these are almost exclusively UI optimizations—try, e.g., manipulating scripts and scrolling in larger projects. In the current s.b.e version I get about 2-4 fps while dragging a script in PathFinder, whereas in this version I get 25-60 fps. Loading is also incrementally faster (~5-20% speedup).

jmoenig:
Awesome, that was quick. Thanks!

jmoenig:
Wow.

That's pretty bad-ass, Nathan...

jmoenig:
So, not only does this load large projects almost 30% faster on my computer, it also runs many projects in unscheduled mode noticeably faster. This is very exciting!

(some of my WARPed benchmarks appear to take slightly longer, but not significantly. Overall this is quite fabulous!)

jmoenig:
Okay, this is *dramatically* awesome!

I've been racking my mind where to shave off a couple of milliseconds to compensate for the slightly longer method-lookup times in the upcoming OOP version, and *this* is a god-sent. Some effects are outright ... unspeakably ... fascinating. For example, for the first time working with clones is actually fun in Snap! Some of my projects that use clones were next to unusable in the current version, and with your optimizations they feel lightweight and ... snappy!!

Did you compare @aranlunzer's SnapMinder project in this version against the current one? Dramatic. That's all I can say.

I really want to pull these changes and encourage other forks of Snap (especially @ericrosenbaum's Beetleblocks, @bromagosa's Snap4Arduino, Snapi etc.) to also incorporate them.

Could we maybe have a quick Skype chat about this later today (tomorrow your time, I guess)? How about @brianharvey and @cycomachead join in that call. I'd like to discuss the consequences of pulling this, and maybe whether you'd object to (slightly) change a few (very, very minor) things? 

Thank you, Nathan!

P.S. We should probably hire you

brianharvey:
In a way I'm relieved that you found a bug, because tomorrow would be much better than today for a Skype meeting.  :-)  We have an all-day conference for teachers today.  I'm totally open tomorrow; you guys pick a time.

nathan:
> Could we maybe have a quick Skype chat about this later today (tomorrow your time, I guess)? How about @brianharvey and @cycomachead join in that call. I'd like to discuss the consequences of pulling this, and maybe whether you'd object to (slightly) change a few (very, very minor) things?

Sure!

nathan:
> it no longer calls isTransparentAt()

Oops, I did not mean to take that out. Fixed! (I don't think it impacts performance much, actually.)

nathan:
I think I found another bug, hang on.

nathan:
…also fixed.

cycomachead:
I'm avaiable for skype later today or tomorrow as well.

jmoenig:
Great, thanks Nathan! How about we all have a Skype chat tomorrow as soon as both Brian and Michael are awake and ready?

nathan:
> How about we all have a Skype chat tomorrow as soon as both Brian and Michael are awake and ready?

Works for me. Around what time would that be?

jmoenig:
I'll leave that to the Californians. Being a friday I'd prefer the earlier the better, you don't have to get dressed, we can do voice-only :-)

cycomachead:
I can do 9am, or 10am


(or even like 4:30am, but that counts as "tonight" for me, and I don't think that works for Brian :P)

brianharvey:
On 6/18/15 1:45 PM, Nathan wrote:
> Around what time would that be?

For me, awake = 8am, ready = 10am, but the latter is negotiable if 
necessary.

cycomachead:
If it's a video call, the delta from awake to ready is usually the time to make :coffee:

brianharvey:
Oh oops I answered the wrong message about the wrong Skype call.  I can do 8am if I have to, which is 6pm Jens time if I'm counting correctly, but I guess 9am is an okay compromise -- although that's exactly lunchtime for Nathan.  :-)  Just let me know.

jmoenig:
So, I've been investigating into some occasional JS stack overflows I've encountered with these optimizations when Map wasn't available. In particular when using the fallback code when ```Map```isn't available evaluating a command ring around more than just a single command block causes a JS stack overflow. In the course of these investigations I've benchmarked and profiled every single optimization to find out which ones a) might be causing those JS stack overflows, and b) which ones are "responsible" for the fabulous speed-up in various browsers.

Turns out the two biggest biggies are using ```Object.create()``` in ```copy()``` instead of the old ```clone()``` function, and - this came as surprise - using imperative iteration (```for```) instead of functional enumeration (```forEach()```) when scanning object properties a lot. These two optimizations alone account for most of the spectacular speed-up, because copying is at the base of so many operations in Snap (even when running scripts). It's also fantastic that @nathan has fixed the longstanding and annoying bug in ```updateReferences()``` which allows us to get rid of many methods that were introduced to work around it!

Another - surprisingly dramatic - optimization turns out to be the ```StringMorph::clearSelection()``` short cut, because, it, too, is performed whenever blocks are copied (and blocks are constantly copied in Snap).

Optimizing ```moveBy()``` is also a significant improvement, because it is at the heart of positioning visual elements in Snap. This together with no longer stepping SyntaxElementMorphs, no longer redrawing label parts unnecessarily anymore and optimizing ```HandMorph::morphAtPointer()``` makes the IDE much more responsive especially while projects are actively running.

A nice improvement is also the ```Block::thumbnail()```fix, which improves how blocks are shown in context menus, albeit at the cost of loosing their shadows, which is totally acceptable!

I've been tweaking these optimizations so they no longer cause the JS stack overflows I've been getting, and so they don't rely on the existence of ```Map```. I was surprised that just using a literal object as dictionary instead of the new ```Map```object didn't seem to have any impact on performance at all. Therefore I think it's okay to use this "traditional" method, which also has the benefit of working across all browsers, even older ones, and doesn't need any fallback code.

I also could not make out any noticeable difference in loading time between using the new ```Object.create()```mechanism instead of the traditional ```new``` for constructors when building up the inheritance hierarchy, and dito for the other optimizations. That's not to say we shouldn't ever apply them, but for now I feel more on the safe side with putting them on hold.

All in all these optimizations are among the very best contributions to both Morphic and Snap! Thank you, Nathan!

I'm not going to integrate all the changes in this pull-request, instead I'll commit - sometimes slightly reformulated - versions of most of these changes. I'll leave this pull request open, so we can keep discussing it, though.

bb010g:
If you could make sure to mention #837 in those commits, that would be great for keeping track.

cycomachead:
An easy way to give credit to @nathan for the work is to use `cherry-pick` to apply a single commit.

`git cherry-pick 6fbd2f27896d9cd749d92cc90495c4c4d7ab0ab0` will apply the parser improvements on the current branch. (I think you'll need to have fetched this pull-request to have the commits on your machine.)

nathan:
> I've been tweaking these optimizations so they no longer cause the JS stack overflows I've been getting,

Can you push these tweaks somewhere so I can see them?

jmoenig:
You're right @nathan , that it's better to merge and then commit changes afterwards, that seems like a good plan. I did try to get in touch with you yesterday on Skype but didn't succeed. If you find the time today maybe we can have another chat.

So, actually I was wrong, my tweaks also didn't work, and I'm now trying some other things to make it work. Can you reproduce the problem, Nathan? In short: The fallback methods when `Map`is not available, don't work - for the same reason my original copying scheme didn't (and which, btw, I still haven't fully understood. I suppose it's because JS Objects don't play well with objects as keys - partly because JS doesn't support identity testing -, which is exactly what the new Map structure addresses, right?)

I guess the clean solution to this would be to just ignore those browsers that don't support `Map` , because, after all, they're now official. This is what I'd prefer. But this compatibility chart that @cycomachead has pointed to is giving me the creeps:

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility

I'm pretty sure that there are many, many people using Snap on IE11, and I'd hate to make it worse for them.

But I'm not even sure whether I interpret that chart right. At first sight it appears that IE11 and Safari7 don't support `Map` at all, but on closer inspection the chart does acknowledge "basic support". Do we need more than "basic support"? Can anybody actually check this in IE11?

In scanning the code it looks to me as though we might even be fine by just relying on `Map` to work. Here are the usages of ```Map```  we would rely on:

in `Morph::fullCopy`:
```javascript
new Map()
```
in 'Morph.copyRecordingReferences':
```javascript
map.set()
```
and in `Morph.updateReferences':
```javascript
map.get()
```
These seem to be the most basic cases imaginable, they're not even covered by this chart. Does that mean that all of these usages are covered by all browsers in the chart? If this is true I'd vote that we don't even try to supply fall back code for older browsers.

Should we just try this? I'm slightly leaning towards this, but I don't want to piss off thousands of users...

Another solution would be to try to come up with fall back code that works without Map. I guess instead of using objects as keys in plain old dictionaries (as I've unsuccessfully tried in the current version) we could stringify them first. But this would both slow down the process a lot and probably not even work reliably, because we cannot ensure that those strings are going to be unique. The other solution would be to work around the issue by handling these cases explicitly in the `copyRecordingRefeferences` methods of each constructor that needs it. We know this solution works, because it's how we've been working around this for years. But it's such a kludge, as @nathan has correctly pointed out!

So, to sum it up, I do want to use `Map`, even exclusively so :-)

nathan:
> But I'm not even sure whether I interpret that chart right. At first sight it appears that IE11 and Safari7 don't support Map at all, but on closer inspection the chart does acknowledge "basic support".

"Basic support" means the empty constructor, `has`, `get`, and `set`, which are all we need. (See [kangax's compatibility table](http://kangax.github.io/compat-table/es6/#Map))

You can simulate a basic `Map` in older browsers by setting a property (like `objectId`) on each object to a unique value and using that property as the key in your empty-object-dictionary. You can also just use associative arrays, which might be somewhat slower but are also simpler and don't mutate the objects.

jmoenig:
Okay, if "basic support" is all we need, as I've suspected, I'd say we just go ahead with it. Thanks! I'll pull all of this later, and then commit a few changes, which we then can argue about :-)

cycomachead:
FWIW, the _vast_ majority of Snap<i>!</i> users are running versions of Chrome in the high 30's and FF is a good second, and then Safari is a distant 3rd. IE has users but really not much. 

(This is based on analytics on s.b.e/run -- so visits to the direct page aren't counted...)

In any event IE 11 would be good to support since it's the only default option on Windows devices... And the basics should be good enough, though it might be worth leaving a comment somewhere.


Also what ways are you (both) using to benchmark -- are you using dev tools are timing blocks in Snap<i>!</i>?

@jmoenig I'm working on getting the last 3 semesters of projects as well, but I need to separate out video files so I don't send 10GB to data.

nathan:
> Also what ways are you (both) using to benchmark -- are you using dev tools are timing blocks in Snap!?

I use the devtools profiler.

jmoenig:
I'm mostly using the timer in Snap :-)

Issue #579, opened 2014-09-13
Clicking "Share" should also update URL to make it easier to get a shared URL.


cycomachead:
Related to #63.

If a project is currently open, and the user clicks "Share" (from either the open or save dialogs) the user URL in the window should update to be the sharable URL for the project. This would make it **much** easier for users to get the URLs when sharing projects.

cycomachead:
Fixed with #624

PR #855, opened 2015-07-06
Fix a rather stupid and glaring typo in project sharing

cycomachead:
Totally my fault -- I think I botched something when fixing a bug when testing #579 -- anyway, this _does_ work correctly. :-)


cycomachead:
P.S. Have two properties that differ by a capital "P" is slightly confusing. ;)

cycomachead:
Tag: #624

PR #866, opened 2015-07-10
Fix URL bug in #821, and fix #864

cycomachead:
Verify it works here: http://snap.berkeley.edu/snapsource/michael/snap.html#open:https://inst.eecs.berkeley.edu/~cs10/labs/cur/programming/algorithms/../../../prog/algorithms/algorithms-framework-new.xml

PR #869, opened 2015-07-15
Fix type issue when adding blanks to lists

### LINKED ISSUE #863: Bug: List Equality Comparisons with Empty items seems broken.
Look at the following project in the sprite called "Contains".

http://snap.berkeley.edu/snapsource/snap.html#present:Username=cycomachead&ProjectName=2048_starter_file-list-bugs

When using the contains block with empty items, the results are unexpected.
It seems like either `add` or `=` isn't doing the right thing here.

`GAME BOARD` is 16 empty items, made using the `add` block.

cycomachead:
This is one small change that will need to be tested!
This makes the add function behaves like `in front of` and
uses a call to `isNil()`. Only values that are `nil` will be
added  to the list as `null` (in JS). This is more consistent with
Snap! so that values like the empty string are handled correctly.

This fixes #863. Contains will properly handle empty inputs and
when re-loading a list with blank elements added, they will no
longer be converted to 0's.

### DIFF
--- a/lists.js
+++ b/lists.js
@@ -158,9 +158,8 @@ List.prototype.add = function (element, index) {
     if no index is specifed, append the element
 */
     var idx = index || this.length() + 1,
-        obj = element === 0 ? 0
-                : element === false ? false
-                        : element || null;
+        obj = isNil(element) ? null : element;
+
     this.becomeArray();
     this.contents.splice(idx - 1, 0, obj);
     this.changed();


jmoenig:
thanks, Michael. This change absolutely makes sense to me at first glance. But I agree that we should test it. Using isNil() seems like the right idea here, and we've added this function specifically for cases like this, so I'd expect things to be okay...

cycomachead:
So far I've played around with my personal projects and as many esoteric combinations of 0's and empty strings and booleans and such as I could think of, and the lists are are reporting the correct values visually, and contains is working correctly.

jmoenig:
Okay, so how pressing is this? I'm planning to commit *a lot* of stuff sometime next week to prepare for the Scratch conference, among them

* a fix for dragging nested sprites
* keyboard editing support for scripts
* (hidden) inheritance of sprite-local variables

all of which I've been testing locally the past weeks. Can you wait until then or are your students about to riot?

cycomachead:
This one would help sooner, but after this Thursday this particular HW assignment will be due, and the issue will probably pass for a little while.

#866 Is actually somewhat more frequently occurring because it makes it hard to import tools from files that we link to in our labs.

PR #878, opened 2015-07-21
Complete Chapter I of the LaTeX version of the manual.

bl-nero:
Complete texification and Polish translation of the first chapter of the manual. This includes renaming the folder to manual-LaTeX as requested by Brian. Also, the title page is now ready (it's a parametrized macro, so that potential other translations of the manual can also benefit from it).

If it's possible, let me know if there are any problems with building (use pdflatex, run 2 times), because I'm working on a Linux distro, and I have never tried building it on a different platform.


jmoenig:
Thank you so much! I'll let Brian pull this...

bl-nero:
Oops, I think I've accidentally merged in the typo corrections to the actual Snap code... Let me fix it.

bl-nero:
OK, done.

jmoenig:
@brianharvey I think it's your turn to decide on this...

cycomachead:
Yes, would there be plans to keep the english version up to date? I'm not the biggest, fan of LaTeX, but I'd definitely be willing to contribute to a manual this way.

Also, we could use something like https://www.sharelatex.com/github/ that could automatically compile changes for us, if we wanted.

bl-nero:
@cycomachead We established with Brian Harvey that I'm going to gradually convert English version to LaTeX in parallel with translating. If you take a closer look at the directory structure, there are separate `en` and `pl` directories. And yes, this way contributing to the manual is going to be much easier. :) That was one part of my thought process when I decided to do the translation this way. Whenever the authors decides to change the book's text, the history will be clearly visible and it will be easier to maintain the parallel translation(s).

Regarding the sharelatex option, it sounds like a great idea. I'll give it a spin on my private repo and let you know whether it's working.

bl-nero:
The ShareLaTeX compiler works; however, it looks like it only supports one root file per repo (unless there are some undocumented options). We could build at least the English version automatically this way.

I'm wondering if we could integrate something more flexible here. Maybe Jenkins?

bl-nero:
Could this be merged?

brianharvey:
Sorry for the delay!  Things keep happening IRL.

PR #887, opened 2015-07-24
Add `ask` and `tell` blocks to the tools library

cycomachead:
See the screenshots in #700 for implementations but these are very
simple blocks which will make basic OOP (zombification) easier.

`ask` takes in a reporter and a string and will call that reporter on
the requested sprite.
`tell` takes in a script and a string and runs the script on the sprite.

These are sensing blocks and there are two differences from the BYOB
tool sprite:
* These use string and not "objects" since that doesn't exist yet.
* I did not (yet) implement the "input list" versions because I'm not
sure they will be used as frequently, though I do think they could be
in some library.


jmoenig:
Again, I'm happy to pull these. Any objections, @brianharvey ?

PR #889, opened 2015-07-24
Implement `Set Rotation Style`

cycomachead:
This completes the implementation.
- setting the block updates the sprite view on the stage
- also updates the buttons in the control bar

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -848,6 +848,19 @@ SyntaxElementMorph.prototype.labelPart = function (spec) {
             );
             part.setContents(90);
             break;
+        case '%rotateStyle':
+            part = new InputSlotMorph(
+                null, // text
+                false, // numeric?
+                {
+                    'left-right' : ['left-right'],
+                    'don\'t rotate' : ['don\'t rotate'],
+                    'all around' : ['all around']
+                },
+                true // read-only
+            );
+            part.setContents(['left-right']);
+            break;
         case '%inst':
             part = new InputSlotMorph(
                 null,
--- a/objects.js
+++ b/objects.js
@@ -290,6 +290,12 @@ SpriteMorph.prototype.initBlocks = function () {
             category: 'motion',
             spec: 'if on edge, bounce'
         },
+        doSetRotationStyle: {
+            only: SpriteMorph,
+            type: 'command',
+            category: 'motion',
+            spec: 'set rotation style to %rotateStyle'
+        },
         xPosition: {
             only: SpriteMorph,
             type: 'reporter',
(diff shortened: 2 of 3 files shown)


jmoenig:
I'll probably not pull this one, because it'll get superseded by v4.1's "attribute" block. In the meantime you can probably just offer a blocks library built around a single-liner JS function block...

cycomachead:
> I'll probably not pull this one, because it'll get superseded by v4.1's "attribute" block. In the meantime you can probably just offer a blocks library built around a single-liner JS function block...

Well, this is in Scratch, so there's a compatibility point. 

Will 4.1 have a `set` function? That wasn't in BYOB 3.1, at least not that I was ever aware of.

brianharvey:
Yes, although it'll be different in 4.1.  In 3.1 we were still trying desperately to minimize the number of new blocks, so we had the SET block for variables accept a lambdafied ATTRIBUTE block as its first input (replacing the variable name pulldown).  Nobody understood that, so in 4.1 there will be a SET ATTRIBUTE command block right next to the ATTRIBUTE reporter.

bromagosa:
This can now be done like this:

untitled script pic

It may not be the most obvious thing in the world, though...

jmoenig:
ha, @bromagosa! As of 5.0 this can now be set much more obviously:

set_property

jmoenig:
... which leads me to believe we can close this, right @cycomachead ?

bromagosa:
OMG I hadn't seen that! Thanks :)

cycomachead:
I mean... I guess but honestly, the current solution isn’t great. Neither Bernat nor I knew about it. :) 

Rotation style is an enum. Why are we making kids set numerical values? 
I don’t know what “2” means by reading that code. I don’t know what the options are. I don’t think selecting rotation style updates the right value with a drop down, does it? That might be acceptable, but having this block is also a point for Scratch compatibility. Minor thing, sure. But those minor things can add up to make going to Snap more confusing. 

-- 
Michael Ball
From my iPhone
michaelball.co

jmoenig:
well, it's in the v5.0 release notes. Brian and I did talk about it, but I didn't document it myself. The numerical values are the exact same ones we had in BYOB. Zero means no rotation, One means rotate freely. Those are fairly obvious I think, 2 means only rotate left / right, which is kinda mneonic, right?

brianharvey:
Once someone tells you the numbers are supposed to be mnemonic, you might be able to guess.  But my _first_ guess was that they'd be in the order of the three little buttons -- but they aren't.  Maybe when you set the first input to rotation style, the second input should become a pulldown.

But I hereby officially declare that Scratch compatibility is not a reason for us to do anything.  Many of our worst kludges are inherited from Scratch, and I hope that over time we'll fix them.

cycomachead:
Fair enough on the scratch compatibility. Though in this case, I don’t think think we are inheriting any kludges. 

But no. Numerical enums are never as good as string versions. I would also have assumed they are the the order of the buttons. Even when you work with the same enums every day you still have to remind yourself what’s a 0 and what’s a 1 and such. 

-- 
Michael Ball
From my iPhone
michaelball.co

PR #892, opened 2015-07-26
Colors

cycomachead:
This supercedes #830.

(Note there is now a merge conflict which is probably the tools library so I'll resolve that.)


cycomachead:
Wait, I'll have to do a bit of git-fu to extract all the colors stuff off of OOP. Hmmm.

PR #893, opened 2015-07-27
First Class Colors

cycomachead:
This is pulled against master, and includes only the color blocks commits.

See #830 for discussion and screenshots. (But new discussion is probably best on this thread.)


jmoenig:
Thanks!

threads.js
```diff
@@ -2572,6 +2586,106 @@ Process.prototype.reportAttributeOf = function (attribute, name) {
...
+        return new Color(list[0], list[1], list[2], list[3] || 1);
+    }
+
+    colorType = this.reportTypeOf(color);
+    throw new Error('expecting a color instead of a ' + colorType);
+};
+
+Process.prototype.colorFromPickerAsList = function (color, type) {
+    // ensure color is the proper format
+    color = this.colorFromPicker(color);
```
cycomachead: This is the line that should be handled by `blocks.js`, IMO.

cycomachead:
@jmoenig @brianharvey 

I just removed the pen color reporter like we discussed in AMS, so this only has two new blocks. (And I resolved the merge with the tools library file.)

I think this is ready. :grin:

jmoenig:
I'd like to think about this a little more and do some more experiments. TL;DR I'm not (yet) convinced that projects involving pixel manipulation make sense in Snap! at all, they work but they're too slow to be any fun.

Recently I've done a few experiments of my own inspired by a group discussion with @mjguzdial (Mark Guzdial) about some of his cool work in JES and GP, including John Maloney and @yoshikiohshima (Yoshiki Ohshima). 

Here's my latest experiment: http://snap.berkeley.edu/snapsource/snap.html#cloud:Username=jens&ProjectName=MediaComp

It extends Snap! with a few pseudo-primitives I've made using the JS-function block, the two important ones are querying the pixels of a costume, which returns a list of pixels, each pixel being a sublist of RGBA values:

pixels

and a command that overwrites a costume with such pixel data:

update

The good news is: Snap! is totally capable of handling image data as lists of lists, and it's fun (I think) to experiment with MAP and other HOFs to make your own graphic effects. Recursion works fine with large lists and that's really a blast. Sorta. Because...

The bad news is: It's s-l-o-w. I mean, really, really slow. Slow as molasses. I had to brutally sample down my demo picture to see anything happening at all before despairing. While I can quickly and easily make some interesting simply graphic effects, like shifting color channel values, it's an open secret that it pays to stay close to the primitives in Snap. Therefore a ```shift colors``` custom block works okay even if it uses a user-built ```map``` block:

shift

I did not use the *abominable* (imo) ```map``` block that's found in our tools (why oh why does it have to be so ugly, when it could be so beautiful?) but my own simple version:

mymap

By "works okay" I'm saying it kinda shows up instantly when applied to a brutally sampled down picture. When I use this block on a more "normally sized" picture of 480 x 360 pixels (Snap's default stage size), it manages to crunch the 172800 items long pixel list in roughly 17 seconds on my MBP.

As soon as I start using Snap's - wonderful! - capability to build domain specific block wrappers, such as:

accessor

which of course uses a - recursive, what else? - implementation of ```index of```:

idxof

etc. to create something students would hopefully write themselves (after an example by @mjguzdial):

decreasered

even ```warp```-ing it blows up the time it takes to crunch my 480 x 360 pixels picture to *over 4 minutes* on my pretty new and powerful laptop.

Of course, that's not even the *fun* and *engaging* part of a project, right? Because, after all, what we want to offer students is the ability to use their own graphic effects *interactively* in their own games, which involves querying, manipulating and redrawing pictures *a lot*.

Bottom line is, I don't want to encourage projects that I know will frustrate kids. If we add pixel manipulation primitives to Snap we're promising fun that Snap doesn't deliver. That doesn't mean we should totally stay away from this. However, a better approach might be to create a pixel extension with pre-made blocks (such as in my example here) and use that sensibly.

Thoughts?

bromagosa:
Actually, there are ways to achieve really fast pixel manipulation in JS by using Uint32Arrays, but I haven't given a thought at how hard would it be to abstract them to fit the Snap! metaphors.

Here's an example: http://snap.berkeley.edu/snapsource/snap.html#present:Username=bromagosa&ProjectName=moire

Based on an example by Mozilla: https://hacks.mozilla.org/2011/12/faster-canvas-pixel-manipulation-with-typed-arrays/

jmoenig:
Thanks, Bernat! So, what makes this example fast? You're using a JS ```for``` loop inside your ```moire``` block, is that it? My examples are slow because they use Snap's evaluator for everything, expecially to enumerate large lists. Creating that pixel data list happens really fast, as does updating the canvas image data with a Snap pixel list. Those aren't worrying me. It's working with large lists inside Snap that just isn't really fast. I think... (?)

cycomachead:
> Bottom line is, I don't want to encourage projects that I know will frustrate kids. If we add pixel manipulation primitives to Snap we're promising fun that Snap doesn't deliver. That doesn't mean we should totally stay away from this. However, a better approach might be to create a pixel extension with pre-made blocks (such as in my example here) and use that sensibly.

Wait -- aren't we getting totally ahead of ourselves?

First Class colors don't get anywhere near pixel-level manipulation in Snap<i>!</i>. They are simply useful for interacting with the pen, and are far preferable to the craziness and confusion that is Scratch colors. 

As far as actual pixel manipulation -- depending on the slowness I'm not sure it would frustrate people that much. There are so many cool things that you can do, that _some_ sacrifice is forgivable. That said, I think a pseudo-primiatve approach would be a good way to go for this, and you can get decent results. 

Those blocks and projects are really cool! (Both @jmoenig And @bromagosa !), and I'm definitely going to play around with them.

Colors are useful for so many things, not just pixel manipulation. We have `label` and blocks for drawing solid objects, and with 4.1 we'll have a way to report the pen color.
Really I see this as a sensible form of colors, that maintains Scratch compatibility.

I'm also still willing to re-think the API, but we definitely want something that gives a way to easily manipulate color.

jmoenig:
Thanks, Michael! I guess one of my questions is whether introducing first class colors as anything else than RGBA-Snap lists at the primitive level is a great idea. If we represent them as "just" lists we can write all the libraries in Snap itself, including hsv etc. I don't think we're getting too much ahead of ourselves when thinking about all the things we want to use colors for, and *how* to use them, including bitmaps and (first-class) costumes.

bromagosa:
@jmoenig I believe what makes it fast is using typed arrays.

In that sense, I think it may be quite hard to abstract this and make it "Snap!-like"... unless we cheat somehow and translate Lists into typed arrays just for internal calculations when possible.

I know, types... :confounded:

cycomachead:
> Actually, there are ways to achieve really fast pixel manipulation in JS by using Uint32Arrays, but I haven't given a thought at how hard would it be to abstract them to fit the Snap! metaphors.

Yeah, typed arrays would help things as a backend for some of the data manipulation. (You could also use UInt8's and have some wrappers which would make things easy, since canvas internally uses UInt8's)

But we could use typed arrays internally as a `List` and I don't think anything will break. Gonna go test....

jmoenig:
I'm afraid Snap's evaluator is by far the biggest culprit in terms of speed, and that any tweaks to the list data structure wouldn't even be noticeable...

cycomachead:
> Thanks, Michael! I guess one of my questions is whether introducing first class colors as anything else than RGBA-Snap lists at the primitive level is a great idea. 

I guess, it depends on what we want to do. I personally feel like primitive Colors are fine, as are things like

cycomachead:
> I'm afraid Snap's evaluator is by far the biggest culprit in terms of speed, and that any tweaks to the list data structure wouldn't even be noticeable...

If we're processing enough pixels, I think there's a chance they could be, but it does seem like a problem. :/ 
However, depending on what we want to do, you can build generic enough primitives that still let you do cool things, like build a filter in Snap<i>!</i> and apply a filter. As long as the critical stuff is fine, I think we can get decent performance.

jmoenig:
Right, Michael, that might work. We could experiment with a low-level (meaning JS) enumeration primitive for pixel data...

cycomachead:
> We could experiment with a low-level (meaning JS) enumeration primitive for pixel data...

I think Brian was thinking about this too...I think his idea was though that we could do it with something like a special for-each / map.

Of course, my favorite stuff will be if we get a good matrix library and have some fast matrix operations for canvas pixel data. :D

jmoenig:
Right, a special ```for each``` or ```map``` is what I'm thinking about, too. Problem is, I already *have* implemented both of these as primitives (in dev mode in the lists category). They are no (not much) faster than their custom counterparts. What's worse, my ```map``` primitive even overflows the JS stack (haha!). Bottom line is, as soon as we're talking about higher order functions we *have* to use Snap's evaluator, because those functions can have side effects and interact with ... anything else inside Snap.

bromagosa:
Hey, but my example was fast not because of the native loop, but because of the native typed array!

I still don't know how would one use typed arrays without resorting to some kind of ugly clutch, but maybe something like ``[ do {closure} for all pixels in (costume) ]`` could internally use a uint32array and run the closure at each iteration.

brianharvey:
Imho the primitive notation for colors should be a color picker square.  And then there should be selectors on colors to get, e.g., RGBA or (nearest) Scratch color number.

Using bitmap as an internal type doesn't have to look kludgy to the user.  Bitmaps just have to implement ITEM and REPLACE ITEM and FOR EACH ITEM.  The implementation can be efficient without telling the user so.

yoshikiohshima:
On Tue, Sep 22, 2015 at 4:48 AM, Brian Harvey <notifications@github.com>
wrote:

> Imho the primitive notation for colors should be a color picker square.
> And then there should be selectors on colors to get, e.g., RGBA or
> (nearest) Scratch color number.
>
> Using bitmap as an internal type doesn't have to look kludgy to the user.
> Bitmaps just have to implement ITEM and REPLACE ITEM and FOR EACH ITEM. The
> implementation can be efficient without telling the user so.
>

Not getting the full background of the discussion, but I'd second this.
Having a Color *and* Bitmap being objects would help it a lot.  (Etoys has
them for that matter).  The homogenous array can be used to represent a
bitmap.

(I heard a rumor that you guys are thinking of implementing Snap in itself;
having a Bimap object to render things would be a way for that...)


-- 
-- Yoshiki

cycomachead:
> Using bitmap as an internal type doesn't have to look kludgy to the user. Bitmaps just have to implement ITEM and REPLACE ITEM and FOR EACH ITEM. The implementation can be efficient without telling the user so.

That would be the intention, I think. We can actually mostly do that with lists right now, but we'd need a few wrapper blocks / functions. However, since JS is somewhat limited in the fast ways to process image data we might run into some issues about presenting things. 

I still think it's worth giving some tools though. Even low res stuff is fun! We used to have a floodfill lab which used a grid that was maybe 25x25 px (equivalent, of course! The pixels were giant. Lol.) and that was really fun.

> And then there should be selectors on colors to get, e.g., RGBA or (nearest) Scratch color number.

Should we have all types of data in Snap<i>!</i>. 4 built in color methods seems a bit much, but all 4 do seem to have good arguments for existing.

brianharvey:
> Should we have all types of data in Snap!

Not lots of data types.  There is only one color type.  The others are numbers or lists of numbers.  I think we had talked about using a pulldown instead of four blocks.

cycomachead:
> The others are numbers or lists of numbers. I think we had talked about using a pulldown instead of four blocks.

Sorry, yeah, I meant having 4 list formats as the primitive open. The current block still has a pull-down with 4 options which each return a list.

I was wondering whether it might not be simipler to have only 1 option for the primitive and have the tools / colors library contain addtional blocks for other conversions?

brianharvey:
Oh, I see.  Yeah, maybe RGBA.  Or just RGB if length=3.

PR #894, opened 2015-07-28
Fix the loading of URLs to be consistent.

cycomachead:
`baseURL` should always end with a trailing /.

Sorry for not catching this!

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -90,7 +90,7 @@ var baseURL = (function getPath(location) {
     path = location.pathname; // starts with a /
     origin = location.origin; // has no trailing /
     slash = path.lastIndexOf('/');
-    path = path.slice(0, slash); // keep a trailing /
+    path = path.slice(0, slash + 1); // keep a trailing /
     return origin + path;
 }(window.location));
 
@@ -2563,10 +2563,10 @@ IDE_Morph.prototype.projectMenu = function () {
         function () {
             // read a list of libraries from an external file,
             var libMenu = new MenuMorph(this, 'Import library'),
-                libUrl = baseURL + '/libraries/' + 'LIBRARIES';
+                libUrl = baseURL + 'libraries/' + 'LIBRARIES';
 
             function loadLib(name) {
-                var url = baseURL + '/libraries/' + name + '.xml';
+                var url = baseURL + 'libraries/' + name + '.xml';
                 myself.droppedText(myself.getURL(url), name);
             }
 


jmoenig:
thanks, Michael. It's online now and libraries, costumes, counds and examples are all working for me.

cycomachead:
Phew! :)

I'm going to also look into some refactoring later, and maybe look at doing things async, but I'm still looking at the blob data stuff.

jmoenig:
naw, forget the refactoring and the async stuff for now, they're not going to take that away from us anytime soon. The blob uri stuff is more important it seems.

brianharvey:
When you get around to worrying about such things, let's make "work offline" work.  Someone smarter than me has to figure out how not to require an enormous initial download.

jmoenig:
I'm tending to think a downloadable stand-alone app like Snap4Arduino or Tom's iOS Snap might be the best method of packaging this...

brianharvey:
Was that in response to me?  I think it's just a question of getting the manifest file really right.

jmoenig:
yeah, but we also want to be able to save projects permanently, and to even start up Snap when you're completely offline and cleared your cache, I guess

cycomachead:
> Was that in response to me? I think it's just a question of getting the manifest file really right.

Yeah, it is. I've actually looked into it -- it's not terribly hard to fix because right now the manifest file isn't linked to the the html file and there might be a formatting bug. _However,_ the big reason I haven't touched anything is that using a manifest file can make it really hard to deliver new updates to users if we aren't careful. It's very aggressive caching. (Additionally, it requires people to directly bookmark snapsource/snap.html which I don't know how well that works...) We'd probably want some automated release system which explicitly enforces cache busting if we went down this path.

That said, I'm in favor of supporting getting software in any way possible. :P So, I think an in browser offline cache wouldn't hurt, but it might not be quite as useful as other methods (like a stand alone app) which have other advantages.

brianharvey:
Right, I'm not suggesting we abandon the standalone app route (although it should update itself too).  The advantage of work-offline is that if we do it right you don't have to have planned ahead.

The current manifest works fine, except that the first time you connect to snap-offline.html you have to stay connected for like an hour while it downloads everything, or it doesn't work.  Someone said that was because of help files and we should find a way to deprioritize those.

cycomachead:
> The current manifest works fine, except that the first time you connect to snap-offline.html you have to stay connected for like an hour while it downloads everything, or it doesn't work. Someone said that was because of help files and we should find a way to deprioritize those.

Last I tried, I couldn't get it to work, but I was going to try again, to see if anything's changed, and I don't think the file is online anymore?

brianharvey:
Hmm!  manifest.mf is there, but I can't find snap-offline.html.  And I can't find it on either of my Macs either.  :-(  I think the list of files in manifest.mf is slightly out of date, too.

PR #900, opened 2015-07-31
Fix cache manifests

kach:
This PR uses HTML5 cache manifests to selectively cache Snap_!_ as follows:
1. When the browser visits `http://.../snap.html` for the first time, it reads `./manifest.appcache` and saves all files listed there to disk.
2. When the browser visits `http://.../snap.html` later, it first tries to download `./manifest.appcache`.
   - If it fails, the browser must be offline so it just displays the cached files.
   - If it loads and is identical to the saved copy, it displays the cached files.
   - If it loads and is different, proceeds as in step (1).

The benefits of this are that you can use Snap_!_ offline with no trouble, and that if you're online, it will generally load much faster because it only downloads the manifest, not the whole application.

The drawback is that you'll have to update the manifest with a nonce (usually a timestamp) everytime you push an update. I think I can automate this with git hooks if it's an issue.

Cheers!


brianharvey:
yay! +1

jmoenig:
Thanks! I'll pull this next

cycomachead:
> The drawback is that you'll have to update the manifest with a nonce (usually a timestamp) everytime you push an update. I think I can automate this with git hooks if it's an issue.

PLEAS make sure we are careful with this. Automation is the way to go for this, IMO, and I'd also be willing to help.

jmoenig:
Okay, we need to talk about this.

jmoenig:
Automation may have its benefits, but also downsides I'd like to avoid. In general I really don't much care about this feature, I'm using Snap offline myself all the time, but I just open a downloaded version...

Hardmath123:
You already update the history file with changes each time you push, so this is just another thing to do. Alternatively, if you update the copy on s.b.e with a git push, you can write a simple script in `.git/hooks/pre-push` that updates the manifest timestamp automagically before pushing to s.b.e.

cycomachead:
> Okay, we need to talk about this.

This is why I didn't make it sound so simple in #894. This PR is the proper fix, but without being careful we risk exacerbating the existing set of caching problems that users already encounter.

> Automation may have its benefits, but also downsides I'd like to avoid. In general I really don't much care about this feature, I'm using Snap offline myself all the time, but I just open a downloaded version...

The nice thing here, is that IIRC, the app cache won't have the existing security restrictions that a current offline version has. (Unless you spin up a small localhost server.)


> Alternatively, if you update the copy on s.b.e with a git push, you can write a simple script in .git/hooks/pre-push that updates the manifest timestamp automagically before pushing to s.b.e.

A `pre-push` would definitely work for this case, though a pre-push won't be fired on events that happen through Github (like merging a PR).

Personally, I vote for using a Github web hook, and the "Releases" features, which are based of Git tags, but also add a nice UI and some cool features, like a nice page of history that includes change logs and zip files of each release, and allows doing things like "latest" and "pre-release", and allows versions on s.b.e to be easily tied to a particular string of commits which makes debugging easier. 

I can write an issue about this, but it's actually not too hard. It can be done either from the GUI or the command line. :)

liam4:
+1 sounds cool.

bb010g:
bump

:+1: for @cycomachead's suggestions. They all look good to me.

PR #904, opened 2015-08-04
Correct and improve Polish translation of the "When I am" hat block.

bl-nero:


jmoenig:
Thanks! I'll pull this later this week and put in online before the Scratch conference in Amsterdam begins.

PR #908, opened 2015-08-14
Split "Variables" category

tjvr:
This splits the "Variables" category in the palette into three new ones: Variables, Lists, and Other.

screenshot 2015-08-14 12 03 50

The Other category shows every custom block, even those which appear under separate categories. When we clone ScratchX's JS extension support, we can also include extension blocks under the "other" category.

The "other" category should probably be renamed "More Blocks" or "Custom Blocks". That can be a separate PR.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -931,9 +931,7 @@ IDE_Morph.prototype.createCategories = function () {
     }
 
     SpriteMorph.prototype.categories.forEach(function (cat) {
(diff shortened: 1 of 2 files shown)


cycomachead:
:+1: 

Yes! And this makes discovery of custom blocks much nicer. :smile:

liam4:
Sounds cool to me!

PR #909, opened 2015-08-15
Add a Ceiling to the monadic math-block-thing

cycomachead:
(Does that block have a name?)

This is good for two reasons:
* it just makes sense to have floor and ceiling together
* It's a minor thing for scratch 2 parity.
* it doesn't clutter anything. :grin:

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -1149,6 +1149,7 @@ SyntaxElementMorph.prototype.labelPart = function (spec) {
                 false,
                 {
                     abs : ['abs'],
+                    ceiling : ['ceiling'],
                     floor : ['floor'],
                     sqrt : ['sqrt'],
                     sin : ['sin'],
--- a/threads.js
+++ b/threads.js
@@ -2115,6 +2115,9 @@ Process.prototype.reportMonadic = function (fname, n) {
     case 'abs':
         result = Math.abs(x);
         break;
+    case 'ceiling':
+        result = Math.ceil(x);
+        break;
     case 'floor':
         result = Math.floor(x);
         break;


jmoenig:
Yeah. The reason I didn't include ceiling in the original function list was that users can easily make their own ceiling block by adding 1 to the floor of n...

brianharvey:
Not quite.  You have to check if it's an integer first.  But still easily written by users.

jmoenig:
Good point, Brian. Okay, I'll pull it together with the next batch of changes. (I've just updated s.b.e with the new Croatian translation, so this will be pulled and integrated next time along with some other fixes).

brianharvey:
Wait, what?  I wasn't suggesting pulling this; I was suggesting a new library.  I think the principle still is that stuff users can write shouldn't be in the core language -- even more so now that we have JSFunc.

jmoenig:
Oh... sorry for misunderstanding you. Yes, I agree, a library would be nice for this function.

cycomachead:
> Yes, I agree, a library would be nice for this function.

Except we still have a problem with the discovery of libraries. Though I fully support the idea of a math library for other functions, especially now that new versions of JS have some cool stuff! Even if we do have a math library, I don't see a reason to split up floor and ceiling. It's like implementing `sin` and `cos` and leaving `tan` out because you only "need" two of the three. 

> I think the principle still is that stuff users can write shouldn't be in the core language -- even more so now that we have JSFunc.

I'm confused by this somewhat. I know we have all the stuff from Scratch that we inherited which, admittedly, is a slightly different argument, bit there are many things in Snap<i>!</i> that users _can_ implement themselves. But the main reason for doing this was the first two bullets I listed above...

jmoenig:
I'm really fine either way. I don't mind pulling this. OTOH I've never needed the ceiling function in any of my own Snap or Scratch projects. The good thing about this pull-request is that it only adds an entry to an already existing drop-down menu, not a whole new block. Of course we'll get a zillion translation updates after this...

cycomachead:
> The good thing about this pull-request is that it only adds an entry to an already existing drop-down menu, not a whole new block.

Yeah, I get the not wanting to add new blocks even if they're useful (like #889, ha)

> Of course we'll get a zillion translation updates after this...

I consider that a good sign of a healthy community! :D

jmoenig:
Okay, who wants to decide? :-)

brianharvey:
Oh, fine, pull it.  :-)

jmoenig:
I'll pull this next up. Thanks!

PR #910, opened 2015-08-16
Add Support for App Icon's.

cycomachead:
This makes @technoboy10 's work mergable, and also adds the ability to have a tab bar icon in the new version of Safari.

(Also I disabled the cache manifest fix pending @Hardmath123 's PR and discussion.)


cycomachead:
Bump @brianharvey and @jmoenig

jmoenig:
I need to know more about this. Why is all this cade special cased for Safari in the formerly simple and straight-forward html? Can't say I like that...

cycomachead:
So, Connor's original work (years ago) shows the Alonzo logo when you load the Snap<i>!</i> to the home screen on an iOS device. There's an icon and a startup Snap<i>!</i> logo for when you open the page to have the Snap<i>!</i> logo while the site is loading, more like an app.

I also added the ability for "pinned tabs" in Safari 9 -- which are pretty fun. Now, you can drag the Snap<i>!</i> tab over to the left and you get a little yellow lambda as the icon. 

None of this work is specific to Safari except for the last bit, which was a one line thing... Android is supposed to read the apple-icon properties, but I need to verify that. Those can probably be tweaked to be more generic, but they work the same way.

PR #929, opened 2015-09-11
Delete those files with the special names.

an-OK-squirrel:
I couldn't find the blocks in the editor but this should still work.


brianharvey:
The blocks in question are tools -- you get them with File>Import Tools.  And, it's not that I don't know how to delete files, I just don't want to do something like that without checking with Jens first.

cycomachead:
@brianharvey These are the old versions of the tools blocks and are no longer needed in the repo, but should potentially still live on s.b.e. In reality though, I don't see many people clicking on help for those blocks in older projects.

an-OK-squirrel:
Okay.

jmoenig:
I *think* this has been taken care of by #944

jmoenig:
Can this be closed? I think so...

brianharvey:
Yes this has been fixed.

jmoenig:
Thanks!

PR #934, opened 2015-09-17
updated Catalan translation

bromagosa:
A long due update! I've added all new strings plus fixed some old translations that were not clear enough.


jmoenig:
Thank you!! I'll pull this and integrate it along with the next batch of changes.

PR #938, opened 2015-09-19
Improve Polish translations of the "stop..." blocks

bl-nero:


jmoenig:
Thank you!! I'll pull this and integrate it along with the next batch of changes.

PR #954, opened 2015-09-26
Fix input sliders

djdolphin:
This prevents input sliders from behaving strangely when the input’s
value is -25, 0, or 25.

Usually they look like this:
screen shot 2015-09-26 at 6 15 47 pm
But at certain values the range gets messed up:
screen shot 2015-09-26 at 6 16 13 pm
screen shot 2015-09-26 at 6 16 41 pm
screen shot 2015-09-26 at 6 17 09 pm (this one's only 1 smaller than usual)

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -5749,9 +5749,9 @@ SliderMorph.uber = CircleBoxMorph.prototype;
 
 function SliderMorph(start, stop, value, size, orientation, color) {
     this.init(
-        start || 1,
-        stop || 100,
-        value || 50,
+        typeof start === 'number' ? start : 1,
+        typeof stop === 'number' ? stop : 100,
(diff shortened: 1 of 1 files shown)


morphic.js
```diff
@@ -5749,9 +5749,9 @@ SliderMorph.uber = CircleBoxMorph.prototype;
 
 function SliderMorph(start, stop, value, size, orientation, color) {
     this.init(
-        start || 1,
-        stop || 100,
-        value || 50,
+        typeof start === 'number' ? start : 1,
```
jmoenig: good catch! How about

```javascript
+start || 1,
```

would that work?

djdolphin: That still wouldn't work because +0 || 1 === 1, and start will still default to 1 when when it should be 0.

PR #956, opened 2015-09-29
Prevent anti-aliasing when merging canvases in paint editor

Hardmath123:
Now, when you un-zoom and zoom repeatedly, lines don't turn into fuzzy ghosts. Vertical and horizontal lines are pure n-pixel-wide and of $COLOR instead of n+1-pixel-wide and faded.

See also: http://stackoverflow.com/questions/195262/can-i-turn-off-antialiasing-on-an-html-canvas-element.

### DIFF
--- a/paint.js
+++ b/paint.js
@@ -57,6 +57,7 @@
     June 4 - tweaks (Jens)
     Aug 24 - floodfill alpha-integer issue (Kartik)
     Sep 29 - tweaks (Jens)
+    Sep 28 [of the following year :)] - Try to prevent antialiasing (Kartik)
  */
 
 /*global Point, Rectangle, DialogBoxMorph, fontHeight, AlignmentMorph,
@@ -595,6 +596,7 @@ PaintCanvasMorph.prototype.scale = function (x, y) {
         this.rotationCenter.y
     );
     c.getContext("2d").scale(1 + x, 1 + y);
+    this.disableSmoothing(c.getContext("2d"));
     c.getContext("2d").drawImage(
         this.paper,
         -this.rotationCenter.x,
@@ -622,11 +624,21 @@ PaintCanvasMorph.prototype.undo = function () {
     }
 };
 
+PaintCanvasMorph.prototype.disableSmoothing = function(ctx) {
+    ctx['imageSmoothingEnabled'] = false;       /* standard */
+    ctx['mozImageSmoothingEnabled'] = false;    /* Firefox */
+    ctx['oImageSmoothingEnabled'] = false;      /* Opera */
+    ctx['webkitImageSmoothingEnabled'] = false; /* Safari */
+    ctx['msImageSmoothingEnabled'] = false;     /* IE */
+};
+
 PaintCanvasMorph.prototype.merge = function (a, b) {
+    this.disableSmoothing(b.getContext("2d"));
     b.getContext("2d").drawImage(a, 0, 0);
 };
 
 PaintCanvasMorph.prototype.centermerge = function (a, b) {
+    this.disableSmoothing(b.getContext("2d"));
     b.getContext("2d").drawImage(
         a,
         (b.width - a.width) / 2,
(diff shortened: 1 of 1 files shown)


brianharvey:
Thanks!

paint.js
```diff
@@ -622,11 +624,21 @@ PaintCanvasMorph.prototype.undo = function () {
     }
 };
 
+PaintCanvasMorph.prototype.disableSmoothing = function(ctx) {
```
jmoenig: I think we should just use the standard setting and wait for all the browsers to catch up...

brianharvey: Noooooo.......  I want this fixed!

jmoenig: Well, and I don't want any workarounds that build up layers of cruft that we have to keep in the back of our minds. disableSmoothing should be fine

jmoenig:
How is this important? I mean, it's not about "zooming" but about "growing and shrinking" a bitmap. Even with smoothing disabled there is going to be loss of data. @bromagosa's student is working on a vector editor that will handle "zooming" perfectly. So, I agree it's nice to suppress smoothing in certain situations, and it's nice to add this, but, frankly, it's just a flag, and I wouldn't want to build a special case aware feature around this...

brianharvey:
I just revised my Mastermind program last night.  It uses six colored square costumes and I wanted to make them all smaller, but still all the same size.  So I shrunk one of them, then copied it and tried to floodfill it with the next color.  But this didn't work, because the fuzzy border was actually a different color, so I ended up with an orange square with a fuzzy red halo.  And it just gets worse from there.  You should be able to make a square in the paint editor!

paint.js
jmoenig: minor quibble:

```javascript
ctx['imageSmoothingEnabled'] = false;
```

would be complained about by old.jslint.com:

```javascript
ctx.imageSmoothingEnabled = false;
```

jmoenig:
Agreed. I always use setSize for this myself, but I see how this can be annoying, and I'm open to blessing and pulling this fix. I just don't want any browser specific stuff in here. Paint.js already is an exception to code quality in many ways, I don't want to make it any worse to maintain and debug.

brianharvey:
Yes, but I want it to work for everyone, today.  :-(

paint.js
cycomachead: > Well, and I don't want any workarounds that build up layers of cruft that we have to keep in the back of our minds.

We can't really avoid browser workarounds... If we're concerned about matinence we could just use a standard comment format or a docs file that lists these things...

jmoenig:
So, Brian wants me to pull this. Can't really say it's "blessed" by me, since I'll have to tweak it right after pulling it... :-(

jmoenig:
Okay, so this isn't exactly solving the issue, right? If we really want to prevent blurring we'd have to use something like Bresenham's algorithm to draw lines directly into the image data... But *then* we'd also have write our own graphics framework to re-draw those lines when we scale or rotate them.

What's worse, now, instead of smoothing we're *loosing* data and the shape of things changes. I think this "fix" changes the paint editor to the worse. Let's face it, what people *want* is vectors, but *if* they're using raster graphics, they probably also want anti-aliasing and automatic smoothing. Now scaling really messes up things like lines and free-form drawings.

Next time just because a teacher complains doesn't mean we should change something to the worse.

Can we *please* revert this?!!

jmoenig:
No way! Just, no. I'm gonna revert this right now. This won't fly. We're totally breaking the paint editor with this. Sorry!

bromagosa:
@jmoenig yesterday we managed to successfully save the contents of the new vector editor, we tried some transformations (scale blocks, etc) and it all worked flawlessly. There are still some issues to solve and the code needs serious reviewing before being included into trunk, but I believe we'll have an at least _functional_ vector editor very soon.

jmoenig:
That's excellent, @bromagosa ! I'm looking forward to integrating a vector editor. In the meantime let's not break the existing bitmap editor :-)

PR #968, opened 2015-10-06
first approach to a stage layers API

bromagosa:
The API is really simple, here's how it works:

A stage layer is a simple JS object that has a name and a canvas. The stage now allows you to:
- add a new layer: `stage.addLayer(aCanvas, 'some layer')`
- find a layer by its name: `stage.getLayer('some layer')`
- remove a layer by its name: `stage.removeLayer('some layer')`

I decided that `StageMorph.prototype.penTrails()` should return the `'penTrails'` layer canvas instead of the whole object, this way we keep backwards compatibility with all old projects. I've tested several of them and found no issue, they all load properly.

A simple test that adds a layer:

```
stage.addLayer(newCanvas(stage.dimensions), 'test');

var layer = stage.getLayer('test'),
    context = layer.canvas.getContext('2d');

context.fillRect(20,20,150,100);
stage.changed();
```

This should draw a black rectangle at the top left corner of the stage. It won't be cleared along with the pen trails and it won't be affected by stage costumes. It'll also behave properly to resizing, exporting as a picture or being drawn in the project thumbnail.


jmoenig:
Thanks, Bernat! Looks pretty much exactly what we talked about :-)
Still, I'll be leaving this open for another while, as I'm fidgeting with other matters that are under construction at the moment. But then we'll integrate it along with the first cut at the new object system.

objects.js
```diff
@@ -4875,6 +4894,41 @@ StageMorph.prototype.penTrailsMorph = function () {
...
+        name: name };
+
+    if (!this.getLayer(name)) {
+        this.layers.push(layer)
+    }
+
+    return layer;
+};
+
+StageMorph.prototype.getLayer = function(name) {
```
jmoenig: is there any particular reason for not using ```detect()``` here (and also further down)? I know that we've recently re-"introduced" the old imperative ```for``` loop to Morphic, but that was really just a performance optimization for two (I think) very special cases in which *every* Morph in the whole world gets enumerated at every mouse event. In these cases here I'm not assuming there to be hundreds or thousands of layers, so I'd suggest just using HOFs as we usually do :-)

(no need to edit or change anything just now, @bromagosa , I only wanted to log this reminder to myself here)!

bromagosa: Actually I thought the new policy was to use ``for`` iterators, but I'm happy to hear it's not! :)

bromagosa:
Hi @jmoenig! Should I update this one using _detect_?

jmoenig:
Hi Bernat, yes, I think that would make it better to maintain and easier to extend in the future. Thanks!

PR #971, opened 2015-10-07
Refactor Loading Resources (Sounds, Libraries, etc)

cycomachead:
~~This is a work in progress so far, but please leave feedback!~~
This is now done.

Functionally, there will be little change here. All this is doing
is giving a needed refactoring to loading of media in `gui.js`.

* The big change is that now item names in menus won't need to be exactly
the file names they are pointing to, and can be added to translation
dictionaries if people choose to.
* Fix a typo in the current `LIBRARIES` file that makes the leap motion library
not loadable.
* Create a consistent structure for parsing that file.
* All `LIBRARIES` to have help text for each library name, though none has been
written, yet.
* This refactoring will help make Snap! much more "portable". Hosting on edX
doesn't allow us to have sub-folders, so by refactoring content, there will
(eventually) be 1 method to override if others want to change things from
subfolders to another structure.

TODO:

- [x] Make the display of sub-menu headers consistent
		(Some plural, and the can be adjusted to be localized more easily)
- [x] Create a method to for loading a folder and filename
- [x] Refactor `getMediaList`
- [x] Cleanup the sort methods used
- [x] Write some documentation of the resource file format.
- [x] DRY code for creating menu items... Make the lambda a real function....
```js
            names.forEach(function (image) {
                libMenu.addItem(
                    image.name,
                    function () {loadCostume(image.file); },
                    image.help
                );
            });
```

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -2466,6 +2466,27 @@ IDE_Morph.prototype.projectMenu = function () {
                 'Costumes' : 'Backgrounds',
         shiftClicked = (world.currentKey === 16);
 
+    // Utility for creating Costumes, etc menus.
+    // loadFunction takes in two parameters: a file URL, and a canonical name
+    function createMediaMenu (mediaType, loadFunction) {
+        return function () {
+            var names = this.getMediaList(mediaType),
+                mediaMenu = new MenuMorph(
+                    myself,
+                    localize('Import') + ' ' + localize(mediaType)
+                );
+
+            names.forEach(function (item) {
+                mediaMenu.addItem(
+                    item.name,
+                    function () {loadFunction(item.file, item.name); },
+                    item.help
+                );
+            });
+            mediaMenu.popup(world, pos);
+        }
+    }
+
     menu = new MenuMorph(this);
     menu.addItem('Project notes...', 'editProjectNotes');
     menu.addLine();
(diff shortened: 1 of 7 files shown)


gui.js
```diff
@@ -4472,22 +4511,6 @@ IDE_Morph.prototype.getURL = function (url) {
     }
 };
 
-IDE_Morph.prototype.getURLsbeOrRelative = function (url) {
```
jmoenig: Wait! You're removing @brianharvey 's function! ;-)

cycomachead: I hope he doesn't mind! After we did the whole `baseURL` thing is was no longer loading from s.b.e so it wasn't even doing it's job. (Also technically, though there was a big argument in July 2014 about this, the fallback mechanism part wasn't working reliably.)

However, in my refactoring, I've realized that we don't need the `baseURL` anyway... We just want to load everything relative to the `snap.html` file anyway, right? 
That said, if we do need it, I now only need to change 1 line of code, not 5. ;-)

jmoenig: Yes, we want to load everything relative to ```snap.html```, that's right. However, I'm not sure what exactly 'loading' means, because XHR2 doesn't support the ```file:``` protocol (I think).

cycomachead: Was the use of this function when loading Snap! In the browser from an HTML with a `file://` url?
In that case, the `baseURL` solution still doesn't work, but it's using the same URL. We would actually need to keep a hardcoded reference to s.b.e/snapsource

Do we want to support loading from a `file://` URL? There are plenty of other exceptions that simply can't be handled. I feel like a desktop editor is a better way to support this... Besides if you have a local snap.html download, you have all the files already available via the import function, or drag and drop.

cycomachead:
@jmoenig Ok, this is ready! :)

My commits contain most of the details for what I did, but essentially, the current behavior of loading libraries, etc. is unchanged to the user. However, for all the files, I've added a much nicer human readable name. Backgrounds, Sounds, Costumes, and Examples now have folders which

cycomachead:
Bump on the done-ness of this! :)

This refactoring will be _really_ helpful to using the latest Snap<i>!</i> in our edX courses. (Currently we have an old version of Snap<i>!</i> and it doesn't even load everything correctly. :/)

gui.js
jmoenig: I don't much mind supporting ```file:```, if you've tested this and it works for you I'll pull it next up. Thanks!

cycomachead: Right now, all the loading is still XHR on a `file://` so it won't work. 
It's been broken since #814 

However, everyone who's downloaded Snap<i>!</i> locally will have all the necessary files, so even if they were offline the menus wouldn't work. I don't think there's much use here to have the fallback. (All the fallback was supposed to do was just attempt to load from s.b.e/X if something failed.)

PR #978, opened 2015-10-18
Fix backspace & tab keys in IE

### LINKED ISSUE #569: bug using backspace in Internet Explorer
When using IE 11 on Windows 7, if you press backspace (such as when logging into and correcting your cloud account name or password), the program tries to exit by giving you the "leaving page" question.  For this reason, I cannot recommend using IE to my students.

bl-nero:
Fixes #569 

Tested on:

* Chrome 46.0.2490.71 (Mac)
* Firefox 41.0.1 (Mac)
* Opera 32.0 (Mac)
* IE 11 (Win 8.1)

(Note: pasting doesn't work outside Chrome, but it looks like it has never worked anyway.)

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -10224,14 +10224,12 @@ WorldMorph.prototype.initVirtualKeyboard = function () {
                 myself.keyboardReceiver.processKeyDown(event);
             }
             // supress backspace override
-            if (event.keyIdentifier === 'U+0008' ||
-                    event.keyIdentifier === 'Backspace') {
+            if (event.keyCode === 8) {
                 event.preventDefault();
             }
             // supress tab override and make sure tab gets
             // received by all browsers
-            if (event.keyIdentifier === 'U+0009' ||
-                    event.keyIdentifier === 'Tab') {
+            if (event.keyCode === 9) {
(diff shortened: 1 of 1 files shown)


morphic.js
```diff
@@ -10224,14 +10224,12 @@ WorldMorph.prototype.initVirtualKeyboard = function () {
                 myself.keyboardReceiver.processKeyDown(event);
             }
             // supress backspace override
-            if (event.keyIdentifier === 'U+0008' ||
-                    event.keyIdentifier === 'Backspace') {
+            if (event.keyCode === 8) {
```
jmoenig: really? *that* solves it? Fantastic, thank you!

jmoenig:
Thank you so much! I'm travelling this week, not sure whether or when I'll have time to work on Snap, but I *will* pull this next thing!

morphic.js
bl-nero: I couldn't believe it either, because it's been a while since I touched such low-level stuff, but it looks like the support for `key` and `keyIdentifier` properties is vastly inconsistent among browsers. Even though they were supposed to replace `keyCode`, the latter is still the simplest way to solve it.

Issue #966, opened 2015-10-03
Meta Issue: Chrome Export Issues Update


cycomachead:
Ok, I think we should have a new discussion about fixing the exporting issues on Chrome -- that is, clicking "Export Project..." with > 2MB of data leads to the "Aw, Snap!" view, causing lost work.

This currently is also the case with the "Export Summary..." feature.

* We should do something about the crash on existing browsers. (We really should have done something long ago...) See #537 
* Chrome 47 - current dev channel (Chrome Canary version doesn't crash!) Woohoo!! However, now there's a bigger issue: **When using long data URIs in Chrome 47 there is no "Save As..." function on even though the page loads** Some things that generate massive URIs, like the all script pics feature, no longer crash Chrome, but end up displaying only a blank page. 

The other good news is that since summer I've been messing around with the library FileSaver.js which gives a good fallback using Blob URIs and gracefully reverts back to existing methods when they don't work. While I know using a library generally isn't acceptable, FileSaver implements 1 function `saveAs` which handles downloading files. So far I don't see much point in pulling code out of FileSaver.js because they've gone through the hard work of identifying and working around browser bugs -- so redoing things ourselves seems to be a waste of effort IMO. In my hacky test, I overwrote `window.open()` with a call to `saveAs` and things worked just fine.

FileSaver: https://github.com/eligrey/FileSaver.js

For Reference:
#433 #726 #749 #883 #433 #964 (would likely be fixed by this, but haven't confirmed) #123

cycomachead:
@jmoenig @brianharvey 

Thoughts?

It has occurred to me that having a consistent `saveAs()` method could also benefit other uses of Snap! such as the iPad app, making it easier to customize grabbing files, or doing anything special.

jmoenig:
My hunch is just taking the block uri part would solve this issue, and we should just do that. This - like many other related issues - is particular to Chrome. I don't want to change the design of Snap because Google decides to suck :-)

cycomachead:
The thing is though, that's basically what the library does... Just that there's a lot of bugs still, even in other browsers, of course. 

I mean, I can totally just cut out parts of the library, but all I'm doing is making something that's harder to maintain and doesn't handle the edge cases...

jmoenig:
Ah, I'm not sure about the "hard to maintain" part. Introducing dependencies on 3rd party software has huge downsides in the long run. Personally I prefer to roll our own...

cycomachead:
[...]

PR #983, opened 2015-10-27
Use Blob API to Save Files (to Disk)

cycomachead:
This is following the discussion on #966 this adds FileSaver as a library
and creates a new function `IDE_Morph::saveFileAs` that is designed to be the
one-and-only way of getting data from Snap<i>!</i> to a user's computer.

I still have lots of work to do, but I wanted to share the code so others could
test if desired. Right now, I've only been messing with the export project
function.

TODO:
- [ ] Lots of testing! :smile: 
- [x] Refactor conditionals in `saveFileAs`
- [x] Write Error messages
- [x] Handle `isCatchingErrors`
- [x] verify file names
- [x] Use blob URLs for new windows? 
- [x] fix "Export Project" to Open and not download ?


jmoenig:
Yay! Thanks, Michael!

jmoenig:
So, this isn't ready for pulling yet, right?

cycomachead:
Nope not yet!

I did deploy this branch to http://snap.berkley.edu/snapsource/michael/

I haven't confirmed I haven't broken anything yet, but the "Export project..." function is working well. We used it last night to save a student's project which had 10+MB of media. Worked great in Chrome!

cycomachead:
@jmoenig OK, I'm getting fairly close with this. The hard work seems to be done. And you should be able to play with it here:
 http://snap.berkley.edu/snapsource/michael/

Right now:
*  Images and the project summary open in a new tab by default, XML is downloaded.
    * This is a change, but I think it's how most users / students would work. Even images would likely be saved to disk by users, but they at least have something useful to see.
* All downloaded items (including potentially images) have a file name that can be specified. Safari doesn't respect this name, but chrome does. Currently I'm usually just using the project name, but we could be more creative / specific if desired.
* For reference: `saveFileAs` is now the function that handles `encodeURIComponent` for all of exporting in Snap<em>!</em> because it's not needed with the Blob APIs, and it's better to pass around unencoded data, and only encode when needed. This doesn't really change anything functionally, but will be worth knowing if we add any export options.


Now I have some thoughts / questions:
* When is the whole `isCatchingErrors` thing necessary?
```js
    if (Process.prototype.isCatchingErrors) {
        try {
            [large block]
        } catch (err) {
            [show error]
        }
    } else {
        [same large block]
    }
```
Wouldn't all those cases be better written as:
```js
try {
    [large block]
} catch (err) {
    if (Process.prototype.isCatchingErrors) {
         [show message]
    } else {
        [throw error]
   }
}
```
I'm noticing a lot of code duplication and I'm making changes to the same thing multiple times, which is really error prone. 

* What should the behavior be for opening in a new window? Getting this consistent would actually make the code a lot easier. (Also @brianharvey)

cycomachead:
Also, "save to disk" and "export project" a subtlely different in how this check for a project's name. Shouldn't they be consistent?

Making them consistent will be better since the code for both those two functions is nearly identical.

jmoenig:
Excellent news, Michael! I'll play with your version and let you know how it feels. Thank you!

You're right that we could probably clean up some of the "grown" code. Feel free to suggest any (more) simplifications that strike you.

cycomachead:
@jmoenig @brianharvey 

I've merged in changes with 4.0.3, and updated my branch on snapsource/michael/ so please test it out! :)

So far, it's all done, with the exception of writing error messages...

I also have the question of what should file names be? 
Current names:
* <projectName>.xml -  (projects)
* `ide.projetName + ' ' + localize('scrip pic')`.png
* `ide.projetName + ' ' + localize('comment pic')`.png
* <projectName>.html, <projectName>.png   - (summary views)
* `this.name`.svg or `this.name`.png - costumes
* `localize('stage').png - for the stage pic
* `this.spec`.xml - exporting a single block
* `ide.projectName + ' ' + localize('blocks')`.xml - exporting custom blocks
*  `sprite.name`.xml (exporting a sprite)

cycomachead:
@jmoenig How should I acess the `IDE_Morph` from within the Blocks Export Dialog?

Currently I have `ide = this.world().children[0];` but that seems ugly... `ide` isn't a property of the editor or anything...

cycomachead:
Error messages:
    'download to disk text':
        'This item could not be opened in a new tab.\n' +
        'It has been saved to your browser\'s downloads folder.',
    'unable to export text':
        'This item could not be exported from Snap!.\n' +
        'It\'s like that your project may contain a lot of media ' +
        '(sounds and images) or that you are using an older browser.' +
        'Please try using a recent version of Chrome, Firefox, or Safari.'

jmoenig:
Thanks!
Is there any way in which we can force the browser to pop up an OS-native file dialog and let the user specify their own file names? (I guess the answer is that on some browsers this just works and on others it just doesn't, right?).
Can you show me where you access the IDE from the block export dialog, I didn't find it by cursorily eye-scanning the code...

cycomachead:
> Is there any way in which we can force the browser to pop up an OS-native file dialog and let the user specify their own file names? (I guess the answer is that on some browsers this just works and on others it just doesn't, right?).

Directly from the browser, with no junk like Flash... We can't do anything. :( Saving directly to disk currently relies primarily on the `a[download` attribute from still, but in combination with new Blob APIs which are semi-file-like primitives.

I should have added, that in the case where a new window / tab is opened (the default for new images), the normal save as dialogs work as expected, provided the user actually selects Save As, of course.

For the line of code, I think I hadn't pushed some commits, here it is:
https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/983/files#diff-afb7a00b1866ca6d6477b267bc8c644cR3372

cycomachead:
This is almost ready, I have a few fixes to make for exporting script pics for the block editor and for a couple file names, but I'd welcome other's input if they've tested.

Testing Plans:
Images should be displayed in a new tab unless they are large.
* blocks.js:1855 - Script pic with Result
* blocks.js:2325 - Script pic (normal)
* blocks.js:5420 - Script pic (block editor)
* blocks.js:11099 - Comment Pic
* byob.js:770 - Script pic (custom block script)
* gui.js:3151 - ALL Scripts pic
* gui.js:5908 - Stage Picture (icon context menu)
* gui.js:6277 - export costume
* objects.js:5740 - stage pic (stage context menu)

Saving an XML should download to disk
* Note Safari is stupid and will just display the darned file.
* byob.js:798 - export single block definition (?)
* byob.js:3383 - export blocks dialog
* gui.js:3023 - export project (normal mode)
* gui.js:3033 - export project (dev mode)
* gui.js:3102 - export sprite
* gui.js:4524 - export media (normal)
* gui.js:4534 - export media (dev)
* gui.js:4553 - export NO media (normal)
* gui.js:4563 - export NO media (dev)
* gui.js:4583 - export cloud data (normal)
* gui.js:4595 - export cloud data (dev)

Additional Cases:
* gui.js:3428 - Project summary HTML: New Tab
* objects.js:7629 - exporting a variable

Browsers:

- [ ] IE 11
- [ ] IE Edge
- [ ] Chrome (Windows)
- [ ] Firefox (Windows)
- [ ] Chrome (OS X)
- [ ] Safari (OS X)
- [ ] Firefox (OS X)
- [ ] Chrome (Linux)
- [ ] Firefox (Linux)

cycomachead:
Ok! @jmoenig @brianharvey I handled the last few bugs and clicked through all the export options in Chrome on the Mac, and through many of them on both Safari and Firefox. Phew. I think things are good now. 

Two minor things for the code:
* there's a fair bit of `projectName || localize('Untitled')`. It's harmless, but a bit ugly imo. I don't know if it really matters since it's short.
* Referencing `ide` from within dialogs like the export blocks one is a bit weird... The current solution seems to be `this.world().children[0]`. It works, but I don't really like it.

jmoenig:
Thanks, this is great!
Why do we need to reference the IDE from within dialogs? Usually those dialogs that need to reference the IDE already have a property for it, or it gets passed in through a callback. Is there any new case now?

cycomachead:
> Why do we need to reference the IDE from within dialogs?

The method to export content out of Snap<em>!</em> is `IDE_Morph.prototype.saveFileAs` (also `saveCanvasAs` and `saveXMLAs`). Since they aren't global, then everything must access the IDE. It seems like the IDE is the best place the functions, but it does make a couple cases a bit wierd.

cycomachead:
There as 3 cases where I needed to do that:
* Export Blocks Dialog
* Export Script pic of custom block (background context menu)
* Export script pic of custom block (script context menu)

jmoenig:
@cycomachead is there a current version of your branch online somewhere? I've been getting funny messages in Chrome lately, when trying to save XML files to disk directly, and I wanted to check whether it's better in your branch. Thanks!

cycomachead:
I think this was in an email, but just for everyone:

https://snap.berkeley.edu/michael

jmoenig:
Thanks, Michael!

cycomachead:
Ok, this is mergable! :smile: 

When merging there was your commented out code that I removed, but it code be added back in:

```
/* alternative download mechanism, commented out in favor of blob-uris
IDE_Morph.prototype.saveProjectToDisk = function () {
    var data;
    if (Process.prototype.isCatchingErrors) {
        try {
            data = this.serializer.serialize(this.stage);
        } catch (err) {
            this.showMessage('Serialization failed: ' + err);
        }
    } else {
        data = this.serializer.serialize(this.stage);
    }
    this.download(data, this.projectName);
};

IDE_Morph.prototype.download = function (data, fileName, mime, suffix) {
    var link = document.createElement('a'),
        myself = this;

    mime = mime || 'data:text/xml';
    fileName = fileName || 'download';
    suffix = suffix || 'xml';

    function saveToDisk() {
        var msg =  myself.showMessage('Downloading to disk...');
        link.setAttribute('href', mime + ',' + data);
        link.setAttribute('download', fileName + '.' + suffix);
        document.body.appendChild(link);
        link.click();
        document.body.removeChild(link);
        msg.destroy();
    }

    if (Process.prototype.isCatchingErrors) {
        try {
            saveToDisk();
        } catch (err) {
            this.showMessage('Saving failed: ' + err);
        }
    } else {
        saveToDisk();
    }
};
*/
```

jmoenig:
Thanks, Michael. We don't need that commented out code, I think, so I'll just leave it out :-)

PR #995, opened 2015-11-02
Bulgarian translation of strings lang-bg.js

ivanistheone:
- Used lang-ru.js as starting point, and translated all, but still some strings appear in English
  - Did my best to use Bulgarian words, but left certain terms like "thread safe" in English
  - Tested with my friend's 8 year old and he seems to understand, so count that as successful user testing

### DIFF
--- a/lang-bg.js
+++ b/lang-bg.js
@@ -1,0 +1,1125 @@
+/*
+
+    lang-bg.js
+
+    Bulgarian translation for SNAP!
+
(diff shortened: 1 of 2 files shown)


jmoenig:
Yay!! Thank you so much @ivanistheone ! This is wonderful, I'm going to pull this with the next batch of changes and let you know when it's online. This is probably going to happen sometime later this week or early next week.

Thank you, thank you!

Issue #349, opened 2014-03-12
Snap doesn't account for Menus longer than the window height


cycomachead:
This makes it really hard to use long lists. One common-ish example, is the letters drop down menu one small screens or with zoom blocks of 1.5x+ (Common for presentations).
The other is with absurdly long lists. :)

See two screen shots.
![screen shot 2013-12-18 at 2 57 09 pm](https://f.cloud.github.com/assets/1505907/2393905/cfb68d5c-a9a1-11e3-8c7a-b1d5598e9a90.png)
![snap menu bugs 2](https://f.cloud.github.com/assets/1505907/2393906/cfb8083a-a9a1-11e3-9a98-22f58b5ee850.png)

There are two things we could do:
1. First detect usable window height...
2. We could either have options that say: "[Next Items]" and "[Previous Items]" 
 Or we could have scrollable menus. 

Personally, I like the second option better, but I don't know what's easier to implement.

jmoenig:
fixed with 922dcc279eb7a27f67767f1fabff573007b99f26

PR #1005, opened 2015-11-15
Bug Fix for Issue #349, menus longer than the window height

timurvural:
Bug Fix for issue #349.

Originally, keepWithin was called on the menu when it extended past the boundaries of its world, which simply shifted it up in the world to keep it visible. When the menu is taller than the window, however, some parts of the menu become inaccessible, as shown below.

image

I wrote a new method keepVisible, which encloses the menu in a scrollFrameMorph when it extends past the world boundary. This makes the rest of the menu accessible by scrolling. I also changed Handmorph's processMouseDown method to make sure the menu stays selected when its scrollFrameMorph is clicked on. An example of a large menu made scrollable by its scrollFrameMorph is below.

screen shot 2015-11-14 at 4 16 23 pm


jmoenig:
very cool! And very timely. Look, what I've been working on myself this past week in GP: Scrollable Menus! :-)

scrollable_menu_gp

Looks just like yours :-)

So, it appears like you're not changing the internal structure of Menus, but instead just always wrap a scroll frame around the whole menu, regardless of whether it all fits onscreen. Did I get this right from just glancing at your code? How do you handle menu labels? Do you start scrolling underneath them (so the label stays always visible), like this:

scrollable_menu_gp_2

or do you scroll everything?

Also, looking at your screenshot it seems that in your code menus are loosing their slight drop-shadows and their surrounding frames (because the surrounding scroll frame clips them). Is that right?

bb010g:
What's GP?

timurvural:
That's right - rather than making the menus themselves scrollable, I wrapped the menus inside a scroll frame. The menu only gets wrapped inside a scroll frame if it can't fit on the screen without being repositioned. 

Menu labels can be scrolled, so they are not always visible. 

Here is a different screenshot for the shadows - they are still visible, but only when they are still within the scroll frame.

screen shot 2015-11-16 at 12 01 31 am

In the screenshots you mentioned, the menu was not originally inside a frame, so there was no frame to clip. If the menu had started off already contained in a frame, however, that frame would have gotten clipped.

khotchkiss13:
Like @bb010g, I too am curious as to what GP is

jmoenig:
GP is a new general purpose blocks programming language currently under development by John Maloney, Yoshiki Ohshima and myself. It's still in early alpha, we did show it off at this year's Scratch conference in Amsterdam and at a recent VL/HCC "Blocks & Beyond" workshop in Atlanta. Here are two things to read about it: http://dspace.mit.edu/handle/1721.1/98451 and http://www.vpri.org/pdf/tr2015003_modsys.pdf and there's also an article in the German Scratch wiki: http://scratch-dach.info/wiki/GP_(Programmiersprache)

timurvural:
I've made some updates. The scroll frame is now created inside the menu rather than outside, and the menu's contents are added to the scroll frame. The scroll frame now won't clip the menu's surrounding frame or the menu's shadow. 

screen shot 2015-12-04 at 4 43 33 pm


Also, if the menu is short enough to fit on screen, the menu is simply shifted up so it stays visible.

screen shot 2015-12-04 at 4 51 26 pm


If the menu has a label, the label stays pinned to the top of the menu.

screen shot 2015-12-04 at 4 52 46 pm

jmoenig:
Nice! Thank you!

liam4:
Awesome! I was having this issue earlier when trying to look through a very long list.

khotchkiss13:
This looks awesome!

morphic.js
```diff
@@ -6694,6 +6694,7 @@ MenuMorph.prototype.init = function (target, title, environment, fontSize) {
     this.isListContents = false;
     this.hasFocus = false;
     this.selection = null;
+    this.scrollMenu=null;
```
cycomachead: Spacing, use `x = y` for `=` stuff

morphic.js
```diff
@@ -6903,18 +6904,37 @@ MenuMorph.prototype.unselectAllItems = function () {
...
     }
     if (this.items.length < 1 && !this.title) { // don't show empty menus
         return;
     }
+    this.keepVisible(world);
     world.add(this);
-    world.activeMenu = this;
     this.world = world; // optionally enable keyboard support
+    world.activeMenu = this;
+    if (this.scrollMenu != undefined && this.scrollMenu != null) {
```
cycomachead: I think this should probably just be `if (this.scrollMenu) {`

morphic.js
```diff
@@ -6903,18 +6904,37 @@ MenuMorph.prototype.unselectAllItems = function () {
...
+        pos = this.position();
+        var labelHeight = 0;
+        if (this.label != null) {
```
cycomachead: For all comparisons you should use `===` and `!==`

morphic.js
```diff
@@ -6903,18 +6904,37 @@ MenuMorph.prototype.unselectAllItems = function () {
...
+            labelHeight = this.label.height() + 2;
+        }
+        this.scrollMenu.world = world;
+        world.add(this.scrollMenu);
+        this.scrollMenu.setPosition(new Point(pos.x + 3, labelHeight + 6));
+        this.bounds.corner.y = Math.round(0.98 * world.height());
+        this.scrollMenu.setWidth(this.width() - 6);
+        this.scrollMenu.setHeight(this.bottom() - this.scrollMenu.top() - 4);
+        for (var i = this.children.length ; i > -1 ; i += -1){
```
cycomachead: Unless there's a specific need to iterate backwards, this should be `this.children.forEach()` style loop.

morphic.js
```diff
@@ -6944,6 +6964,16 @@ MenuMorph.prototype.popUpCenteredInWorld = function (world) {
     );
 };
 
+MenuMorph.prototype.keepVisible=function(world){
```
cycomachead: Update the spacing here.

morphic.js
```diff
@@ -7040,6 +7070,10 @@ MenuMorph.prototype.destroy = function () {
     if (this.hasFocus) {
         this.world.keyboardReceiver = null;
     }
+    if (this.scrollMenu != null) {
```
cycomachead: `!this.scrollMenu`

morphic.js
```diff
@@ -9566,9 +9600,10 @@ HandMorph.prototype.processMouseDown = function (event) {
             if (!contains(
                     morph.allParents(),
                     this.world.activeMenu
-                )) {
-                this.world.activeMenu.destroy();
-            } else {
+                ) && !contains(morph.allParents(), this.world.activeMenu.scrollMenu)) {
+
+                this.world.activeMenu.destroy();}
+            else {
```
cycomachead: Spacing on these lines

cycomachead:
I added a bunch of comments for code style, this looks great!

The only bug I see is that the top and bottom of the scroll bars are cut off, but I don't immediately see why.....

cycomachead:
Also, it looks like there should be a teeny bit of padding between the right edge of the menu and a scrollbar.

morphic.js
```diff
@@ -6903,18 +6904,37 @@ MenuMorph.prototype.unselectAllItems = function () {
...
         return;
     }
+    this.keepVisible(world);
     world.add(this);
-    world.activeMenu = this;
     this.world = world; // optionally enable keyboard support
+    world.activeMenu = this;
+    if (this.scrollMenu != undefined && this.scrollMenu != null) {
+        pos = this.position();
+        var labelHeight = 0;
```
cycomachead: This should be the first line of the function.

morphic.js
timurvural: In this case, I iterated backwards because members of this.children are getting removed during iteration.

timurvural:
Thanks for catching all of that! I've updated the style of the code and added a little bit of padding. It turned out the scroll button wasn't being clipped - just had to make the slider more transparent so the button could show better.

timurvural:
@jmoenig, Would you mind taking another look at this? I've fixed all the issues that were pointed out in this thread.

jmoenig:
@timurvural sure, thanks!

brianharvey:
I'm confused, was this pulled?

cycomachead:
No, but the issue was resolved by Jens building the same thing. :)

PR #1021, opened 2015-11-19
Update NL-translation

sdmeijer:
Fixed the translations of the word 'any'.


jmoenig:
Thanks! I'll pull this with the next batch of updates.

PR #1025, opened 2015-11-22
Fix indent issues

liam4:


liam4:
Please do this so I can have the 1000th commit

with <3 from liam

brianharvey:
Liam, being as I am a wannabe kid myself, I find this mildly amusing -- but it's not a good way to get Jens to respect you.  (Sorry for saying this in public, but I don't seem to have an email address for you.)

liam4:
Nobody likes jokes. :P

PR #1032, opened 2015-11-29
Make localization use JSON files and move them to a directory(work in progress)

Alwahsh:
Hi,

There is a similar PR to this one #287, but I'm not planning to depend on any external resource. I've just moved the files to a directory and made the gui take the json file. I didn't convert all language files to JSON yet but I wanted to ask for your opinion about this improvement before I write a script to do that so as to not do work in vain.

Thanks


cycomachead:
Right now, I don't think Jens wants more subfolders in the repo. See #197 and #690 

Personally, I think the current format vs JSON isn't a big difference, though losing comments is a bit sucky. Jens' call there.

However, I do think we should (for the next few years, sadly) avoid using `Promise`, at least while IE11 and Android 4.4 as still at fairly large levels. You can just use the normal `getURL` function, I think and things should be fine.

Alwahsh:
I wanted to make it easier to translate by making some tool that allows one to translate using a web interface but I guess it's not time for that yet. Thanks for the advice with getURL though :smile: . I'm closing this.

PR #1039, opened 2015-12-03
Handling of diacritics , [Alt] + key in input fields

DarDoro:
Workaround for diacritics handling in browser #4. Sequences [Ctrl] + key should be processed only when [Alt] is not pressed. 
Root of this problem is browsers javascript KeyboardEvent. Sequence [Alt] + key , generates [Alt] + [Ctrl] + key event. In "morphic.js" this is mishandled as [Ctrl] + key and prematurely disposed. To prevent this, test for "ctrlKey" should be expressed as "event.ctrlKey && (!event.altKey)".

This fix effectively disables [Ctrl] [Alt] + anything (acts as [Alt]...) in favor of proper handling of diacritics.

Works with Chrome, Opera, Internet Explorer, Edge on Windows 10.

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -4515,7 +4515,7 @@ CursorMorph.prototype.processKeyPress = function (event) {
         return null;
     }
     if (event.keyCode) { // Opera doesn't support charCode
-        if (event.ctrlKey) {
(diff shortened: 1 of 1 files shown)


jmoenig:
genius! Thank you so much!!

PR #1042, opened 2015-12-03
Code Style Checking

cycomachead:
This is a discussion about code style and what rules to allow/disallow. 

Since old.jslint.com is no longer, I think we should use a more modern solution which is more configurable. I'm working on setting up [ESLint](eslint.org) which has a config file, and a command line tool. 

So far I've started with a relatively standard set of rules, and you can see the changes I've made, but this is by no means ready...


morphic.js
```diff
@@ -1270,7 +1270,7 @@ function copy(target) {
         }
     } else {
         c = {};
-        for (property in target) {
+        for (property in target) { /* eslint guard-for-in: 0 */
```
cycomachead: How do we feel about comments like this for special cases in code? (Things like eval, etc). They should be rare, but if we want to be thorough, they will exist.

cycomachead:
Here's an example of error output in the current config:
```Michael at Yosemite-Retina in ~/Dropbox/Projects/snap/snap on style
$ eslint *.js

/Users/Michael/Dropbox/Projects/snap/snap/blocks.js
   1730:17  error  Expected no return value  consistent-return
   3060:9   error  Expected no return value  consistent-return
   5441:5   error  Expected no return value  consistent-return
  11718:9   error  Expected no return value  consistent-return
  11736:9   error  Expected no return value  consistent-return

/Users/Michael/Dropbox/Projects/snap/snap/byob.js
  2054:9  error  Expected no return value  consistent-return

/Users/Michael/Dropbox/Projects/snap/snap/cloud.js
  641:5  warning  Unexpected alert  no-alert

/Users/Michael/Dropbox/Projects/snap/snap/gui.js
  4666:9  error  Expected a return value  consistent-return
  5024:5  error  Expected a default case  default-case

/Users/Michael/Dropbox/Projects/snap/snap/lang-dk.js
  860:9  error  Don't use octal: '\00'. Use '\u....' instead  no-octal-escape

/Users/Michael/Dropbox/Projects/snap/snap/morphic.js
  1341:9  error  Expected a default case                         default-case
  1365:5  error  Expected a default case                         default-case
  1665:9  error  Expected literal to be on the right side of <=  yoda
  6963:9  error  Expected no return value                        consistent-return
  6965:9  error  Expected no return value                        consistent-return
  6967:9  error  Expected no return value                        consistent-return

/Users/Michael/Dropbox/Projects/snap/snap/objects.js
  1553:13  error  Expected no return value  consistent-return
  3719:5   error  Expected no return value  consistent-return
  3828:9   error  Expected a return value   consistent-return
  5160:9   error  Expected a return value   consistent-return
  5164:9   error  Expected a return value   consistent-return
  5168:9   error  Expected a return value   consistent-return
  5172:9   error  Expected a return value   consistent-return
  5176:9   error  Expected a return value   consistent-return

/Users/Michael/Dropbox/Projects/snap/snap/paint.js
  782:9  error  Expected no return value  consistent-return

/Users/Michael/Dropbox/Projects/snap/snap/store.js
   345:9   error  Expected an object to be thrown                               no-throw-literal
   527:17  error  ["var"] is better written in dot notation                     dot-notation
   531:17  error  ["var"] is better written in dot notation                     dot-notation
   589:9   error  Expected an object to be thrown                               no-throw-literal
   621:9   error  Expected an object to be thrown                               no-throw-literal
   702:9   error  Expected an object to be thrown                               no-throw-literal
   994:17  error  ["var"] is better written in dot notation                     dot-notation
  1133:5   error  Expected a default case                                       default-case
  1395:1   error  Array prototype is read only, properties should not be added  no-extend-native

/Users/Michael/Dropbox/Projects/snap/snap/threads.js
  2304:13  warning  Unexpected alert         no-alert
  2579:13  error    Expected a default case  default-case

✖ 36 problems (34 errors, 2 warnings)
```

morphic.js
```diff
@@ -2690,9 +2689,7 @@ Morph.prototype.drawOn = function (aCanvas, aRect) {
 
 Morph.prototype.fullDrawOn = function (aCanvas, aRect) {
     var rectangle;
-    if (!this.isVisible) {
-        return null;
-    }
+    if (!this.isVisible) {return; }
```
cycomachead: This is an example of fixing the inconsistent return error. 

See these docs: http://eslint.org/docs/rules/consistent-return.html

cycomachead:
A nowhitespace diff: https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1042/files?w=0

Ugh, autofixing doesn't do well with the translation files. I'll be sure to fix those.

morphic.js
liam4: That should probably be either as removed or all on one line without brackets, i.e.

```
if (!this.isVisible) {
    return null;
}

if (!this.isVisible) return null;
```

Also the code you added doesn't work the same way as the code you removed, because `null !== undefined`.

cycomachead: The Snap<em>!</em> style has brackets around all signle line cases. (IMO, the spacing semantics Jens' chose are a bit unconventional, but hey, it's his code and it's readable. :grinning:)

As far as `null !== undefined`...yes, these are different, but this was an edge case that I actually found using the linter. `fullDrawOn()` shouldn't actually return anything, and later in the function, it (implicitly) returns undefined. Also, the draw mechanisms aren't really supposed to return a value, so using the default `undefined` makes the most sense...

cycomachead:
Closing because dealing with merge conflicts is too much. Also, I assume Jens will probably object to this. ;)

PR #1046, opened 2015-12-05
Fix isAutoFill to allow false values

cycomachead:
currently the expression isAutoFill || true will always evaluate
to true! :O

(That makes the argument somewhat useless though.) So, this fixes things...


cycomachead:
Note that this only comes up when working to embed Snap<em>!</em> in part of a page, but that's what I'm doing!

jmoenig:
thanks, good catch!

Issue #383, opened 2014-04-08
Snap! on Retina Displays


cycomachead:
This definitely isn't a priority, but there's a problem on retina displays, at least the 15" Macbook Pro (Safari, sometimes Chrome but not FF) where there's 'shadows' of blocks being shown after they're being dragged.

See this video, since that's the easiest. The 'ghosting' disappears when you click the background. 
https://drive.google.com/file/d/0Bz4PCnNknwnkSWpMeWFqVXlrbHc/edit?usp=sharing

Also, Snap is kinda blurry (especially text) on Retina screens. Is there anyway we could tweak morphic so that it uses the full pixel area of the screen? Currently it seems to be just  using the half pixel size that's then scaled up.

bl-nero:
The simple trick with scaling just one canvas element won't help, because every element creates its own canvas, and it has to be scaled accordingly. If you set the scale on the outermost canvas, then you'll end up drawing zoomed versions of low-resolution images on the same canvas, with equally poor results.

There are two solutions:

1. As a temporary stop-gap measure, I'd suggest using the image-rendering CSS property, partially supported by the latest versions of Chrome and FF. It can just display the pixelated, low-resolution UI. I'll prepare a pull request this week.

2. A full solution would be more challenging, since it would require resizing all of the canvas elements. I'm making some experiments now.

cycomachead:
@bl-nero Yeah, I tried that a bit, but didn't get super far. Feel free to poke around #628

bl-nero:
I managed to get a bit further; the only thing left behind is SymbolMorph. Here's my approach: create every canvas in the device resolution, and immediately apply the scaling factor. However, it's tricky: whenever you refer to canvas' width or height properties, you get device-pixel values, so you need to carefully scale these things. Also, whenever you change canvas dimensions, you need to apply scaling again, because the transformation matrix is reset.

I'm considering using a simple canvas polyfill that would just hijack read and write access to the width and height properties; that would solve our problems for now, and prevent them from reappearing in the future. An alternative solution is to write a couple of utility functions for converting dimensions, but then you risk forgetting about using them.

Here's what I've got now:

<img width="1440" alt="screen shot 2015-10-12 at 09 57 00" src="https://cloud.githubusercontent.com/assets/9391503/10423537/a47bca64-70c7-11e5-93d4-64677a6f410f.png">

bb010g:
[...]

PR #1063, opened 2015-12-13
Make Snap look better on retina screens

bl-nero:
This fixes issue #383. Here's how it works: Whenever a width or height of a Canvas element is read/written, it is scaled appropriately so that on Retina displays we can use high-resolution canvas elements, and still treat them as if their dimensions were the same as they would be on a non-retina display. This way we don't need to take this correction in every single routine that draws anything.

The retina-enabled canvas is not used as sprites, however, so that the Snap projects maintain compatibility between retina and non-retina systems.

**Note that this change is quite intrusive**, because it's performed in an "aspect programming" style, by modifying behavior of the Canvas element, so I strongly encourage everyone to test it thoroughly. I tested it on Chrome and Firefox using a Mac system.


cycomachead:
How does one decide whether they should call `newCanvas` or `newNonRetinaCanvas`?

bl-nero:
* You MAY use `newNonRetinaCanvas` if the image in question displays an image of a sprite. Using a retina canvas for this purpose wouldn't harm, but it would be a waste of memory, since it would display a stretched image anyway.

* You MUST use `newNonRetinaCanvas` if at any stage of the image's lifecycle it represents the *source of truth* about an image of the sprite. This is because the images saved into the project are saved without DPI information. So—aside from issues with increased project size—after reopening the project, an image in question would appear to be twice as big. And it would grow every time the project is saved and read again.

* You SHOULD use `newRetinaCanvas` in any other case to prevent the rest of the UI from looking blurry.

Do you think that I should put this as a comment to the source code?

bl-nero:
Note: @show0k has noticed a weird issue that occurs when you use the browser's zoom feature, so even if you like it, please don't merge it just yet. I'm trying to fix this.

cycomachead:
Yeah, I'd add it to the source code. :+1: 

And this is tagged "in progress" so it won't get merged until ready. :)

But it's looking good! I'm excited to use this.

bl-nero:
FYI, it looks like the "zoom bug" is something transient - when the page is refreshed and it *starts* zoomed, everything is OK. From the way the UI behaves, I think that the dimensions of widgets are properly recalculated, but some of the canvases don't get recreated during this operation.

cycomachead:
Interesting, there's probably some call to `drawNew()` or `changed()` maybe that could be added?

bl-nero:
I've already tried that. So here's the deal:

Changing zoom is kinda like changing dimensions. The UI thinks that it's smaller, and it triggers the fillPage function. However, what also happens is that the browser kindly informs the application that the device pixel ratio has changed (obviously). And guess what—I've gone out of my way to cheat the system, ignore that, and draw every canvas 1:1 on the screen. And that's precisely what is being done. :) It works in the current release, because the pixels are stretched, which means that when UI is zoomed, it physically shrinks, but most of the canvases are intact; they're just copied to the screen with a different ratio.

I've blindly tried with drawNew, but it didn't work; turned out that this operation is not recursive. But even if I recurse through the entire UI, it creates chaos; it looks like some widgets have never been designed to  support calling drawNew() at any point of their lifetime.

I'm kinda stuck here. The only thing that comes into my mind is to read the device pixel ratio once at the beginning, and then using this as a constant. However, it would pose a problem with dual-display systems, where one of the displays is a traditional monitor.

But hey, now when I came to think about this, the dual-display scenario should now suffer from the same bug... I'm wondering whether this case is somehow different, or it's just that it's been a long ago since I tried to move Snap window between displays, and the latest version of this code just broke this. I'll need to check this.

bl-nero:
OK, I've tried recursively calling drawNew one more time, to investigate what's happening. It looks like almost everything is OK, with one exception: some of the morphs suddenly jump to the top left corner of the screen. They still pass the mouse events to their parents, so they are still in the hierarchy; it's just their positions that are being suddenly reset to (0,0).

bl-nero:
I finally got some spare time to get back to this. Unfortunately, I haven't found any decent solution to the problem above, so I decided to stick to a workaround that makes the zoom in/out case at least as good as previously in Snap (making it zoom in/out with the original DPI setting; the quality of the views should be much better after refreshing the page).

bl-nero:
I guess this is pretty much going to be the final version, and it's already quite old (sorry, my fault obviously) so approvals would be appreciated. :)

bl-nero:
Ping?

bl-nero:
Hey, folks, can anyone take a look? @cycomachead @jmoenig It's getting old.

liam4:
If I ever get a chance I'll try to test this on a retina MacBook Pro..

cycomachead:
Sorry I've been really busy lately, but overall I'm very excited about this! :+1: :+1: :+1: 

The only concern I have is performance and memory. On a desktop computer, I think this should be fine, but on smaller devices like an iPad or iPhone. I'd have to test this to be sure. Basically, I think it might be nice to be able to disable the retina scaling even on a retina screen. (Perhaps not a setting, just in code.)
I think that would only affect the `newRetinaCanvas` function right?

ubertao:
This is exciting! I tried it a bit on Retina MBP, the UI is much better! Haven't found any problem with Chrome/Firefox so far.

issue1: doesn't work with Safari.
issue2: when dragging from external display (non-retina) to retina display, gui stays blurred.

liam4:
I'm curious how well this effects the iPad Pro.. as you can see, from [this screenshot](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/issues/1192#issuecomment-215520556), it looks *very* blurry (online Snap*!* as it is now; compared to the tab bar).

screenshot

Anybody able to take a screenshot of Snap*!* on this branch on an iPad Pro?

cycomachead:
> I'm curious how well this effects the iPad Pro.. as you can see, from this screenshot, it looks very blurry (online Snap! as it is now; compared to the tab bar).

This _should_ work the same on all retina iPads. However, I have serious concerns about memory usage and iPads...A blurry but usable Snap<em>!</em> is preferable to a crisp but unusable version.

jmoenig:
I agree about the memory concern. Maybe we can just host a version in some dev directory and try it!

morphic.js
```diff
@@ -1052,6 +1052,185 @@
...
+            this._isRetinaEnabled = enabled;
+            if (getPixelRatio(this) != prevPixelRatio) {
+                this.width = prevWidth;
+                this.height = prevHeight;
+            }
+        }
+    });
+
+    var uberWidth = Object.getOwnPropertyDescriptor(canvasProto, 'width');
+    Object.defineProperty(canvasProto, 'width', {
```
ubertao: This is causing error on Safari/Mobile Safari:
`TypeError: Attempting to change the setter of an unconfigurable property.`

ubertao:
As a result of above error, Mobile Safari on iPad shows a blank page, just as Safari on Mac does.

brianharvey:
What's going to happen about this?  I may be forced to buy a new Mac soon (stupid Turbo Tax is dropping 10.7 support next year) and I want it to be perfect. :)

as-com:
@brianharvey Aren't upgrades free? (unless your computer is actually that old)

brianharvey:
@as-com Yes, but MacOS gets worse and worse with each "up"grade.

jmoenig:
So, this is great work! I'm reluctant to go ahead and integrate it mostly because it's not supported in every browser, and also because it uses 4x more memory just for graphics, and we're already pretty low on that. I guess what we could and should do is put up a fork of HD-Snap on our server and play with it for a while on different platforms to find out more about possible limitations.

The pressure to do something about retina displays has been greatly reduced (imho) by recent updates to MacOS, that sugnificantly improve how non-retina graphics are displayed on retina screens. It used to be really bad, but has gotten so much better to a degree where am no longer profoundly unhappy about using Snap on my retina screen.

brianharvey:
How recent, exactly?  We just got a complaint about this on the Scratch forum, and the (adult, sympathetic) user calls it "unusable."

jmoenig:
Yeah, I've seen that, and I'm not sure. All I can say is that Snap certainly felt "unusable" to me on my retina Mac at one point, but that it seems to have gotten much better. So, I can't say when exactly this happened but I think it's been better for quite a while now. Hmm... or do you presume I've just gotten used to crappy graphics?

brianharvey:
I dunno -- I've never used a retina display.

cycomachead:
> The pressure to do something about retina displays has been greatly reduced (imho) by recent updates to MacOS, that sugnificantly improve how non-retina graphics are displayed on retina screens.

Chrome is still pretty bad to me...I'm not sure it's gotten better in the past couple years.

> or do you presume I've just gotten used to crappy graphics?

Probably. The blurriness bothers most people I know, including myself.


> I'm reluctant to go ahead and integrate it mostly because it's not supported in every browser, and also because it uses 4x more memory just for graphics,

We can handle the browser support by checking if those functions are available. (And we should do that.)

jmoenig:
Hmm.. do maybe the claim that retina scaling has gotten better is just wishful thinking on my side :-/

Okay, fact is, @bl-nero 's changes are totally solving the issue in a way that makes Snap look breathtakingly beautiful on retina displays. For this reason alone we really should work on making it generally available to everbody, by which I mean integrating it. 

How about I release v4.0.7 now and then we all work on this one next?

cycomachead:
> How about I release v4.0.7 now and then we all work on this one next?

Yes. I _really_ love this work, but it's not a blocker on anything, per-se.

The other "solution" is to allow disabling retina support via a setting. (I will allow someone to pinch me every time I suggest this.) Given that high-DPI displays seem to be more affordable/available than fast processors, this might be worth considering. I'm thinking on iPads, and the 12" Macbook that's as underpowered as it is thin, not to mention the slew of Android and Windows tablets.

Klortho:
I'm the "adult sympathetic" user. Here's another data point: I haven't done the latest upgrade that my macbook keeps bugging me about lately, to "El Capitan", because I lost my Apple ID.

jmoenig:
I agree, Michael. I would love to have such a setting myself, and I don't think there's anything wrong with optionally disabling retina support at all, especially on high-res low-CPU hardware such as phones. I think that's exactly the way to go. The other nice think I love about @bl-nero 's code is its compactness and self-contained-ness. It looks easy enough to maintain it in the long run with all the relevant changes really being in mostly a single chunk of code in Morphic.js.

morphic.js
```diff
@@ -1052,6 +1052,185 @@
...
+                ctx.mozBackingStorePixelRatio ||
+                ctx.msBackingStorePixelRatio ||
+                ctx.oBackingStorePixelRatio ||
+                ctx.backingStorePixelRatio || 1;
+
+    // Unfortunately, it's really hard to make this work well when changing zoom level, so let's
+    // leave it like this right now, and stick to whatever the ratio was in the beginning.
+    var originalDevicePixelRatio = window.devicePixelRatio;
+
+    function getPixelRatio(imageSource) {
```
cycomachead: Thought: If we are only setting the pixelRatio once per-canvas, could we remove this function and make it a property of the Canvas object? It seems like the only change happens when retina enabled is set, so maybe we could just add the line in that function?

jmoenig:
Hi, @Klortho , I might actually be wrong about things having gotten better on MacOS recently, Michael certainly disagrees about that, and Michael is usually right on these things. So, looks like @bl-nero has has already solved this issue in a perfect way which is awesome. I'll push a new release today that will *not* incorporate these changes, because I want to give them some more testing (the retina-supported version feels a little more sluggish to me, but that might not be a big deal) and also include @cycomachead 's suggestion to let users optionally turn retina support off. Thanks!

jmoenig:
Wait! @cycomachead , are you suggesting we could / should pull this *now*? Do you think we could just add the switch to the settings menu (and perhaps an availability test for Safari/mobile Safari) and add retina support to the upcoming v4.0.7 release?

morphic.js
ubertao: pixel ratio will change if you drag the browser window from retina display to an external non-retina display.

jmoenig: true, @ubertao , we have the same problem in GP, thanks for pointing it out...

jmoenig:
Okay, so some of my benchmarks show a slowdown of the retina version compared to the non-retina one of up to a factor of 3.6 for cloning intensive projects. That's to be expected, I guess, when dynamically creating lots of hi-res bitmaps.

We need to do more tests on this. I'm now going to push the new release features for first-class sprites, and after that we should all work on these important pull-request, so they get integrated next thing.

morphic.js
cycomachead: I'll have to re-read, but I thought that the current solution wasn't able to automagically detect the DPI changing if you move the window.

In any case, I'm not sure that dragging a window between different DPI screens would be that common. I say that, knowing that I do this decently often but in my experience dual monitor setups aren't really all that common.

cycomachead:
> Do you think we could just add the switch to the settings menu (and perhaps an availability test for Safari/mobile Safari) and add retina support to the upcoming v4.0.7 release?

I definitely think we should wait... I think we'll need to do some testing to see how we can improve performance. 

We should also catch the JS errors.

cycomachead:
> Okay, so some of my benchmarks show a slowdown of the retina version compared to the non-retina one of up to a factor of 3.6 for cloning intensive projects. That's to be expected, I guess, when dynamically creating lots of hi-res bitmaps.

Thought: We could provide a crisp UI,  but a not so crisp stage? Obviously, this isn't preferable, but I think the blurry text is the worst of the annoyances (to me).

jmoenig:
With the release out of the way, let's get this included!

jmoenig:
Okay, turns out with some optimizations that I've wanted to try anyway, the lags I've benchmarked yesterday totally disappear. This is exciting!

as-com:
First, optimize all the things. There is no excuse for slow code. 😜

If that isn't enough, we could try detecting lag (like excessively low frame rate), and switching to a non-retina canvas.

bl-nero:
Folks, sorry for being silent on this thread for a while; I've been a bit
busy/offline. I'll take a look at it by EOW, and will try to implement the
menu switch, as this seems to be an lowest-hanging fruit (I don't really
have enough inside knowledge of the project to be able to perform a
benchmark).

On Tue, May 3, 2016 at 1:04 PM Andrew Sun <notifications@github.com> wrote:

> First, optimize all the things. There is no excuse for slow code. 😜
>
> If that isn't enough, we could try detecting lag (like excessively low
> frame rate), and switching to a non-retina canvas.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1063#issuecomment-216496656>
>
-- 
--
Bartosz Leper

jmoenig:
Awesome! Don't worry about the benchmark, I'm confident it's going to be okay after having conducted some tests myself. Thanks!

jmoenig:
I'm going to do a little maintenance release later today, and then I'll be on vacation until the end of May. It would be so great to include this PR for the next release after my vacation, first week of June. @bl-nero and @cycomachead , what do you think, is that realistic? I really love this contribution!!

cycomachead:
Yeah, that sounds good. Perhaps we can put this on snapsource/dev/ in the mean time? That would give others an easy way to test. 

I now have a 4K desktop monitor to test things with, and I'd like to do that but nothing is going to happen for a little while. (Thesis writing and graduation!)

jmoenig:
Okay, I'm playing with this right now, will upload a first test version to dev soon. Thank you so much for this awesome code, @bl-nero , it's really enlightening and very nicely done!

jmoenig:
Since I totally suck at Git but couldn't wait to get this integrated into the current dev version I've copied it over manually. Sorry, @bl-nero that this pull-history doesn't reflect that it's all your code. I've tried to make up for that by including credits to your contribution all over the code, and also in the "about" dialog's credits section. I sincerely apologize for my incompetence and hope that you're okay with this solution. Now I'll close this pull request. Again, THANK YOU, THANK YOU, for this awesome and sorely needed contribution! It's such an honor and such a pleasure to attract contributors who are so much more knowledgable in these tricky polyfill issues.

jmoenig:
I've posted the live dev version including your changes to:
  http://snap.berkeley.edu/snapsource/dev/snap.html

liam4:
How to git:

*undo most recent commit* (simple `git reset HEAD^`)
*re-open pull request*
*click accept pull request*

### SUCCESS!

liam4:
No difference on a retina MacBook Pro (early 2015, OS X 10.10.4)..
screen shot 2016-05-10 at 4 27 08 pm
screen shot 2016-05-10 at 4 27 11 pm

as-com:
@liam4 Confirmed, Safari 9.1 on OS X 10.11.4 does not display the retina version

jmoenig:
Snap resorts to non-retina on browsers that don't let you overload canvas::width etc. as is the case in Safari.

Folks, please *do not* tell me about bugs and glitches in retina version and don't spam the issues or whatever forums with them, I'm well aware of them and will work on them after my vacation (in June). I just wanted to get the dev version on Github so other involved persons can work on it, I'm *not* soliciting feedback at this stage of development, there will come a time for that. Thanks!

cycomachead:
For reference, the next version of Safari will work fine with retina screens!

https://developer.apple.com/safari/technology-preview/

I'm using it and everything is good! 👍 That means we'll probably be limited to retina support on iOS 10 devices, but I think that's probably acceptable...

Klortho:
I commented on this issue a while ago, on the forum. My son and I have just started really getting into snap a lot lately. I just noticed today, when I cloned the repo and ran it locally, that it looks *much* better on my retina display, and I'm wondering why this hasn't been deployed to the main site? 

I hate to bug, but it seems like such an important change. And, I can't help noticing >400 open issues and 70 pull requests. I'm brand new here so I don't want to be too presumptuous, but, @jmoenig , is there any change of you adding a few more committers and delegating some of the work?

Klortho:
BTW, I've been poking around in the code, and I want to say, it is *gorgeous*. This is the first time, ever, that I having had to fight the urge to reformat everything. And, as another complement, let me let you know that my son, who is 4, is really learning to program, and it's very exciting to watch.

PR #1069, opened 2015-12-16
Add native C&P support to morphic text inputs!

cycomachead:
All the hard work goes to the folks working on Edgy for this one!
@cyderize wrote Edgy's implementation, and I copied that almost exactly.

https://github.com/snapapps/edgy/blob/08f6c473aed6526202eaa84f82d4819e18ccbf05/edgy/clipboard.js

* In the current edgy implementation, there's a `hiddenDiv` which
contains the `hiddenText` textarea. I found that div to not be
necessary, so it was removed.
* I tweaked a couple `==` to be `===`.

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -4496,19 +4496,18 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
         this.target.setAlignmentToLeft();
     }
     this.gotoSlot(this.slot);
-    
+
     // Add hidden text box for copying and pasting
     this.hiddenText = document.createElement('textarea');
-    this.target.hiddenText = this.hiddenText;
     this.hiddenText.style.position = 'absolute';
     this.hiddenText.style.right = '101%'; // placed just out of view
-    
+
     document.body.appendChild(this.hiddenText);
-    
+
     this.hiddenText.value = this.target.selection();
     this.hiddenText.focus();
     this.hiddenText.select();
-    
+
     this.hiddenText.addEventListener(
         'keypress',
         function (event) {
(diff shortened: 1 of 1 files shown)


jmoenig:
Yay!!

morphic.js
```diff
@@ -4496,6 +4496,56 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
         this.target.setAlignmentToLeft();
     }
     this.gotoSlot(this.slot);
+    
+    // Add hidden text box for copying and pasting
+    this.hiddenText = document.createElement('textarea');
+    this.target.hiddenText = this.hiddenText;
```
jmoenig: why do we need to cache the hidden text in the target? Won't that reference stick around even and be orphaned if the hidden text gets deleted?

cycomachead: I'm not sure, since to be honest. I'll try removing it and see what happens. :-)

Though, may @cyderize knows, since he really deserves the credit for this!

cyderize: I actually don't see a reason for `hiddenText` to be in the target... I might've just forgotten to remove it after doing tests.

jmoenig: Ah, I already thought so but wasn't sure whether or not there might be some other functionality depending on this. Thanks, @cyderize, for pioneering this wonderful and badly needed feature!

cycomachead:
Removed that line. Thanks!

scspaeth:
Thanks for this improvement! Does anyone know what the limits are on size of the paste buffer?

cycomachead:
That's probably OS dependent. OS X doesn't seem to care but definitely can choke sometimes on BIG clipboards. Morphic also seems to have issues with really long text elements but these are not likely to show up in practice. 

-- 
Michael Ball
From my iPhone
michaelball.co

Issue #1071, opened 2015-12-16
Passing an empty list to length of causes infinite loop


cycomachead:
From a recent issue see this screenshot:
![screen shot 2015-12-16 at 1 27 24 pm](https://cloud.githubusercontent.com/assets/1505907/11854455/dbd57f5a-a3f8-11e5-8990-46b2f8abc51e.png)

jmoenig:
yeah, this isn't new in any way, and we'll introduce type checks for all list primitives anyay (already done in my dev version). I don't think any of this is urgent right now. :-)

jmoenig:
Oh, and it doesn't cause an "infinite loop"...

cycomachead:
> Oh, and it doesn't cause an "infinite loop"...

Well, it's more of a hang I guess, which is actually worse, IMO.

(Which is weird in a way because passing in `NaN` does actually cause an infinite loop. And `undefined - 1` is `NaN`...)

Issue #1067, opened 2015-12-15
Bugs in the UI clicking interface


cseager:
Today a bunch of new bugs appeared in the interface. When creating a custom block, I can't click on dropdown lists such as "set variable to...". When I try to click on it to select a variable, the mouse just picks up the block instead... I'm being forced to use all global variables because I can't select script variables due to this bug. I also can't seem to pick up some of my custom blocks out of the palette area and into the scripting area... no matter where I click, the block just executes and causes a type error. Also I can't save, because of a type error somewhere in the project. Clicking on the stop button in the tool bar to stop all scripts from executing does not solve the problem. Here's a screenshot of the error I get when trying to save. Deleting the custom block allows me to save. 

<img width="446" alt="screen shot 2015-12-15 at 5 24 23 pm" src="https://cloud.githubusercontent.com/assets/2761922/11826109/3f09d4fc-a351-11e5-85d1-b9c008fcd7e6.png">

cseager:
update: I downloaded the repo and switched the HEAD to 42fd60427f8224311a613426b9589c5d7791c940. That version seems to be working as expected without problems.

jmoenig:
Please make sure to flush your cache until the module versions show the following:

![404151215](https://cloud.githubusercontent.com/assets/2596285/11826663/1cfdb9b0-a387-11e5-8887-bc9931e8a59c.png)

Let me know if the problems still persist after that.
Thanks!

cseager:
@jmoenig ever since I loaded SNAP from my local rolled-back repo yesterday, I have not been able to reproduce some of the original bugs. The module versions are as you show when I load it from the web now. I'm pretty confused about what's going on here... I had no problems working over the weekend, and had cleared my cache out a couple of weeks ago when I got a login error. So I think it's pretty likely that I was working with up-to-date modules until yesterday, and I'm not sure in what way they could go wrong to cause this. 

I was able to reproduce some other crashing behavior I got yesterday. When you click on the following test block without inputting a list, it basically crashes the browser tab to the point where the only way to stop is to close the tab. I can't click on the block, I can't click the stop button, I can't even reload the page at that point. Raising a TypeError was the previous behavior. 
![image](https://cloud.githubusercontent.com/assets/2761922/11848109/bd553d22-a3ee-11e5-8e13-aa252431bcdf.png)

Here's a link to a test program. Just click the test block to execute it to see what I mean. 
[...]

PR #1072, opened 2015-12-16
Improve error handling in `length of`

cycomachead:
Immediate fix for #1071, tagging #1067

This makes the `length` check more robust,
and also adds a more helpful error message if you can't check
for length.

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -1553,7 +1553,12 @@ Process.prototype.reportListLength = function (list) {
     if (list instanceof List) {
         return list.length();
     }
-    return list.length; // catch a common student error
+    if (list && list.hasOwnProperty('length')) {
+        return list.length; // catch a common student error
+    }
+    throw new Error(
+        'expecting a list as input, but received ' + this.reportTypeOf(list)
+    );
 };
 
 Process.prototype.reportListContainsItem = function (list, element) {


jmoenig:
Thanks, I've already fixed this in my dev version, along with type checks for the other list primitives.

jmoenig:
(at this time I want the code base to "cool down", and the user base to settle...)

cycomachead:
In the case of this length of bug, and the fact that bad inputs cause the browser to hang, I really don't think we should wait! Repeat length of is all too common.

(though you're more than welcome to use your fix, I just think sooner is better in this case. The other repeat fixes can wait. They'be been there for years!)

jmoenig:
The "length of" bug has also been there for years. At this time I'm not going to introduce any more fixes for bugs that haven't been introduced in v4.0.4 yesterday. We can plan for a maintenance release later this month of next month, but unless some showstopper comes up that we've introduced *yesterday* I'll just push it on the stack of next changes.

cycomachead:
> The "length of" bug has also been there for years.

But is hasn't. Because before, bad inputs threw an error preventing any hanging.

jmoenig:
Oh wait, you're right!
Well, In this case I have to think about what I want. I have a version with type checks for every list primitive that I want to commit with the next batch of changes. Okay, maybe some time early next week then.

Actually, looking how complicated a simple primitive like this has become I think I might revert and *not* allow text inputs again...

cycomachead:
> Well, In this case I have to think about what I want. I have a version with type checks for every list primitive that I want to commit with the next batch of changes. Okay, maybe some time early next week then.

Thank you, thank you, thank you!!

I was starting to work on this on my own. 

> Actually, looking how complicated a simple primitive like this has become I think I might revert and not allow text inputs again...

To be honest, if there is a clear error message, I think that's fine.

jmoenig:
I don't share Brian's enthusiasm about error messages at all, and I also don't share many teacher's infatuation about "exploring" what happens when you compare a list with a Boolean. I really, really want to keep Snap "snappy" and make it snazzier yet so we can do the data stuff we've been planning for. Getting bogged down and bloating the evaluator with costly type checks and literate error messages isn't going to help. If something /breaks/ as the "length of" block currently does (only because I let myself being talked into *allowing* to be sloppy about a particular type by a teacher, geez!!) that needs to be fixed. But I need to think about this whole process some more. Maybe I'll introduce a "laggy but literate" mode (not kidding) and a "quick and quiet" one, and you have to go to a different web page for the "literate errors messages" version of Snap :-)

jmoenig:
... an I'll rename the "type-nitpicking-literate-error-messages" version of Snap into "Lag"... :-/

cycomachead:
I don't know that typechecking everything is necessary, but we either have errors or we don't. In either case, there needs to be some check because showing native JS errors to students is ineffective and confusing.

brianharvey:
I wouldn't oppose a checkbox in Settings controlling the extent of error checking.  But I would argue that once you do the error check, giving a *good* error message doesn't slow down the performance of error-free programs compared to giving an incomprehensible error message.  It only slows down programs with errors, which is fine.

Actually the checkbox is growing on me...  You could run in fast mode until something goes wrong in your project and you can't figure it out, and then you run in strict mode and get a helpful error message.

cycomachead:
> I wouldn't oppose a checkbox in Settings controlling the extent of error checking. But I would argue that once you do the error check, giving a good error message doesn't slow down the performance of error-free programs compared to giving an incomprehensible error message. It only slows down programs with errors, which is fine.

Performance-wise, the code to handle errors whether it's a try/catch or a set of if/elses will slow

brianharvey:
Yes, I understand checking for errors slows things down (which is why there might be a mode).  My point is that once you've found an error, it doesn't matter if giving a good message is slower than giving a bad message, because it only happens if there *is* an error.

cycomachead:
Yeah, that's true.

Of course, there's the question of whether 3-4 conditional checks really slows things down that much?

brianharvey:
I think Jens thinks that once we start down that route people will demand strict type checking on all the primitives.

cycomachead:
> I think Jens thinks that once we start down that route people will demand strict type checking on all the primitives.

Yeah, but that's why I suggest some spec or doc. It we write these things down, then it's much harder to complain. 

> My point is that once you've found an error, it doesn't matter if giving a good message is slower than giving a bad message, because it only happens if there is an error.


So, I actually had a second idea, that's performant when there's no errors. But it requires a little more work. But we should do it anyway. ;)

We already have `handleError` which takes in an error and the block. A block _already_ encodes it's type information in the block spec! All we need to do is pass the function inputs to handle error, (unless we can already get them) and parse the block spec. If there are multiple parameters, the first type error bubbles up. 

This requires one other change, which I think will useful:
We need to amend some method declaring input slot types so that they can have their type information inspected via code. Doing this could allow other cases of introspection. I actually had a concrete example of something I wanted to do a while ago, but now I've forgotten it. :/

Also, we should have a way of generating human-friendly names from block specs, or just manually writing them. That what when we say "expected a list, but received a number" we can say "length of ...". (Or maybe we have some custom error strings attached to each block that are prepended to any type errors. Then we can write _really_ friendly messages.)

I could make an example PR with this idea.

jmoenig:
slots, and also embedded reporters know about the type that's declared for them, the evaluator needs that type spec to decide whether it should evaluate an input or not (if it's type is "unevaluated"). The biggest recent speed-up has been to cache that type information instead of looking it up everytime the evaluator needs it. But regardless of the caching this already is totally introspectable.

But after all the recent panic I honestly need a break. I want Snap to settle down into stability (again) and not discuss error handling for a couple of months. I also hope to not get flooded with zillions of new "nice to have" features requests and half-panicked teachers complaining that Snap doesn't totally pre-empt all stupidity.

Happy holidays, everyone! :-)

brianharvey:
Yeah, I need some time for things other than answering Piazza posts, too.  :-)

jmoenig:
:-)

PR #1075, opened 2015-12-17
Make logoURL be a property of IDE

cycomachead:
This makes it easier to embed Snap! in another site where the JS code
and the html file are in different directories.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -221,6 +221,7 @@ IDE_Morph.prototype.init = function (isAutoFill) {
     this.projectName = '';
     this.projectNotes = '';
 
+    this.logoURL = 'snap_logo_sm.png';
     this.logo = null;
     this.controlBar = null;
     this.categories = null;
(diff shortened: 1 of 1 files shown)


jmoenig:
Hmm... how is changing line 224 any more convenient than changing line 490?

cycomachead:
Because `logo.texture` is inside `createLogo()`, so whomever is embedding Snap! would need to re-write that entire function just to change one line. (And the logo not working is somewhat of a pain whenever I start embedding Snap<em>!</em> into other things because by default it has to be in the same directory as the html file.

With this change, someone just needs to:

```js
ide.logoURL = '/some/other/path.png';
ide.createLogo();
```

Or, what I'm really hoping to do, is to start working on a guide for _responsibly_ embedding Snap<em>!</em> in other sites. As the usage of things like edX, Edgy and BeetleBlocks keeps growing, we need to make the embedding a little easier where we can.

So, my preferred pattern will actually look like this:

* `my-snap.html`:
```js
var IDE = new IDE_Morph();

IDE.initSnapCustomizations(); // or even: SnapOverrides.initCustomizations(IDE) 
IDE.openIn(world);
```
Then there can be an additional JS file that loads the customizations stuff.

jmoenig:
makes sense, thanks for the explanation!

PR #1097, opened 2016-01-06
Allow exporting script pics with results that error

cycomachead:
The allows you to easily get a script pic of a block showing an error.


jmoenig:
Thanks, Michael!

jmoenig:
I'll pull this for the next version that introduces new features. Right now I'm only after fixing bugs, but I'm not forgetting about this contribution!

PR #1101, opened 2016-01-06
Fix #1099, bug was introduced in #983

### LINKED ISSUE #1099: export in costume icon does not work
Throws an error in console

cycomachead:


jmoenig:
Thanks, Michael!

Issue #1047, opened 2015-12-06
How to support non-English typing in textbox in Snap?


gengshenghong:
Is it possible to enable multiple-byte input in SNAP? Example: Input Chinese for the "Project Notes". It seems all Chinese input will be ignored by snap.

After searching on internet, it seems this is an issue for HTML5 canvas-based input box. It is discusses in below thread (which already fixed the issue, though it seems like still not working from my test...)

https://github.com/kangax/fabric.js/issues/2051

Any idea whether it is possible to fix in snap? If yes, how? Appreciate if you have any idea to share.

Thanks,
Shenghong

gengshenghong:
Another thread discussed about the same issue:
http://stackoverflow.com/questions/21372051/fabricjs-itext-is-support-unicode-hangeul-not-moified/24205247#24205247

The demo provided there works for me in my test.

Thanks,
Shenghong

gengshenghong:
anyone has an idea? This issue should affect all non-English input method.

Basically, I saw 2 fix from what I learnt from Internet.
1. implement the canvas input box supporting multiple-byte input method (like Chinese).
Example: https://github.com/goldfire/CanvasInput
2. Have a HTML <input> on the canvas input box
The issue for this solution is that we need to know the position of the canvas input morph, I have no idea how to get the absolute position in Morph.js.

Thanks,
Shenghong

jmoenig:
IME support has now been introduced with https://github.com/jmoenig/Snap/releases/tag/v5.0.6

PR #1103, opened 2016-01-07
Enable IME Composing for some diacritics on Mac OS #4 X & ideograms #1047 & Android virtual keyboard

DarDoro:
Some keyboard layout on Mac OS X activates, so called, IME Composing for entering diacritics with "dead  key". During "composing", usual keyboard events (keydown, keypress) are useless. Keystrokes are modified by browser or even discarded. Composition results are supplied to JS by appropriate events (compositionstart, compositionupdate,compositionend). Composition events are fired , only for text input (according DOM level 3 events specification).
Changes, "morphic.js":
1)  added "compositionend" event handler to hidden input field "VirtualKeyboard" (used so far for touch devices input) and send simulated kypress to Morphic canvas
2) enabled "VirtualKeyboard" by default, not only for touch devices


jmoenig:
Wonderful, thank you so much for your awesome contributions!

jmoenig:
Can we perhaps fold C&P into this (I think we can)? That would really be wonderful!

cycomachead:
> Can we perhaps fold C&P into this (I think we can)? That would really be wonderful!

I'm working on it...but it will be kind of weird since `CursorMorph` has the selection data.

DarDoro:
I tried to merge clipboardHandler and virtualKeyboard but found that clipboadHandler has some drawbacks. "Mouse only" operation (selecting text, copying with menu) are not handled properly. Reselecting input field after every keystroke may also break some, more complex, IME scenarios. So I tried to use native Clipboard API (virtualKeyboard must be enabled). I made changes to my repository. I'm not sure if it's sufficient or should I make another pull request? 
Desscription of changes https://github.com/DarDoro/Snap/commit/0e124d49cbb3fe7e2fa5031fe7cb9d773f11e3da :

Native, generic Copy & Paste support for current browsers. "oncopy", "onbeforecopy" events. Removed "clipboardHandler".
Added "document.body" events :
- "beforecopy" to allow copy operation UI/shortcut
- "copy" supply morph data to clipboard

Minor changes:
- "paste" event : browser independent handling
- "keydown" remove "preventDefault" to allow keyboard shortcut for clipboard operation 

Tested with 
- Windows 10 : Chrome 47, Opera 34, Firefox 43, IE11, Edge
- Mac OS X, Safari 8

cycomachead:
This was automatically updated. :)

gengshenghong:
I found my question #1047 is referenced in your discussion and come to this thread. It is great to see this issue is implemented here. I've downloaded your code and tested, one issue is that in none-English typing mode, you cannot enter numbers or punctuation, is it possible to fix/improve this??

I'll also take a look at details of your code and learn it. Thank you for your hard and great working!

Thanks,
Shenghong

DarDoro:
I've checked this issue. Browser (Win10 IE11/Chrome/FF) did not emit any useful key nor composition events in this case. There is only "input" event.  Simulating key entry with "input" is not so obvious, as it requires some kind of delta processing and multiple events coordination to avoid duplication and breaking any existing functionality. I'm afraid that possible are only makeshift solution.

BTW: Win10/Edge browser emits composition events with entered numbers.

gengshenghong:
Yes, I've also debugged it just now and you are correct only "input" even is emitted. To support numbers/punctuation, we'll have to handle multiple events (input + compositionend + compositionstart) to avoid duplication. And I agree with you that it is a makeshift solution, but it seems there is no better way if using HTML input as virtualkeyboard to simulate the key entry.

I've tested this as below draft code and it works well from my testing, you may check whether you can add it to your code? (it is really a draft, I used 2 global varialbe to record the status...maybe need to put it in "WorldMorph" propery...):

// theser are 2 globals...put it outside anywhere...
var g_compositionend = null;
var g_iscomposing = false;

// in WorldMorph.prototype.initVirtualKeyboard
this.virtualKeyboard.addEventListener(
        "input",
        function (event){
          // console.log(event.isComposing);
          // https://w3c.github.io/editing/input-events.html
          if (g_iscomposing) {  // Note: input event property isComposing is not supported by browsers...
            return; // Ignore all "input" during composing
          } else {
            if (g_compositionend != null) {
              g_compositionend = null;
              return; // ignore the "input" after "compositionend".
            }            
            
            if (myself.currentKey == 229) { // double confirm it is in IME mode.
              var v = event.target.value; // Any better way to get the "last char" directly?
               var newEvent = {ctrlKey : false, altKey : false, shiftKey : false, metaKey : false,
                  keyCode : 0, charCode : 0, preventDefault : function(){}};
              newEvent.keyCode=0;  
              newEvent.charCode=v.charCodeAt(v.length -1);  // last char
              myself.keyboardReceiver.processKeyPress( newEvent);
            }
          }
        }
    );
    
    this.virtualKeyboard.addEventListener("compositionstart", function (event){ g_iscomposing = true; }, false);
        
    this.virtualKeyboard.addEventListener(
        "compositionend",
        function (event){       
        g_compositionend = event;
        g_iscomposing = false;
        
        var newEvent = {ctrlKey : false, altKey : false, shiftKey : false, metaKey : false,
                  keyCode : 0, charCode : 0, preventDefault : function(){}};
                  if (myself.keyboardReceiver) {
                    for (var i = 0; i < g_compositionend.data.length; ++i){
                      newEvent.keyCode=0;  
                      newEvent.charCode=g_compositionend.data.charCodeAt( i);
                      myself.keyboardReceiver.processKeyPress( newEvent);
                    }
                }
       },
        false
    );

DarDoro:
@gengshenghong I made suggested changes. I found only one drawback. Cursor navigation isn't properly blocked in "virualKeyboard.keydown" so after left or right arrow wrong character is entered. "preventDefault()" should be called only for proper C&P keyboard shortcut for every platform.

gengshenghong:
@DarDoro  Thank you. In my Chrome on Ubuntu 14.10 (and Chinese IME - soguo), I do not see issues with cursor navigation. Maybe IME-specific or browser specific?

DarDoro:
@gengshenghong My testing scenario:
Snap empty edit field. Entering with IME: "1", "2", cursor left, cursor left, "3", "4"
Expected results: 3412
Actual result: 2212

Not always last character from virualKeyboard is the just entered key.

gengshenghong:
@DarDoro  I've tested according to your scenario now, and it works for me with expected results in my environment. :)

SmallBookworm:
For fix the bug about tab,I think can add ```
this.virtualKeyboard.onkeydown=function(event){
        if (event.keyIdentifier === 'U+0009' ||
            event.keyIdentifier === 'Tab') {
            myself.keyboardReceiver.processKeyPress(event);
            event.preventDefault();
        }
    };
```
But after adding this snap will skip some selectable morphic.
And I can not enter shift-punctuations and number when I use IME as @gengshenghong said.So this can not be fixed?

**Comment by SmallBookworm** [comment]
And I pulled a request for the first character bug.

**Comment by ubertao** [comment]
During IME composition in a normal html/native text field, key sequence entered so far is shown (as place holder text), allowing user to inspect what he/she has typed. 

In this solution, the key sequence entered is not shown, making it hard for user to correct typos during composition.  Screenshots below illustrate the issue:
composition

This is crucial for languages like Chinese, of which all characters has to be entered by IME.

Possible fix: handles `input` event, update text field with place holder text.

The other problem is that in Safari, sometimes I can't enter some characters (punctuations, numbers, etc.), this happens a lot, but I haven't found the pattern.

**Comment by DarDoro** [comment]
I'm afraid, this PR was discarded (not accepted nor rejected, without any complaints, for months) ;(

I designed this PR primarily as easy and loosely coupled solution to old problem #4 . IME events gives INCREMENTAL information about newly "composed" input. This information is send to Snap as stream of simulated keystrokes. Characters are only added to edit field. Any other editing (deletion, selection, scrolling) is left up to the Snap. 
@ubertao  To get raw input feedback, characters should be constantly added and removed from input field. I thought it is quite difficult and error-prone task. Of course, if this PR will be accepted, it can be done in several ways.
1. As separate UI element, solely for this purpose -  ex: "say" or "think" bubble with "compositionupdate" data (easiest way)
2. Use composition* handlers to perform "event dance" (state machine) and constantly update entry field
3. Most promissing: don't hide html text input. Just style them, with CSS, and move in the right place to perform editing.

On the other hand: "input" event is more "generic". Raw input feedback and copy/paste are handled effortlessy, but you have only entire field value. To get usefull results every editing action and state should be mirrored, back an forth, beetwen Morph and html edit. As you can see in PR #1165, it is quite difficult ,tedious task and requires a lot of changes to original code base.

**Comment by jmoenig** [comment]
I neither discarded nor rejected this PR or the other one that addresses the same issue. Quite on the contrary, I'm very, very excited to open up Snap to languages with multi-byte charactersets! At one point soon we'll want to integrate one of these ways, and I honestly don't know which one. I promise to be more communicative about it after my vacation, i.e. in June :-)

**Comment by ubertao** [comment]
@jmoenig Yay! So let's keep this discussion on.

@DarDoro :

> IME events gives INCREMENTAL information about newly "composed" input. This information is send to Snap as stream of simulated keystrokes. Characters are only added to edit field. 

Are you sending all simulated keystrokes come out from one composition? Chinese (and maybe other languages) IMEs allow composing of phrases or even sentences, which means by the end of one composition, multiple characters get inserted at once. And this is the most common way of typing Chinese text, because typing characters one by one will be much more time consuming. Per my test, sometimes when I compose a phrase, only the last one character get inserted, others are lost.

> To get raw input feedback, characters should be constantly added and removed from input field. I thought it is quite difficult and error-prone task. Of course, if this PR will be accepted, it can be done in several ways.
> 1. As separate UI element, solely for this purpose - ex: "say" or "think" bubble with "compositionupdate" data (easiest way)

Not sure it will work. According to the [doc](https://developer.mozilla.org/en-US/docs/Web/Events/compositionupdate#Browser_compatibility):
```
[2] compositionupdate events are fired before editor contents are actually changed; 
that is, the value of an editable element has not yet been changed when 
compositionupdate event handlers are called. 
```
> 2. Use composition* handlers to perform "event dance" (state machine) and constantly update entry field

Same problem as above.

> 3. Most promissing: don't hide html text input. Just style them, with CSS, and move in the right place to perform direct editing.

Actually most parts of IDE can be done this way, but I'm not sure this is in align with the design guideline of Snap.

> On the other hand: "input" event is more "generic". Raw input feedback and copy/paste are handled effortlessy, but you have only entire field value. To get usefull results every editing action and state should be mirrored, back an forth, beetwen Morph and html edit. As you can see in PR #1165, it is quite difficult ,tedious task and requires a lot of changes to original code base.

Agreed, but I think it's still doable, as I already have a near-complete solution. I might be able to submit a PR in a few days, it can serve as a backup plan in case all issues cannot be solved by this one or PR #1165.

DarDoro:
- 
Added IME input feedback during editing. Newly entered characters are emphasized by selection.
There is a still problem with candidate window placement. I want to correct this problem with separate commit to ensure readability. But there is also Snap problem with cursor position caused by variable-pitch font, additionaly exagerated for multi-byte character sets.
There are no spaces in input fields below. Only stream of "1". Cursor at end of text.
 
snap

- 
Remove special case handling for punctuation marks and numbers. 

- 
Re-focus to prevent focus steal by candidate window.

@SmallBookworm You submitted the same bug-fix to both repository (master and my fork). Only master is necessary. Also there is a small mistake in WorldMorph.prototype.previousTab.

ubertao:
`StringMorph`: works well! A minor issue - when I cancel a composition (by pressing ESC), the uncompleted char sequences stay in the input field. The expected behavior (as in a html input field) is that the uncompleted char sequences are removed.
`TextMorph`: (e.g. Project Notes or comments) cursor movement (by mouse or keyboard) is broken.

DarDoro:
IME changes are removed when composition is canceled. As You wish ;) @ubertao .

SNAP Fix : Cursor position for String & Text Morph is calculated for the entire text not character by character. Modified _slotPosition()_, _slotAt()_.

ubertao:
Great, another step forward!

**One issue**: Can't clear composition with `backspace/delete`. (At least on a Mac).

To reproduce, select an IME (e.g. Japanese or Chinese), press `a`, press `delete/backspace`.
Result: composition remains in input field.
Expected: composition is cleared.

**Can't input some characters when IME enabled**

Reported above, now I've found the reason: IME will surpress `keypress` event for some keys. 
E.g. with _Japanese Romaji IME_ enabled, alphabet keys won't fire `keypress`, with _Chinese Pinyin IME_ enabled, number keys and punctuation keys won't fire `keypress`.

Possible fix: move char insertion to `keydown` handler.

jmoenig:
Okay, so not it appears we have two proposed solutions to this issue, this one right here, and @swiperthefox's #2322 . Is there any consensus among @DarDoro @swiperthefox @ubertao @Simon-Mong @bromagosa @cycomachead on which is the best to actually pull? Tnx

DarDoro:
For me it's ok to use #2322

jmoenig:
Thank you, @DarDoro !

cycomachead:
I’ve read through a bunch of stuff but I still feel like I don’t have a super great understanding. 

So far we don’t have a PR that uses compositionend events but it sounded like that would potentially be the best solution? #2332 seems pretty good to me, though I think we need pretty rigorous testing of any of these PRs. 

Michael 

-- 
Michael Ball
From my iPhone
michaelball.co

DarDoro:
First I'm not "native" user of IME, so most findings was made by @gengshenghong, @smallbookworm, @ubertao.
"composition*" events designed to support IME have some flaws in browsers. Somtimes browsers "forgot" to emit any useful event for numbers and punctuation. No "composition*" nor "key*" event but  textarea is updated correctly.

"Input" event is far more "generic" and with proper handling is enough for common scenarios. Even #1103 uses "input" for patching "composition*" flaws. 

I'm think, that for touch devices, all-native input controls should be optionally allowed. Android soft keyboard has no cursor keys. C&P is initiated by long press in edit control. Cursor navigation can be supported by enlarged visual anchor or swipe over "space" gesture. Lack of such editing support gives bad user experience.

cycomachead:
> Thanks! That's helpful.


I think this approach is pretty reasonable then.

PR #1105, opened 2016-01-07
Fixed #1090

### LINKED ISSUE #1090: Morph.fullImageClassic doesn't work for morphs with children to the left/above them
Morph's `fullImageClassic()` function has a comment asking why it doesn't work for all blocks (currently [here](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/morphic.js#L2736)). I believe I know the answer.

The function translates the canvas using `this.bounds`, then calls `this.fullDrawOn()`, which uses `this.fullBounds()` to offset its drawing. This works fine, unless the full bounds have a different origin than the bounds, which occurs when a morph's children are above or to the left of it. I believe a few simple changes will fix the problem.

This should fix it:

```diff
 Morph.prototype.fullImageClassic = function () {
-    // why doesn't this work for all Morphs?
-    var fb = this.fullBounds(),
+    var fb = this.cachedFullBounds || this.fullBounds(), // use the cache since fullDrawOn() will
         img = newCanvas(fb.extent()),
         ctx = img.getContext('2d');
-   ctx.translate(-this.bounds.origin.x, -this.bounds.origin.y);
+   ctx.translate(-fb.origin.x, -fb.origin.y);
    this.fullDrawOn(img, fb);
    img.globalAlpha = this.alpha;
    return img;
}
```

Still needs to be tested more thoroughly, but since you only really use this function for screen captures and dragging, should be limited use cases.

P.S. Why do you have both functions in the first place? Efficiency? Seems like it makes it easier for bugs to crop up in one or the other like this.

thomaswp:
Fixes #1090, a bug in Morph.fullImageClassic() where it uses the bounds instead of the fullBounds.

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -2733,11 +2733,10 @@ Morph.prototype.toggleVisibility = function () {
 // Morph full image:
 
 Morph.prototype.fullImageClassic = function () {
-    // why doesn't this work for all Morphs?
-    var fb = this.fullBounds(),
+    var fb = this.cachedFullBounds || this.fullBounds(), // use the cache since fullDrawOn() will
         img = newCanvas(fb.extent()),
         ctx = img.getContext('2d');
-    ctx.translate(-this.bounds.origin.x, -this.bounds.origin.y);
+    ctx.translate(-fb.origin.x, -fb.origin.y);
     this.fullDrawOn(img, fb);
     img.globalAlpha = this.alpha;
     return img;


jmoenig:
Thanks! I'll pull this for the next version that introduces new features (at this time I'm only pulling fixes for bugs that actually affect Snap use in schools, so please don't think I'm forgetting this :-)

PR #1109, opened 2016-01-12
Refactor labelPart

cycomachead:
This doesn't change any functionality -- it's just a refactoring. It's not urgent, but I wanted to see what other's thoughts on this would be. (@bromagosa and anyone else doing extenstions.) 

I'm trying to extract out the block Spec labeling because of some work I'm doing for autograding. I'm not 100% sold on this design but I do think it makes the code more readable. It will be nice to have a list of all possible spec types in code, and this makes adding new spec types in an extension easier.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -1508,7 +1508,7 @@ SyntaxElementMorph.prototype.labelPart = function (spec) {
             rcvr.costumes.asArray(),
             function (each) {return each.name === id; }
         );
-        var part = new SymbolMorph(cst);
+        part = new SymbolMorph(cst);
         part.size = this.fontSize * 1.5;
         part.color = new Color(255, 255, 255);
         part.isProtectedLabel = true; // doesn't participate in zebraing
--- a/blocks.js
+++ b/blocks.js
@@ -754,6 +754,724 @@ SyntaxElementMorph.prototype.fixBlockColor = function (
 
 // SyntaxElementMorph label parts:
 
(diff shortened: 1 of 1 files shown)


jmoenig:
Hmmm. I'm not sold on this.

I see your intention of creating a "dictionary of truth" about the possible specs rather than having to look at the ```switch``` statement in the source code. What I do like about the ```switch``` statement, however, is, that 1) it's easy to read (no closures necessary) and 2) it's definitive. You can be sure that *all* there is has to be within that single ```switch``` statement, whereas your dictionary can be changed by other code. Then the truth no longer is in the code but in the data. I presume that's the point behind your suggested change, but I disagree.

As a general design principle I want to describe a system in code. I want people to be able to understand a system by reading the code. This seems like a no-brainer, but in fact, many systems are not described that way, instead the code describes how to set up a bootstrap system that *generates* the target system. Such indirection has benefits as everyone knows that "there is no problem in CS that cannot be solved by introducing another lever of indirection". keeping the slot specs in a dictionary makes them extensible, there can be other modules changing the appearance or function of existing specs or adding other ones by means of JavaScript. I'm not enthusiastic about any of this, because that way JavaScript becomes the target system instead of Snap. For 4.1 extensibility I was thinking myself about a ```CustomInputSlot``` object that could be edited within Snap itself, same as CustomBlocks can be. That way regular libraries could have their own special slots, but they would be Snap objects rather than a JavaScript code change.

PR #1117, opened 2016-01-26
Updated bounds after pos change and added btn width check. Fixes #1116

### LINKED ISSUE #1116: DialogBoxMorph width doesn't stretch to fit buttons
In `fixLayout`, the DialogBoxMorph does not check that the width of the buttons fits on the dialog box.

btn-overflow

Also, when the (row) AlignmentMorph is calling `fixLayout`, it doesn't fix the position of the buttons while calculating the bounds. This results in the bounds being equal to the widest button rather than the actual bounds of the positioned buttons.

brollb:
This fixes #1116

### DIFF
--- a/widgets.js
+++ b/widgets.js
@@ -2513,6 +2513,12 @@ DialogBoxMorph.prototype.fixLayout = function () {
                     + this.buttons.height()
                     + this.padding
         );
+        this.silentSetWidth(Math.max(
+                this.width(),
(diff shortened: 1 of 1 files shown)


jmoenig:
Nice, thanks!

jmoenig:
Oh, looks like I forgot about this one. Will pull next week!

PR #1125, opened 2016-02-02
Fix #1124, bad call to finding an IDE_Morph

### LINKED ISSUE #1124: Help context menu item causes error insude of custom blocks
Inside the custom block editor, if a user right-clicks on a block and selects the "Help" item, nothing will happen. The browser throws an exception saying the IDE is null.

untitled

Here's the stack trace:
```
Uncaught TypeError: Cannot read property 'resourceURL' of null
BlockMorph.showHelp @ blocks.js:2640
TriggerMorph.trigger @ morphic.js:8350
MenuItemMorph.mouseClickLeft @ morphic.js:8583
HandMorph.processMouseUp	 @ morphic.js:9753
(anonymous function) @ morphic.js:10365

```

cycomachead:
Fix #1124 
This one was my fault....

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -2599,7 +2599,7 @@ BlockMorph.prototype.restoreInputs = function (oldInputs) {
 
 BlockMorph.prototype.showHelp = function () {
     var myself = this,
-        ide = this.parentThatIsA(IDE_Morph),
+        ide = this.parentThatIsA(BlockEditorMorph).siblings()[0],
         pic = new Image(),
         help,
         comment,


blocks.js
```diff
@@ -2599,7 +2599,7 @@ BlockMorph.prototype.restoreInputs = function (oldInputs) {
 
 BlockMorph.prototype.showHelp = function () {
     var myself = this,
-        ide = this.parentThatIsA(IDE_Morph),
+        ide = this.parentThatIsA(BlockEditorMorph).siblings()[0],
```
jmoenig: Oh wait, that can't work, right? What if the block is in the regular scripting area, then it won't find the IDE anymore. Also, I'm a little nervous about assuming the IDE to always be the first sibling of a BlockEditor, although, I guess, it usually should be.

How about something like this (I didn't test this, just making this up now as I write this):

```JavaScript
var ide = this.parentThatIsA(IDE_Morph),
    blockEditor;
if (!ide) {
    blockEditor = this.parentThatIsA(BlockEditorMorph);
    if (blockEditor) {
        ide = blockEditor.target.parentThatIsA(IDE_Morph);
    }
}
````

cycomachead: Ugh, I don't why I didn't try that...but you're right! 

Your solution definitely seems like it should work.

jmoenig:
Just tagged this to be included in the upcoming release. It's "blessed" but I need to look it over, so it's not going to be pulled, but the issue will be solved :-)

jmoenig:
I've fixed this in the dev version that I'm about to push...

jmoenig:
fixed in big commit today

PR #1140, opened 2016-02-18
Refactored symbolCanvasColored. Fixes #1139

### LINKED ISSUE #1139: Suggested symbolCanvasColored Refactor
Currently, it is difficult to add a new symbol to the SymbolMorph (in `symbolCanvasColored`). The large switch case makes extending it a little tricky (ideally, you would subclass SymbolMorph and override this method - with a `super` call of course). Of course, you can also override the entire function with a slightly modified version but that is ugly and will not allow you to receive any new symbols, etc from the core Snap! project. This approach also requires that `SymbolMorph.prototype.names` is modified to also contain the new symbol.

It would be nice if this functionality was stored in a dictionary like SymbolMorph.SYMBOLS which would allow the easy addition of new symbols by simply adding the entry and then writing the associated function (no method override/subclassing or updating `SymbolMorph.prototype.names`)

brollb:
This is a suggested refactor for issue #1139

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -1460,7 +1460,7 @@ SyntaxElementMorph.prototype.labelPart = function (spec) {
         // allow GUI symbols as label icons
         // usage: $symbolName[-size-r-g-b], size and color values are optional
         tokens = spec.slice(1).split('-');
-        if (!contains(SymbolMorph.prototype.names, tokens[0])) {
+        if (!SymbolMorph.SYMBOL[tokens[0]]) {
             part = new StringMorph(spec);
             part.fontName = this.labelFontName;
             part.fontStyle = this.labelFontStyle;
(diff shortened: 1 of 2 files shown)


cycomachead:
You might find #1109 interesting. :) I personally :+1: this change, but I'm not Jens!

jmoenig:
Yes, this looks reasonable to me.
As an alternative I've been thinking about ever since introducing SymbolMorph we could get rid of the the whole symbol-drawing method binding and simply follow the convention that the drawing method's name always follows the conventions:

```JavaScript
'drawSymbol' + sym[0].toUpperCase() + sym.slice(1)
```

This would be even less code while being way more flexible...
What do you think?

cycomachead:
I'd personally like to see
```js
'drawSymbol' + sym[0].toUpperCase() + sym.slice(1)
```
as
```js
'drawSymbol' + capitalize(sym)
```
Where `capitalize` is defined anywhere convenient.

brollb:
I like the idea of enforcing the naming convention. The only drawback to this approach is it makes the enumerating the supported symbols difficult (as done in [byob.js](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1140/files#diff-afb7a00b1866ca6d6477b267bc8c644cL2456))

jmoenig:
Enumeration of available symbols is a good point, @brollb

PR #1142, opened 2016-02-18
Border color functions

bthomson2:
Added the functions and blocks to CSnap to set the border colors. It seemed that adding it via CSnap vs the module loader was relevant. My reasoning being that having availability of some border settings, but not all of them, in different modules may cause confusion. General availability allows for more creativity in any drawing that may be done in CSnap.


jmoenig:
Wow, I'm very happy to see CSnap being so actively developed! I need to get an update of all the cool things you're doing with it. It looks like this pull-request should probably be addressed to the main CSnap branch instead of Snap, right?

SillyInventor:
Sorry, I don't come here often enough to notice this. Yes, Brenda is new and getting used to git. You should close this.

If you're interested in the dev going on you can always ping me and I will fill you in. I hope to do a downstream merge of the awesome things you've been doing later this month.

PR #1084, opened 2015-12-23
"Save and Share..." option in the file menu

cycomachead:
This builds of @bromagosa's work. There is a now an item in the file menu
which when selected will save the current project to the cloud, then will
present the user with a text box to copy the URL.

This makes the following changes from what's currently in BettleBlocks:
- If the project hasn't been saved yet, then show a save dialog box first
- Presenting the URL is now done with a native morphic dialog box! Now that
  C&P works well, this looks much nicer. :)
- Adds a new IDE method `publicProjectURL = function (username, projectName)`
  - This gives the public project URL. I'm probably going to use this to
    refactor some existing code for getting URLs later.

There are some TODO elements:
- [x] Definitely needs a thorough code review
- [x] The dialog box looks a bit off. I think I could use some pointers here...
- [ ] Until we have context menus, we probably want some text to tell people
  _how_ to copy text.
- [x] Don't make a cloud request for the "share" service if the project is
  already public
- [x] Update the browser's URL on successful sharing.

Deployed at http://snap.berkeley.edu/michael

PR #1146, opened 2016-02-24
Support Loading More Resources from the `resourceURL` method

cycomachead:
These are a few places which load resources, that might also be in subfolders. I ran into these for autograding and for the desktop app.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -221,7 +221,7 @@ IDE_Morph.prototype.init = function (isAutoFill) {
     this.projectName = '';
     this.projectNotes = '';
 
-    this.logoURL = this.resourceURL('', 'snap_logo_sm.png');
+    this.logoURL = this.resourceURL('snap_logo_sm.png');
     this.logo = null;
     this.controlBar = null;
     this.categories = null;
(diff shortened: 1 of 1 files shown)


jmoenig:
Hmm... These resources are the tools file, translations and the logo. I think these are special and should actually be always present and loaded from their absolute locations, especially the translations and the logo, don't you agree?

cycomachead:
For s.b.e this doesn't change anything. 

For other apps, the issue is that the current method doesn't account for the case if you need to serve Snap! from a dynamic server. That's what's being done for autograding, and for the snap desktop app too. So in those case the override for `resourceURL` looks a bit different, but still relatively simple. (For autograding and the desktop all that's needed is to preprend some stuff to the URL, but that has to happen for the tools xml and all files, not just the content in subfolders.)

jmoenig:
My concern is that I want translations to also work when I simply download the sources and open snap locally on my computer without setting up a local web server. It looks like this should still work. Have you tried that? I really need that to test all the translations, and it's a big feature because there are very many people using Snap that way.

cycomachead:
Yeah, that all works fine. 

Essentially, in normal Snap<em>!</em> `resourceURL` is basicaly a no-op. (Well, if you're passing in folders, it will add the `/` for you, but otherwise, it's a pretty simple function.)

jmoenig:
Okay, then I'm pulling it now. Thanks!
Any other last minute PRs coming in, or any really pressing things that should go into the upcoming release?

cycomachead:
Not super pressing, I suppose, but #1084 is done from my end, I think, and would be helpful.

PR #1165, opened 2016-03-16
Support Non-English Characters in Morphic

SmallBookworm:
support non-English typing in textbox

Tag #1047

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -4512,7 +4512,8 @@ CursorMorph.prototype.initializeInputHandler = function () {
     this.inputHandler = document.createElement('textarea');
     this.inputHandler.style.position = 'absolute';
     this.inputHandler.style.right = '211%'; // placed just out of view
-
+    this.inputHandler.start=0;
+    this.inputHandler.end=0;
     document.body.appendChild(this.inputHandler);
     this.inputHandler.focus();
     this.inputHandler.value = this.target.text;
@@ -4534,47 +4535,80 @@ CursorMorph.prototype.initializeInputHandler = function () {
         function (event) {
             myself.target.deleteSelection();
             myself.target.text = myself.inputHandler.value;
-            myself.gotoSlot(myself.inputHandler.selectionStart);
-            myself.target.drawNew();
+
             myself.target.changed();
+            myself.target.drawNew();
+
+            myself.gotoSlot(myself.inputHandler.selectionStart);
+            myself.target.changed();
+            myself.target.drawNew();
         },
         false
     );
     this.inputHandler.gotoSlot = function (Slot) {
         this.setSelectionRange(Slot, Slot);
     };
-
-    this.inputHandler.insert = function (str, staPos,endPos) {
+this.inputHandler.saveSelection=function(){
+    this.start=this.selectionStart;
+    this.end=this.selectionEnd;
+};
+    this.inputHandler.restore=function(){
+        this.setSelectionRange(this.start,this.end);
+    };
+    this.inputHandler.insert = function (str, staPos,endPos,changed) {
         var selection=this.selectionStart;
         var tmpStr = this.value;
-        this.value = tmpStr.substring(0, staPos) + str + tmpStr.substring(endPos, tmpStr.length);
-        //if selectionStart != selectionEnd,this is wrong.
-        if (endPos <= this.selectionStart){
-            myself.gotoSlot(selection + str.length);
-        }else {
-            myself.gotoSlot(selection);
-        }
+        this.value = tmpStr.substring(0, staPos===null?selection:staPos) + str + tmpStr.substring(endPos===null?selection:endPos, tmpStr.length);
+        if(changed){
+            //if selectionStart != selectionEnd,this is wrong.
+            if (endPos <= this.selectionStart){
+                var length=str=='\n'?0:str.length;
+                this.gotoSlot(selection + length);
+            }else {
+                this.gotoSlot(selection);
+            }
+            myself.target.text = this.value;
 
+            myself.target.changed();
+            myself.target.drawNew();
+
+            myself.gotoSlot(this.selectionStart);
+            myself.target.changed();
+            myself.target.drawNew();
+        }else {
+            //if selectionStart != selectionEnd,this is wrong.
+            if (endPos <= this.selectionStart){
+                var length=str=='\n'?0:str.length;
+                this.gotoSlot(selection + length);
+            }else {
+                this.gotoSlot(selection);
+            }
+        }
     };
 };
 
 
 CursorMorph.prototype.processKeyDown = function (event) {
+
     var shift = event.shiftKey;
     if (event.ctrlKey && (!event.altKey)) {
-        this.ctrl(event.keyCode);
+        this.ctrl(event.keyCode,event);
         // notify target's parent of key event
         this.target.escalateEvent('reactToKeystroke', event);
         return;
     }
     if (event.metaKey) {
-        this.cmd(event.keyCode);
+        this.cmd(event.keyCode,event);
         // notify target's parent of key event
         this.target.escalateEvent('reactToKeystroke', event);
         return;
     }
-    //prevent the error cased by inputHandler
-    if (shift)event.preventDefault();
+    if (shift){
+            if(event.keyCode>64&&event.keyCode<91){
+                this.inputHandler.insert(String.fromCharCode(event.keyCode),null,null,true);
+            }
+        //prevent the error cased by inputHandler
+        event.preventDefault();}
     switch (event.keyCode) {
         case 37:
             this.goLeft(shift);
@@ -4609,12 +4643,11 @@ CursorMorph.prototype.processKeyDown = function (event) {
             }
             return this.target.tab(this.target);
         default:
-            nop();
+          nop();
     }
     // notify target's parent of key event
     this.target.escalateEvent('reactToKeystroke', event);
 };
-
 // CursorMorph navigation:
 
 /*
@@ -4752,7 +4785,7 @@ CursorMorph.prototype.undo = function () {
     this.inputHandler.gotoSlot(0);
 };
 
-CursorMorph.prototype.ctrl = function (aChar) {
+CursorMorph.prototype.ctrl = function (aChar,event) {
     if (!isNil(this.target.receiver)) {
         if (aChar === 68) {
             this.target.doIt();
@@ -4762,11 +4795,13 @@ CursorMorph.prototype.ctrl = function (aChar) {
             this.target.showIt();
         }
     }
-
-
+if(aChar==65){
+    this.target.selectAll();
+//    event.preventDefault();
+}
 };
 
-CursorMorph.prototype.cmd = function (aChar) {
+CursorMorph.prototype.cmd = function (aChar,event) {
     if (!isNil(this.target.receiver)) {
         if (aChar === 68) {
             this.target.doIt();
@@ -4775,6 +4810,10 @@ CursorMorph.prototype.cmd = function (aChar) {
         } else if (aChar === 80) {
             this.target.showIt();
         }
+    }
+    if(aChar==65){
+        this.target.selectAll();
+  //      event.preventDefault();
     }
 };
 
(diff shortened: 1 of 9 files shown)


jmoenig:
Thank you!
I'm just now working on packaging a minor release today. After that I'll have a closer look at this pull request and start testing it. So please don't think I'm rejecting this if it doesn't get pulled immediately, I'm very glad that you've worked out a way to enter non-Western characters, Snap really needs that! In the meantime can you tell us a little more about your architecture? What's the general idea behind it?

SmallBookworm:
User type in a hidden textare and then give the value to Morph's text.And i find i forgot to remove the code for test......

SmallBookworm:
Now I have only fond one bug.In StringMorph I added some code for mose selection,but it do not work.I do not know why

SmallBookworm:
I added some file of workspace by accident.

ubertao:
This is exactly what I thought! Similar solution as [CanvasInput](https://github.com/goldfire/CanvasInput) is using.

@SmallBookworm you might want to rename `CursorMorph.clipboardHandler` to something like `CursorMorph.inputHandler` as it's now handling all text editing event.

I'm teaching my 8 year old daughter Snap and this is so far the biggest problem for me. I hope this PR could be merged soon.

morphic.js
```diff
@@ -4504,101 +4504,47 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
 };
 
 CursorMorph.prototype.initializeClipboardHandler = function () {
-    // Add hidden text box for copying and pasting
+    // Add hidden text box for texting
     var myself = this;
 
     this.clipboardHandler = document.createElement('textarea');
     this.clipboardHandler.style.position = 'absolute';
     this.clipboardHandler.style.right = '101%'; // placed just out of view
```
ubertao: One suggestion: Instead of positioning the underlying `<textarea>` element outside the view port, you may want to put it at the same position of related `StringFieldMorph`, so that the IME's candidates panel can be shown near the input box. (Inspired by [CanvasInput](https://github.com/goldfire/CanvasInput/blob/master/CanvasInput.js#L127))

```javascript
var stringFieldBounds = this.target.parent.bounds;
this.clipboardHandler.style.left = stringFieldBounds.origin.x + 'px';
this.clipboardHandler.style.top  = stringFieldBounds.origin.y + 'px';
this.clipboardHandler.style.width = (stringFieldBounds.corner.x - stringFieldBounds.origin.x) + 'px';
this.clipboardHandler.style.height = (stringFieldBounds.corner.y - stringFieldBounds.origin.y) + 'px';
this.clipboardHandler.style.pointerEvents = 'none';
this.clipboardHandler.style.opacity = 0;
```

ubertao:
Some preliminary tests show good results, it works almost perfectly with Mac OS's bundled IME as well as some popular 3rd party IMEs (on Chrome/Safari/Firefox).

The only issue encountered: When cursor is at the beginning of the line, I cannot select the first character use `Shift + →`. But this seems to be a bug already in current Snap version.

SmallBookworm:
Yes.And I find I can not comment for unknown reasons.so we have found two bugs.I will try to fix them.

SmallBookworm:
At last if you right-click and add comment,
you could not write a  long word.Now you can write long word but the cursor is showed in wrong position.

DarDoro:
[Shift] + character doesn't work at all. Also [Ctrl] + A (select all). 

Chrome 50, Firefox 46, IE11 @ Windows 10

SmallBookworm:
Yes,and I added some file of workspace by accident....The two bug is easy to fix.The biggest bug is comment.It is wrong from the last version.

DarDoro:
The same problem for "say" block (and probably every autosized text entry field). I think, the edit field must accomodate new content (resize, scroll) before cursor position can be set/calculated. This works for me:
**_morphic.js_**
```
this.inputHandler.addEventListener(
        'input',
        function (event) {
            myself.target.deleteSelection();
            myself.target.text = myself.inputHandler.value;
            / ************
            myself.target.changed();
            myself.target.drawNew();
            / ************
            myself.gotoSlot(myself.inputHandler.selectionStart);
            myself.target.changed();
            myself.target.drawNew();
```

SmallBookworm:
I tried your way.It works.That is great!

jmoenig:
Okay, so how ready is this pull-request? I'm thinking we should maybe integrate it rather sooner than later... (?)

SmallBookworm:
But I did not fix the two small bug([Shift] + character doesn't work at all. Also [Ctrl] + A (select all)).

jmoenig:
Okay, so we should look into this next. This is exciting!

SmallBookworm:
I think this version can be used.You can test.

jmoenig:
Thanks, I will!

ubertao:
@SmallBookworm There are still some issues (tested with Chrome on Mac OS X):

* Character get inserted into wrong position. To reproduce: enter '123', press shift+left, then press left (now cursor is between 1 and 2: '1|23'), then enter 'x', you will get '12x3' while '1x23' is expected.
* Shift-punctuations e.g. `()""@$<>:` doesn't work.
* IME's candidates window is away from editing field, you have to constantly look at different places to choose the right characters. This is very annoying for languages like Chinese, which heavily relies on the candidates window during input.
* When editing a block's input slot: selection is not replaced but appended. (Not supporting `TextMorph.selectAllAndEdit()` and `StringMorph.selectAll()` well).
* Console errors when pasting text: `Uncaught TypeError: Cannot read property 'getData' of undefined`

ubertao:
(I wanted to comment on your code, but diff won't show up as the change set is too big.)

About the console error, it's caused by:

```javascript
    document.body.addEventListener(
        "paste",
        function (event) {
            var txt = event.inputData.getData("Text");
//                          ^^^^^^^^^ event.clipboardData.getData("Text");
            if (txt && myself.cursor) {
                myself.cursor.insert(txt);
            }
        },
        false
    );
```

I think you can totally remove these lines as 'paste' is always followed by a 'input' event, which you already processed for updating the text.

DarDoro:
There are still some issues
Entering a block's input slot with mouse or TAB selects entire field (as expected) but subsequent keystrokes are not handled properly.

It's partialy Snap fault. Edit field content should be edited then selected. Not the other way.  
**morphic.js**

```javascript
WorldMorph.prototype.nextTab = function (editField) {
. . .
       next.edit();
       next.selectAll();

WorldMorph.prototype.previousTab = function (editField) {
. . .
        prev.edit();
        prev.selectAll();
```

Entering field with mouse click tries to restore selection. But, there are no any  information stored. Selection data should be stored before moving cursor.

```javascript
StringMorph.prototype.mouseClickLeft = function (pos) {
        if (cursor) {
>>>>>       cursor.inputHandler.saveSelection();
            cursor.gotoPos(pos);
            if (this.endMark != this.startMark)
                cursor.inputHandler.restore();
```


Shift + letter does not work as expected. Special case for Shift modifier in __*CursorMorph.prototype.processKeyDown*__ 
```javascript
if (shift){
           if(event.keyCode>64&&event.keyCode<91){
               this.inputHandler.insert(String.fromCharCode(event.keyCode),null,null,true);
    //prevent the error cased by inputHandler
    event.preventDefault();}
```

does not handle Caps state at all. Always an uppercase letter is inserted.

Also entering Shift+letter into already selected field do not replace entire content as expected.
__*inputHandler.insert*__ clears selection (in inputHandler) and new character is appended.
Works for me after removing this code (also Shift-punctuations  ()""@$<>: ).

Chrome@Windows 10

DarDoro:
As for "paste" event - it can be cross-browser

```javascript
document.body.addEventListener(
          "paste",
          function (event) {
            var txt = event.clipboardData.getData("Text");		
            clpbrdData = event.clipboardData || window.clipboardData; // W3C || IE
            var txt = clpbrdData.getData("Text");
```
I used this in PR #1103 *'Enable IME Composing for some diacritics on Mac OS # 4 X & ideograms # 1047 & Android virtual keyboard'*, already submited few months ago, to solve exactly the same problem slightly different way.

SmallBookworm:
If you delete 
```
if (shift){
           if(event.keyCode>64&&event.keyCode<91){
               this.inputHandler.insert(String.fromCharCode(event.keyCode),null,null,true);
    //prevent the error cased by inputHandler
    event.preventDefault();}
```
When you select words by [Shift] + left(right) key,there is bug that you select more than you want.

SmallBookworm:
Thanks for  @DarDoro .The bugs are fixed,but in OS X IME's candidates window is not away from editing field,maybe the bug is about the IME.

morphic.js
```diff
@@ -4604,11 +4604,13 @@ CursorMorph.prototype.processKeyDown = function (event) {
         return;
     }
     if (shift){
-            if(event.keyCode>64&&event.keyCode<91){
-                this.inputHandler.insert(String.fromCharCode(event.keyCode),null,null,true);
-            }
-        //prevent the error cased by inputHandler
-        event.preventDefault();}
+        var sCode=[37,38,39,36,40,35,9];
+        for(var i in sCode){
```
ubertao: Or with simpler code: 
```javascript
if (sCode.indexOf(event.keyCode) !== -1) {
    event.preventDefault();
}
```

morphic.js
```diff
@@ -4604,11 +4604,13 @@ CursorMorph.prototype.processKeyDown = function (event) {
         return;
     }
     if (shift){
-            if(event.keyCode>64&&event.keyCode<91){
-                this.inputHandler.insert(String.fromCharCode(event.keyCode),null,null,true);
-            }
-        //prevent the error cased by inputHandler
-        event.preventDefault();}
+        var sCode=[37,38,39,36,40,35,9];
```
ubertao: Maybe you should also include `33(page up)` and `34 (page down)`.

SmallBookworm: ok

ubertao:
Unfortunately, none of my last reported issues get fixed, I can still reproduce them reliably.
Also I suggest you do more tests with `TextMorph`, it has more issues than `StringMorph`.
Actually along the road of investigating IME support, I've come up of my own nearly complete solution, but I still hope this PR can be merged soon so it won't be necessary for me to create another PR.

As for IME candidate panel issue, here's a screenshot:
snap-ime

The panel is always positioned at top-left corner of browser while the comment morph is far way from it.

It's not IME's bug, instead, it's a result of the absolute position of underlying `<textarea>`, I've proposed a fix in previous [comment](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1165#discussion_r60109028).

SmallBookworm:
really?Except the IME the other issues are fixed.

SmallBookworm:
And I think the PR[ #1103 ](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1103) by  @DarDoro is great ,but tab is not supported.

ubertao:
Indeed the bugs are fixed! (Sorry, I thought I pulled latest commit but I didn't!)

One more issue: Chinese sentence (or ascii text without space) longer than `TextMorph`'s width, keyboard editing (cursor movement, copy&paste) will mess up.

The problem is with current word wrapping logic, see inline comments below:
```javascript
    this.words.forEach(function (word) {
        if (word === '\n') {
            myself.lines.push(oldline);
            myself.lineSlots.push(slot);
            myself.maxLineWidth = Math.max(
                myself.maxLineWidth,
                context.measureText(oldline).width
            );
            oldline = '';
        } else {
//          if (myself.maxWidth > 0) {             <-- current code
            if (myself.maxWidth > 0 && oldline) { // fix: don't add empty line
                newline = oldline + word + ' ';
                w = context.measureText(newline).width;
                if (w > myself.maxWidth) {
                    myself.lines.push(oldline);
                    if(myself.cursor.inputHandler){
                        myself.cursor.inputHandler.insert('\n',slot-1,slot);
                    }
                    myself.lineSlots.push(slot);
                    myself.maxLineWidth = Math.max(
                        myself.maxLineWidth,
                        context.measureText(oldline).width
                    );
                    oldline = word + ' ';
                } else {
                    oldline = newline;
                }
            } else {
                oldline = oldline + word + ' ';
            }
            slot += word.length + 1;
        }
    });
```

SmallBookworm:
I think you should have a try in PR [#1103 ](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1103).It is really great.And I will fix the bug about first word,so I close this one. @jmoenig

ubertao:
I tried @DarDoro's solution and found some issue, will leave some comments there.

SmallBookworm:
So do you think his way is better than me?

ubertao:
I don't know which one is better without seeing more test result.

Your solution use `input` event, and @DarDoro uses `compositionstart/compositionended`, each of them may have some browser compatibility issue. 

Perhaps `input` is safer because Snap has been always using it and it has wider browser support:
- [input event browser compatibility](https://developer.mozilla.org/en-US/docs/Web/Events/input#Browser_compatibility)
- [compositionend event browser compatibility](https://developer.mozilla.org/en-US/docs/Web/Events/compositionend#Browser_compatibility)
- [compositionstart event browser compatibility](https://developer.mozilla.org/en-US/docs/Web/Events/compositionstart#Browser_compatibility)

PR #1179, opened 2016-04-06
Use a tab character in the libraries file

cycomachead:


jmoenig:
Thanks, Michael! I'll pull this asap, probably even before 4.0.7

PR #1185, opened 2016-04-19
Retrieved message from error. Fixes #1184

### LINKED ISSUE #1184: `getURL` failing silently
When I am previewing the examples (from the 'open project' dialog) on chromium/linux, the error message is not converted to a string automatically at https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/gui.js#L4810. This results in the error message failing in the TextMorph's [parse](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/morphic.js#L7755) function.

This could be fixed by simply calling `showMessage` with `err.message` rather than just `err`.

brollb:


jmoenig:
Thanks, Brian! I'll pull this when I prepare the upcoming release, probably next week.

gui.js
```diff
@@ -4807,7 +4807,7 @@ IDE_Morph.prototype.getURL = function (url) {
         }
         throw new Error('unable to retrieve ' + url);
     } catch (err) {
-        myself.showMessage(err);
+        myself.showMessage(err.message);
```
ubertao: Would you consider changing `err.message` to `err.toString()`, which include `err.name` in the result? 
Or, equivalently `''+err`.

brollb:
updated it to use `toString()`

PR #1187, opened 2016-04-21
updates Slovenian translation

bh213:
current one was pretty out of date

### DIFF
--- a/lang-si.js
+++ b/lang-si.js
@@ -183,11 +183,11 @@ SnapTranslator.dict.si = {
     'language_name':
         'Sloven\u0161\u010Dina', // the name as it should appear in the language menu
     'language_translator':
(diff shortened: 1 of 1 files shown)


jmoenig:
Thank you so much! This is great timing, as I'm planning to push the next release within the next days, and for this I'll also pull your update.

bh213:
Is there a way to check if I missed any translation or if there are obsolete keys that are no longer used? I used Croatian and German as template (since I am familiar with them) and they use different keys in many places.

ubertao:
@bh213 I wrote a small tool to simplify my own Snap translation work. I just created PR #1190 for it.

It's a single html file, so you can download and use it now if you want to give it a try.

However it can't tell you which keys are obsolete. I thought about automate that part but it seems it's non-trivial.

PR #1189, opened 2016-04-27
update Simplified Chinese translation

ubertao:
Simplified Chinese translation is quite outdated (3 years old), this pull request not only added newly strings since last translation, but also corrected some inaccuracies in old one.

### DIFF
--- a/lang-zh.js
+++ b/lang-zh.js
@@ -1,28 +1,27 @@
 /*
 
-	lang-zh.js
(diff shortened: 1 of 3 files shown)


jmoenig:
Thanks! this pull-request comes just in time for the upcomping release, which i'm holding off until next week because we have a big teacher workshop in NY this week. I'll probably pulll it early next week.

PR #1190, opened 2016-04-28
Localization tool, first version.

unotao:
A small tool that helps Snap! localization. 

**Feature**
- Compares 2 translation to identify added/removed entries;
- Generate updated translation file with newly added entries;
- Basic syntax checking (e.g. placeholder type/order in block spec).

**To test**: open it from a local web server.

**Screenshot**:
snap-l10n

### DIFF
--- a/l10n.html
+++ b/l10n.html
@@ -1,0 +1,661 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+		<title>Snap! L10N</title>
+		<link rel="shortcut icon" href="favicon.ico">
+		<style>
+			body {
+				font-family: sans-serif;
+				font-size: 14px;
+				background-color: #373737;
+				color: white;
+			}
+			input, textarea, select {
+				background-color: #B0B0B0;
+				border-width: 1px;
+				color: black;
+				font-family: sans-serif;
+				font-size: 14px;
+				border-radius: 0px;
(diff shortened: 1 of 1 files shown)


jmoenig:
Awesome. Thanks!

cycomachead:
This is great! I have some comments, but I'm not sure if you're finished?

ubertao:
Yes, all the features are finished.

l10n.html
```diff
@@ -0,0 +1,661 @@
...
+        -->
+        <script type="text/javascript">
+            var modules = {}; // for locale.js
+        </script>
+        <script type="text/javascript" src="locale.js"></script>
+        <script type="text/javascript">
+        (function () {
+            // Space separated puncutation list (same order as 'en'), used by lint().
+            // If one puncuation can have multiple translations, just group them together.
+            SnapTranslator.punctuations = {
```
cycomachead: Wouldn't it be better to maintain these in each translation file?

ubertao:
I didn't put them into translation file mainly because it's not part of translation and completely useless at runtime.

There's a chance the linting rules grow to be more complex in future, you might want to keep it separately so it doesn't pollute the translation file a lot.

l10n.html
ubertao: I didn't put them into translation file mainly because it's not part of translation and completely useless at runtime.

There's a chance the linting rules grow to be more complex in future, you might want to keep it separately so it doesn't pollute the translation file a lot.

PR #1193, opened 2016-04-28
Fix #1191; Fix #1174 Comment out `Canvas.toBlob()`

### LINKED ISSUE #1174: Problem to save locally the picture of a script
capture d ecran 2016-04-01 a 21 20 48

It is not possible to have the image of script saved on your computer with snap! 4.0.6.
It was possible with Snap! 4.0.4.

I need them to help my students. I have to  copy part of the screen to save the image, actually. Like this picture :
http://nathalierun.net/lycee/piwigo/picture.php?/5690/category/82

Thanks for your help.
Nathalie (a french-Reunion island teacher)

### LINKED ISSUE #1191: Chrome 50: Exporting script pics no longer works when offline
since the last update Chrome no-longer displays null-blob uri's with generated script pics when running Snap locally, i.e. not from a web-server. Calling in @cycomachead to look into this, if you get the chance to.

I'm extremely annoyed at recent browsers' efforts to discourage running JS stuff locally. While this seems to be just Chrome specific for now and only affect script pics I'm heavily inclined to switch back to simple data-urls for exporting script pics, if you can't find a reasonably simple fix.

Could this be the writing on the wall for local project saving to also run into problems?

Hmm... Did I mention I hate browsers?

cycomachead:
This was removed due to instability on the part of browser vendors.
`Canvas.toBlob()` _should_ be the preferred way of exporting a canvas
as an image. Blob URIs are shorter and don't cause Chrome crashes like
the `data:` URI scheme does. They are also faster to encode images, and
don't clog up browser history.

**However**, there are currently issues in browser support which make
it hard to save images generated by blob URIs. Chrome and Firefox both
have different mechanisms of exporting blob images to disk. (In Chrome
you must right click the image, whereas in Firefox you must use the Save Page As command.)

The current solution is this:
* When a script pic is generated, create a large `data:` URI/
* Check if the browser is Chrome and if the length is a problem
* If so, convert the URI to a blob manually.
* Use the `a[download]` attribute to force the browser to save to disk.

Note that the conversion to a manual blob is _much_ slower than a
native method, but as this is an exporting operation it shouldn't cause
too much of a headache for users since it should be a fairly low
occurrence that users hit that code path.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -3811,15 +3811,19 @@ IDE_Morph.prototype.saveFileAs = function (
 
 IDE_Morph.prototype.saveCanvasAs = function (canvas, fileName, newWindow) {
     // Export a Canvas object as a PNG image
-    // cavas.toBlob() is currently only supported in Firefox and IE
-    var myself = this;
-    if (canvas.toBlob) {
-        canvas.toBlob(function (blob) {
-            myself.saveFileAs(blob, 'image/png', fileName, newWindow);
-        });
-    } else {
-        this.saveFileAs(canvas.toDataURL(), 'image/png', fileName, newWindow);
-    }
+    // Note: This commented out due to poor browser support.
+    // cavas.toBlob() is currently supported in Firefox, IE, Chrome but 
+    // browsers prevent easily saving the generated files.
+    // Do not re-enable without revisiting issue #1191
+    // if (canvas.toBlob) {
+    //     var myself = this;
+    //     canvas.toBlob(function (blob) {
+    //         myself.saveFileAs(blob, 'image/png', fileName, newWindow);
+    //     });
+    //     return;
+    // }
+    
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks Michael! Does this fix still open the pic in a new browser tab? (I really like and rely on that feature, rather than directly saving to disk)

cycomachead:
If the file is small enough, yes. If it's too big, it tries to download to disk.

jmoenig:
Thanks, that's perfect!

jmoenig:
Yep, this change fixes the issue for me. Will pull. Thanks, again, Michael, for the quick and successful investigation and fix!

jmoenig:
I'm also in good hopes that this fix will also take care of #1174

cycomachead:
> I'm also in good hopes that this fix will also take care of #1174

It should as well.

PR #1213, opened 2016-05-09
media import dialog with thumbnails

ubertao:
This PR changes import costumes/backgrounds menu to dialog boxes with thumbnails of media, making import process more intuitive/fun.

Given the limited size of current media library, pagination was not implemented, this is something should be improved in the future.

Screenshots:
import
import-bg

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -2597,6 +2597,69 @@ IDE_Morph.prototype.projectMenu = function () {
         };
     }
 
+    function importMediaDialog(mediaType) {
+        var dialog = new DialogBoxMorph().withKey('import' + mediaType),
+            frame = new ScrollFrameMorph(),
+            size = 500,
+            padding = 4, x = padding, y = padding,
+            selectedIcon = null,
+            turtle = new SymbolMorph('turtle', 60),
+            items = myself.getMediaList(mediaType);
+
+        frame.padding = 6;
+        frame.setWidth(size);
+        frame.acceptsDrops = false;
+        frame.contents.acceptsDrops = false;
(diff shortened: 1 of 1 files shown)


jmoenig:
Oh wow. This is really beautiful! Thank you, I'll pull this later today and integrate it right away!!

liam4:
This looks great! Will it cause an issue on devices with access to bad internet, though? If the preview image loading is synchronous, it may take these devices a long time to open up the media library.

ubertao:
Once the library file is read, the dialog will open right away with place holder images (currently the turtle icon, sorry I couldn't find a more proper resource for loading status), then thumbnails will be loaded in background.

cycomachead:
This is really awesome! 👍 

My only quibble is that the background color of the dialog looks off.

I'll add an issue about pagination later. Now that graphics are done, it'd be great to implement the same thing as a preview for the sounds dialog, where you get a grid of play buttons.

jmoenig:
Don't worry, @cycomachead , I've already fixed that, and I'm also working on a few other small issues...

jmoenig:
Thanks again, @ubertao ! I've tweaked your contribution somewhat, mostly making the dialog box resizable, the icons double-clickable and avoiding some causes of minor JS errors: https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/commit/711b9a431fa34fbfc49669ff2255b52ca201b5a5
But the original contribution and the design of it is still very much yours. I hope you don't mind, this is great, I really appreciate it!

ubertao:
Great, this is definitely better!

Issue #1216, opened 2016-05-10
add power function to math Operators


seattleacademy:
I teach high school math and programming classes, and am using Snap for the first time with my students.  The lack of a power function makes modeling things exponential growth/decay, geometric series, working with most power functions, etc. difficult.  Adding the equivalent of the javascript Math.pow(base,exponent) would make using Snap for the models we use in high-school math much easier.  I would suggest adding this either to the basic set of Operator blocks, or at least to the "official library of powerful blocks".
Note that for a workaround, we create a block that makes base^exponent =  e^(exponent \* ln(base)).  We also have tried making a javascript block that calls Math.pow(base,exponent).  However both workarounds turn a simple math lesson into a configuration exercises to use Snap for exploring math ideas that come up frequently in high school.

--Gary Anderson

jmoenig:
We left the exponent reporter out mostly because you can build it yourself (as you've done) and because Scratch also doesn't have it. However, I agree that it could just be in a library. @brianharvey , your call!

brianharvey:
This has long been the paradigmatic example of a feature left out on purpose because it's such a rich mine of student exploration, starting with an integer-exponent version as an easy recursive function, and working up to the full real-exponent version as students get into log and exponential functions.  But I agree that we could put it in a math library.  I've flagged this "discussion ongoing" because there are other such things, e.g., ccw-from-east directions, remove-one-copy-of-item-from-list, index-of-item-in-list, basically the whole Common Lisp primitive set.  Is "put it in a library" always the right answer to such requests?  Or are there problems so rich that we should avoid solving them for students?  (How about exact rationals, for example?  At Berkeley we deliberately degraded the division operator in our version of Scheme to give a decimal result for division of integers with a non-integer quotient, because SICP uses exact rational arithmetic as an extended example of data abstraction.)

cycomachead:
I think the request, and desire for a more powerful set of Math functions comes up. `Math.pow()` was the first thing I wrote as a JS Function!

I certainly agree that you _can_ gain lots of benefits by writing a power function yourself, and there's lots of cool math tricks to be explored in that process. I think the problem is that a lack of these functions can get in the way of students exploring and creating other projects. So it becomes about teaching priorities. 

In that sense, the library makes the most sense. (To me.) However, once you put something in a library, you remove a LOT of the motivation to build things yourself or to learn how to do so. I don't think you lose all the motivation though, it just requires more work than telling students "let's build a power function", which is hardly interesting to most people.


[...]

PR #1218, opened 2016-05-11
added preliminary basic-math.xml library for issue #1216

jordanei:
This is Jordan Isaacs, a ninth-grader of Gary Anderson's software development class at Seattle Academy. This is a preliminary library of blocks that are secondary-school math related. After your feedback, we can add 5-10 more blocks that lower the threshold of entry for middle and high school students to do math projects in Snap!.

### DIFF
--- a/libraries/LIBRARIES
+++ b/libraries/LIBRARIES
@@ -8,3 +8,4 @@ cases.xml	Multi-branched conditional (switch)
 leap-library.xml	LEAP Motion controller
 setrgb.xml	Set RGB or HSV pen color
 penTrails.xml	Save and restore pictures drawn by pen
+basic-math.xml	Additions for secondary-school math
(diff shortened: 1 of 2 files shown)


seattleacademy:
Good start, Jordan.  Other items for this library might be the javascript toFixed(digits) method, atan2(y,x), DegreesToRadians, RadiansToDegrees and xor.  Stat functions and matrix functions might also be considered as it would be useful, but the important thing is to consider what should be left "as an exercise for the reader".

cycomachead:
Nice!  👍 I'll have some minor comments about things we should do for libraries, but this is a good start.

Curious - how did you export the library code, or did you reformat the XML file at all? It doesn't make a difference, but more Snap<em>!</em> files are exported without any whitespace. (I don't think anything changed?)

jordanisaacs:
Thanks! I exported the XML file using Snap! and then reformatted it using Indent XML in Sublime Text 3.

Issue #1131, opened 2016-02-08
"Make a block"  should check for text  when  "OK" is clicked


kbakalar:
If the user opens the Make a Block dialog and then clicks "OK" without entering any text, the dialog should remain open and they should be reminded to enter the text.  As currently written, if the text is absent then "OK" and "Cancel" buttons have identical functions.

cycomachead:
Yeah, I agree this aspect is confusing.

Another solution is to simply make "OK" not be clickable, but this requires #649

kbakalar:
That would do as well. But it might not be clear to the user why the OK box is unclickable.  A message saying why an OK click is bad would be more helpful.

bobbybee:
Currently working on this.

ilayraz:
I am new to this project. I have been working on this bug for the last week, almost finished. Should I switch to another bug?

brianharvey:
Oh.  If you're almost finished, you might as well finish and file a pull request.  I should explain, though, that Jens is very, very fussy about what code he pulls, and also is on vacation for a few weeks, so nothing will happen about your PR right away, and in the end it might be rejected.  I am saying this not having seen your code, so don't be insulted; I'd say it to anyone.

Do we know you?  How did you come to be interested in contributing to Snap<i>!</i>?

ilayraz:
I was introduced to you through Sean Morris. You directed me to the issue tracker. I have been learning the code for a bit and started picking at this issue some time ago.

brianharvey:
Oh!  Right, I'm sorry, I remember about you, of course, but had forgotten your name.  Mea culpa.  Carry on!

kbakalar:
I have installed ilayraz's fix. It is just what I suggested above. But now I think I was wrong and cycomachead is right: The pop-up is too noisy.  Instead, the "OK" button should be grayed out. A gray "OK" is unambiguous in this context.

PR #1236, opened 2016-05-25
Bug fix to issue #1131

ilayraz:
"Make a block" checks for text when the "OK" button is clicked, and notifies the user if no text was inputted.

### DIFF
--- a/byob.js
+++ b/byob.js
@@ -1745,13 +1745,24 @@ BlockDialogMorph.prototype.fixLayout = function () {
 };
 
 BlockDialogMorph.prototype.accept = function () {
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks!

ilayraz:
A way to get around that is to check for InputFieldMorph in the BlockDialogMorph's children as well as check this.getInput().spec.

I'm not sure if that is the most elegant solution, however it works.

kbakalar:
I think it would be better to gray out "okay" until a name is entered rather than popping up a message window after "okay" is erroneously selected.

PR #1240, opened 2016-05-31
fixes #1239

### LINKED ISSUE #1239: deleted blocks still keep running
I've found a confusing issue using snap, reproduce the bug as below:
1. write snap script as below
image

2. click "green flag" to make it run
3. now, delete the running scripts's block (drag it to the left area)
4. on the stage, the script keeps running, you have to click the "stop button" to stop it. But for children, this is very confusing, the script is empty but the sprite is moving...
Suggestion: the script should stop when a block is running.

Reproduced with the online version snap: http://snap.berkeley.edu/snapsource/snap.html , 4.0.7.2

bromagosa:
It works both when dropping and ``right-click → delete``-ing whole scripts and script parts.

The ``topBlock`` check is needed because otherwise deleting just a part of a script would result in its process stopping.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -3484,8 +3484,9 @@ BlockMorph.prototype.destroy = function () {
     this.allComments().forEach(function (comment) {
         comment.destroy();
     });
-    
-    if (this.topBlock() === this && this.activeProcess()) {
+
+    if (!this.parent || !this.parent.topBlock
+            && this.activeProcess()) {
         this.activeProcess().stop();
     }
 
--- a/blocks.js
+++ b/blocks.js
@@ -3484,6 +3484,11 @@ BlockMorph.prototype.destroy = function () {
     this.allComments().forEach(function (comment) {
         comment.destroy();
     });
+    
+    if (this.topBlock() === this && this.activeProcess()) {
+        this.activeProcess().stop();
+    }
+
     BlockMorph.uber.destroy.call(this);
 };
 


blocks.js
```diff
@@ -3484,6 +3484,11 @@ BlockMorph.prototype.destroy = function () {
     this.allComments().forEach(function (comment) {
         comment.destroy();
     });
+    
+    if (this.topBlock() === this && this.activeProcess()) {
```
jmoenig: Just a really tiny quibbly, maybe totally irrelevant:
Looks like we're scanning for topBlock several times, especially when using the context menu, which can be a potential strain if the block is somewhere way at the bottom of a giant script (I've seen those ^^). Instead of scanning for the topBlock and comparing it to ```this``` one can't we just assume that if the current block either doesn't have a Morphic parent or that parent doesn't know the ```topBlock``` method (e.g. if it's a Hand or a ScriptsMorph), then it must be a topBlock itself?

```JavaScript
if (!this.parent || !this.parent.topBlock) {
        // I must be a top block myself
        this.activeProcess().stop();
}
```

Okay, up to you, fine with me :-)

bromagosa: Yep, that's right, we can assume that!

Actually, I thought calling ``topBlock()`` would be the right way because delete is a one-time operation, so this extra time wouldn't pile up. But yeah, it may be a problem in huge scripts. Now that I think about it, that could also be a warning that somebody's script is WAY TOO LONG, so they should be breaking it down into manageable pieces, but that's another discussion ;)

Let me update this commit.

jmoenig:
Thanks, Bernat! I'll pull this once I'm done with my current batch of new things for ```blocks.js``` (basically experimenting with GP style toggle switches for Boolean input slots, not sure how it's going to work out, but making progress...)

bromagosa:
Oh I love GP toggles!!! Can't wait to see it live! ^^

PR #1243, opened 2016-06-04
First class sounds

liam4:
Makes `Sound` objects work with the play sound primitives. My favourite way to use this is with the jukebox dev-only block :)

Anything to be done before a merge?

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -145,11 +145,11 @@ radians, useBlurredShadows, SpeechBubbleMorph, modules, StageMorph,
 fontHeight, TableFrameMorph, SpriteMorph, Context, ListWatcherMorph,
 CellMorph, DialogBoxMorph, BlockInputFragmentMorph, PrototypeHatBlockMorph,
 Costume, IDE_Morph, BlockDialogMorph, BlockEditorMorph, localize, isNil,
-isSnapObject, copy*/
+isSnapObject, copy, PushButtonMorph*/
 
 // Global stuff ////////////////////////////////////////////////////////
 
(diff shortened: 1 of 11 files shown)


liam4:
Yup - this still needs to be fixed:

issue

(It should be a list of sound icons.)

liam4:
Another idea - though the menu handling for the "of" block would probably need to be changed - is to be able to do things like "name" of (sound) and "duration" of (sound).

jmoenig:
Hmm... So, as you've discovered internally in the Snap code we have constructors mimicking classes of almost anything Snap is made of, and one aspect of making something 'first' class is simply to expose it to the IDE, which ususally only takes very little code, because the idea of a "class" *is* to make its instances "first class". But there's more to think about, and I'm currently not so much interested in making "everything" in Snap first class, at least not until Snap gets the object system I'm working on. Then all the components of the micro world can be represented as Snap objects, and we don't need to introduce special cases for each new "first class thingy" we want to expose.

That said your changes currently only affect what's visible in dev mode, and it's certainly fun to play around with these.

joefarebrother:
I suppose the function that displays things in speech bubbles and variable/list watcher could be refactored into a ".asMorph()" function to be defined in List, Sprite, Costume, Sound, etc, which would open the door to user defined objects being able to be displayed in these contexts, part of the way to making them fully first class

liam4:
Closed as the commits are all messed up now - if you need this again I can make a pull request from [`liam4:first-class-sounds`](https://github.com/liam4/Snap--Build-Your-Own-Blocks/tree/first-class-sounds).

PR #1244, opened 2016-06-04
Add missing graphic effects from Scratch

djdolphin:
This adds the color, fisheye, whirl, pixelate, and mosaic effects from Scratch. It also changes the brightness effect to use HSV and adds an additional saturation effect.


jmoenig:
Awesome and excellent, @djdolphin ! Finally somebody did it *right*, this is very much appreciated. I'll integrate this in the 4.0.8 release. Thank you so much!

PR #1249, opened 2016-06-09
Update Swedish translation with latest strings

mrpeachum:
This PR updates the existing Swedish translation with missing translations and corrects some mistakes in the old translation.


jmoenig:
Awesome, thank you so much! I'll pull this and integrate it into the next minor release (4.0.8), which might take another couple of weeks, so hang on!

PR #1250, opened 2016-06-09
Implements execute on change for boolean toggles

bromagosa:
This can be done in multiple ways, but since ``BooleanSlotMorph`` and ``InputSlotMorph`` inherit from ``ArgMorph``, I decided to move both the preference flag and the event handler up in the hierarchy. This leaves the door open for future widgets that can benefit from live changes too.

Still, now the ``executeOnSliderEdit`` name is a bit misleading because toggles aren't sliders. Should I rename it to something like ``executeOnChange`` or ``executeOnEdit``?

Please let me know if you like this solution or you would have handled it differently! :)

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -5799,6 +5799,40 @@ ArgMorph.prototype.init = function (type, silently) {
     this.setExtent(new Point(50, 50), silently);
 };
 
+// ArgMorph preferences settings:
+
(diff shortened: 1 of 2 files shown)


jmoenig:
Thanks, Bernat! I'll play with this when I get the chance to (now about to travel)...

PR #1259, opened 2016-06-20
Allow the new boolean block to be findable

cycomachead:
Necessary for the find bar and typing features

### DIFF
--- a/objects.js
+++ b/objects.js
@@ -868,7 +868,7 @@ SpriteMorph.prototype.initBlocks = function () {
             category: 'sensing',
             spec: 'current %dates'
         },
-        reportGet:{
+        reportGet: {
             type: 'reporter',
             category: 'sensing',
             spec: 'my %get',
@@ -971,7 +971,8 @@ SpriteMorph.prototype.initBlocks = function () {
         reportBoolean: {
             type: 'predicate',
             category: 'operators',
-            spec: '%bool'
+            spec: '%bool',
+            alias: 'true false boolean'
         },
         reportJoinWords: {
             type: 'reporter',


jmoenig:
Thanks!

brianharvey:
This is a good catch!  But I think the patch doesn't quite solve the problem.  It has to be that if you get the block by typing "true" then the block that appears is in the true state, whereas if you get it by typing "false" then it has to be in the false state.  Otherwise it won't work for blind users.  Right?

jmoenig:
Well, that we cannot do, unfortunately. We have to find a name for the block that describes that it's a Boolean reporter....

cycomachead:
Nothing works for blind users right now. We don't have a mechanism to support inserting a block in a specific state. 
I think we need to figure out a plan of action separate from this. 

Michael Ball
michaelball.co
From a small device...




On Mon, Jun 20, 2016 at 12:21 PM -0700, "Brian Harvey" <notifications@github.com> wrote:












This is a good catch!  But I think the patch doesn't quite solve the problem.  It has to be that if you get the block by typing "true" then the block that appears is in the true state, whereas if you get it by typing "false" then it has to be in the false state.  Otherwise it won't work for blind users.  Right?

brianharvey:
This could be a deal-killer for the combined block, imho.  Or, at least, we might have to have two versions of it in the palette, one in the true state and one in the false state.

cycomachead:
No. I think when we actually get to supporting blind users we can fix dropping in a block with different parameters.
 you can change the switch with the keyboard currently 
Michael Ball
michaelball.co
From a small device...




On Mon, Jun 20, 2016 at 12:33 PM -0700, "Brian Harvey" <notifications@github.com> wrote:












This could be a deal-killer for the combined block, imho.  Or, at least, we might have to have two versions of it in the palette, one in the true state and one in the false state.

jmoenig:
let me think about this...

cycomachead:
I think we are getting ahead of ourselves. IMO we shouldn't really implement anything designed for blind users until we have more of a plan. (One which includes audio feedback.)

Otherwise, the real issue here is searchability of different input slots. We don't yet have a way to search input slots (like drop down menus) which could be useful. That would be a separate feature which I think deserves its own thread. 

Question also worthy of its own thread: are the search aliases localized?

jmoenig:
Wait, I think I might have been wrong before. I think I can make it so there is only the ```true``` version in the palette, but both ```true``` and ```false``` are findable (and insertable) using the keyboard...

cycomachead:
I don't think we should special case this block just yet. 
Right now you can find both versions but the default (true) form is always interested. Then you can use the keyboard to change its state. This is consistent with all the other blocks. 
If we are going to support inserting blocks in different states we should also do this for dropdowns. 
Michael Ball
michaelball.co
From a small device...




On Mon, Jun 20, 2016 at 1:23 PM -0700, "Jens Mönig" <notifications@github.com> wrote:












Wait, I think I might have been wrong before. I think I can make it so there is only the true version in the palette, but both true and false are findable (and insertable) using the keyboard...

brianharvey:
Actually I just now finished meeting with a high school intern who's interested in the voice response thing.  I pointed him at Stefik.  Maybe nothing will come of it, we'll see.

I would argue for special casing this one just for backward compatibility, not with project but with users, who've gotten used to the two blocks.  But I'm not adamant about it.

cycomachead:
Can we PLEASE plan this thing out before we  jump into just working on stuff? We are running up against limitations of of Morphic"s design and it would be best to talk about what we want to get to and what types of use we are imagining.
As far as backwards compatibility - that's true but I don't know that many people currently uses the typing functionality to do most of their programming. For users who type then drag a block out there's little impact, since after they drop the block there's just one click. 
Michael Ball
michaelball.co
From a small device...




On Mon, Jun 20, 2016 at 4:59 PM -0700, "Brian Harvey" <notifications@github.com> wrote:












Actually I just now finished meeting with a high school intern who's interested in the voice response thing.  I pointed him at Stefik.  Maybe nothing will come of it, we'll see.



I would argue for special casing this one just for backward compatibility, not with project but with users, who've gotten used to the two blocks.  But I'm not adamant about it.

jmoenig:
Relax, Michael, I'm just talking about a small set of experiments once I've settled down from my current bout of travelling, which might take another two weeks :-)
I don't think we're really dealing with a special limitation of Morphic here, and the case of "true" and "false" does appear to be a somewhat special case, for some teachers the world hinges on spelling those out, preferably in English even :-). But I also agree that this is *not* the place to even begin planning screen reader support.

cycomachead:
I was referring to support for voice (which is more a browser thing than others) and screen readers which is a big problem as a result of using canvas. Our biggest potentional impacts are tablet users which require different interactions than just voice. 

Michael Ball
michaelball.co
From a small device...




On Mon, Jun 20, 2016 at 5:46 PM -0700, "Jens Mönig" <notifications@github.com> wrote:












Relax, Michael, I'm just talking about a small set of experiments once I've settled down from my current bout of travelling, which might take another two weeks :-)

I don't think we're really dealing with a special limitation of Morphic here, and the case of "true" and "false" does appear to be a somewhat special case, for some teachers the world hinges on spelling those out, preferably in English even :-). But I also agree that this is not the place to even begin planning screen reader support.

jmoenig:
Oh yeah, absolutely, nobody touches voice or any screen reader stuff over before we do a thorough and strategic planning of this!

dekrain:
Fun with the Developer's mode:
fun with dev mode
:P

PR #1288, opened 2016-07-07
Updating catalan translation

jguille2:
Hi,
I've updated catalan translation (agreed with @bromagosa ) and added new strings from german translation (updated on *2016-05-02*).

Thanks

Joan


jmoenig:
Thanks!

PR #1302, opened 2016-07-11
Update project title in document

### LINKED ISSUE #1301: Show Project Name in Page Titles
Approved by Brain! :)

https://scratch.mit.edu/discuss/topic/4464/?page=198#post-2049686

I second this idea, and completely forgot that it wasn't already here.

I suggest the format `Snap! | <ProjName>` or `-` instead of the pipe.

liam4:
Excuse my previous lack of organizing commits into branches, and the disgraceful reverting/moving of those commits, please - git is hard :)

This fixes #1301 in the format of `Snap! - <projname>`.

It also implements a couple of less specific functions, both in the `IDE_Morph` prototype:

```
updateDocumentTiitleText(title = this.projectName)
getDocumentTitleText(title = this.projectName)
```

(PS! You might want to [squash and merge](https://github.com/blog/2141-squash-your-commits) this so that all those old commits for first class sounds aren't there. Only the gui.js file has been changed in this pull request anyway.)

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -2597,6 +2597,69 @@ IDE_Morph.prototype.projectMenu = function () {
         };
     }
 
+    function importMediaDialog(mediaType) {
+        var dialog = new DialogBoxMorph().withKey('import' + mediaType),
+            frame = new ScrollFrameMorph(),
+            size = 500,
+            padding = 4, x = padding, y = padding,
+            selectedIcon = null,
+            turtle = new SymbolMorph('turtle', 60),
+            items = myself.getMediaList(mediaType);
+
+        frame.padding = 6;
+        frame.setWidth(size);
+        frame.acceptsDrops = false;
+        frame.contents.acceptsDrops = false;
+        frame.setHeight(size);
+        frame.fixLayout = nop;
+
+        items.forEach(function (item) {
+            var url = myself.resourceURL(mediaType, item.file),
+                img = new Image();
+            var icon = new CostumeIconMorph(new Costume(turtle.image, item.name));
+            icon.setPosition(new Point(x, y));
+            icon.userMenu = null;
+            icon.action = function () { 
+                if (selectedIcon === icon) return;
+                var prevSelected = selectedIcon;
+                selectedIcon = icon;
+                if (prevSelected) prevSelected.refresh();
+            };
+            icon.query = function () {
+                return icon === selectedIcon;
+            };
+            frame.addContents(icon);
+            x = icon.right() + padding;
+            if (x + icon.width() > frame.width()) {
+                x = padding;
+                y = icon.bottom() + padding;
+            }
+            img.onload = function () {
+                var canvas = newCanvas(new Point(img.width, img.height));
+                canvas.getContext('2d').drawImage(img, 0, 0);
+                icon.object = new Costume(canvas, item.name);
+                icon.rawImage = canvas;
+                icon.refresh();
+            };
+            img.src = url;
+        });
+
+        dialog.ok = function () {
+            myself.droppedImage(selectedIcon.rawImage, selectedIcon.labelString);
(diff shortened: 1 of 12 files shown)


gui.js
```diff
@@ -1552,9 +1552,24 @@ IDE_Morph.prototype.fixLayout = function (situation) {
 IDE_Morph.prototype.setProjectName = function (string) {
     this.projectName = string.replace(/['"]/g, ''); // filter quotation marks
     this.hasChangedMedia = true;
+    this.updateDocumentTitleText();
     this.controlBar.updateLabel();
 };
 
+IDE_Morph.prototype.updateDocumentTitleText = function (string) {
+    var text = this.getDocumentTitleText(string || this.projectName);
+    title = document.getElementsByTagName('title')[0];
```
jmoenig: okay, I admittedly suck at web programming, so I don't know better. Can you explain to me why we can't just write

```JS
    document.title = 'Snap! - ' + this.projectName;
```

(?) Thanks!

liam4: Oh.

I don't program webpages that change much.. :)

gui.js
```diff
@@ -3703,6 +3714,7 @@ IDE_Morph.prototype.rawOpenProjectString = function (str) {
         );
     }
     this.stopFastTracking();
+    this.updateDocumentTitleText();
```
jmoenig: I'm not sure, do we really need to set the document title again just because we're resizing a canvas element on the page?

gui.js
```diff
@@ -1552,9 +1552,20 @@ IDE_Morph.prototype.fixLayout = function (situation) {
 IDE_Morph.prototype.setProjectName = function (string) {
     this.projectName = string.replace(/['"]/g, ''); // filter quotation marks
     this.hasChangedMedia = true;
+    this.updateDocumentTitleText();
```
jmoenig: So, my hunch is, we can just write a single line with an assignment statement here, instead of introducing two additional functions that we don't ever call anywhere else...

cycomachead:
There should probably be a single function for this, but I think only one function is necessary.

And yes, we should squash these commits when we merge the PR.

gui.js
liam4: Hrm.

I was sure I put this inside the load project from string function..

gui.js
cycomachead: I think we should have a single function, not just a statement. The reason is that we will likely need to call this function in at least 2 or 3 places, and we might want to change the display. Secondly, for applications like BettleBlocks or others it will be easier for them to override a short, simple function rather that every instance where a project name might be displayed.

jmoenig: Okay. Where else would we need to change the title other than when we're opening a project?

cycomachead: Off the top of my head, this is the list. Note, I'm thinking of user actions, not code paths, so some of these might actually call the same functions already:

* We need to be sure we _don't_ switch the title when changing languages of untitled projects.
* opening a project
* saving a new project for the first time
* exporting a new project
* Save As

jmoenig: Thanks!

gui.js
liam4: I *did!* Look at the lines of code before `);` - it's not *actually* inside `setExtent`, that's just where GitHub (git?) put the diff. Inconveniently.

Image

liam4:
For some reason when I change the language the project gets entirely reset.. it doesn't do this on the master repo, and the only differences between this and that are - as far as I know - these.. I hope nothing broke when I was [messing with git!](http://i.cubeupload.com/7Hmu2H.png)

cycomachead:
I think this is useful, and one of those really nice things - and super helpful if you're working with multiple Snap! Instances. (The fact that few sites use meaningful page titles is painful).

But Jens' call :)

PR #1330, opened 2016-07-19
Fix a few Integer 0 vs String Errors.

### LINKED ISSUE #1322: Length of Reports 0 (not 1) for numeric 0 input.
length

screen shot 2016-07-15 at 3 28 33 pm

screen shot 2016-07-15 at 3 30 15 pm

cycomachead:
I also renamed the internal variables to be slightly more generic, but
now the function reads better. (IMO)

Fix #1322

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -2407,7 +2407,7 @@ Process.prototype.reportLetter = function (idx, string) {
         return '';
     }
     var i = +(idx || 0),
-        str = (string || '').toString();
+        str = isNil(string) ? '' : string.toString();
     return str[i - 1] || '';
 };
 
@@ -2416,7 +2416,7 @@ Process.prototype.reportStringSize = function (data) {
         return data.length();
     }
 
-    return (isNil(data) ? '' : data).toString().length;
+    return isNil(data) ? 0 : data.toString().length;
 };
 
 Process.prototype.reportUnicode = function (string) {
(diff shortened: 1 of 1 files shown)


threads.js
```diff
@@ -2411,12 +2411,12 @@ Process.prototype.reportLetter = function (idx, string) {
...
-    if (string instanceof List) { // catch a common user error
-        return string.length();
+Process.prototype.reportStringSize = function (data) {
+    if (data instanceof List) { // catch a common user error
+        return data.length();
     }
-    var str = (string || '').toString();
-    return str.length;
+
+    return (isNil(data) ? '' : data).toString().length;
```
cycomachead: It seems to be that these checks are better written explicitly as `isNil`? Is there any reason to prefer the other style other than it being a few less characters?

In this case, `isNil` is the correct choice, but I'm wondering if it would make sense to mostly banish the `||` pattern, since I know we've been bitten by this before.

jmoenig: how about return zero instead of an empty string in case data is nil? I'd like to make sure that ```length of``` always returns a number...

cycomachead: `.length` is still there, so this should always return a number, no?

jmoenig: ```isNil``` certainly makes sense here, I agree!

cycomachead: Though there are these awkward `()` that I can just avoid...

cycomachead:
Also, I found a related bug in `letter of`

threads.js
jmoenig: Oh duh, you're right! :-)

cycomachead:
I have found more in the `split` function.

I hearby propose that we declare `varname || ''` an antipattern.

Issue #235, opened 2013-11-19
Snap! doesn't play sounds on mobile devices


ddgarcia:
http://tinyurl.com/bjcmolemash

...plays wonderful sounds on all computers but on no mobile devices. Can this be fixed for the Hour of Code asap?

Thanks!

dan

cycomachead:
Updating this after a long while:
There might be a solution if we could somehow initialize the sounds on each tap event. (Like a hat block / green flag.)

https://twitter.com/cycomachead/status/758212145284730881

sinuhe69:
Yes, I can confirm that the issue still exists on mobile devices. I tested on 2 Android devices (Kitkat 4.2.1 and 4.4), the "Play Sound" block doesn't work in Chrome or Firefox for Mobile (both latest versions). However, when I click on the Play button, I can hear the sound without any problem! It's really weird :(

Demo here: https://goo.gl/rZlO2O

PR #1343, opened 2016-07-26
Add missing sound features from Scratch

djdolphin:
This adds instruments, drums, volume blocks, and the loudness block. It seems to work in every major browser except for mobile Safari. I still have no idea why. :(

I've moved sound-related code to a separate file since objects.js is already *really* long. I haven't done the header because there seems to be a pretty specific format, and I don't want to mess it up. :)


gui.js
```diff
@@ -5009,22 +5023,28 @@ IDE_Morph.prototype.setCloudURL = function () {
...
-        request.open('GET', url, false);
-        request.send();
+
+    if (!errorCall) {
+        errorCall = function () {
+            myself.showMessage('Error: unable to retrieve ' + url);
+        };
+    }
+
+    request.open('GET', url, true);
```
cycomachead: Shouldn't we just leave of `true` since it's the default mode?

cycomachead:
WooooooHooo!!

Async loading! Yes!

Minor comments from skimming:
- [ ] The default block comment should be at the top of the sounds.js
- [ ] `soundbank/` → `Sounds/` ? (keeps with the rest of our folders)



> except for mobile Safari. I still have no idea why. :(

Unfortunately, we have significant limitations in Safari to deal with. AFAIK, Mobile Safari will only play audio as a direct result of a user action.

jmoenig:
Aw, man. A giant PR that bundles lot of individual changes into one big package that I have to either accept or reject in total. Glancing cursorily over the code it looks nice, but this is going to keep me tied up forever, once I've managed to catch up with all the retina issues that still keep surfacing and chasing other issues that also keep coming up. Couldn't this have been broken up into smaller pieces? Couldn't asynchronous resource loading have been parcelled out, same as microphone support and volume? The way it is I won't be able to pull this for a long while, because it's going to draw all my attention to follow-up issues :-(

cycomachead:
> Couldn't this have been broken up into smaller pieces?

Yeah, this would actually be really easy to tease apart. At least, the async stuff.

objects.js
```diff
@@ -2780,6 +2826,23 @@ SpriteMorph.prototype.reportSounds = function () {
     return this.sounds;
 };
 
+// SpriteMorph volume
+
+SpriteMorph.prototype.setVolume = function (volume) {
+    if (!isNaN(volume)) {
```
dekrain: Shouldn't be `if (!isNaN(+volume)) {` ?

djdolphin:
Sure thing, I'll separate these and address those other things in the new pull requests. They should probably be added in this order, since they sort of depend on each other:
- the async stuff
- instruments and drums
- volume and/or loudness.

cycomachead:
Tagging #235 

https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/issues/235#issuecomment-235678875

PR #1371, opened 2016-08-14
Added Better Lists Utilities

dekrain:


cycomachead:
Can you explain what is better, or what changed? Screenshots or any easy link to test would be welcome. 

Also, what's with the diff on `libraries/LIBRARIES`?

brianharvey:
Sorry, but I'm not convinced that any of these are improvements.  You are basically trying to turn linked lists into arrays, which will slow things down as they are converted back and forth.  You have a ϴ(n²) reverse algorithm, if I'm understanding it correctly.  Etc.

PR #1377, opened 2016-08-19
Add CONTRIBUTING.md

tjvr:
These guidelines will be linked from [above the new issue page](https://github.com/blog/1184-contributing-guidelines).

@cycomachead Thoughts?

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,11 @@
+If you think you've found a bug in Snap!, [ask in **the Snap! forum**](http://snapforum.cs10.org/t/about-bugs-development) to confirm whether it's a known bug (or just a mistake in your project!).
+
+Make sure you include:
+
+* What the bug is
+* How to trigger it
+* Your browser & computer (eg. Chrome on Mac OS X)
+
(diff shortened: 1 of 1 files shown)


cycomachead:
Ah, does this replace the default `CONTRIBUTING.md` top-level file?

I don't mind this, I think we really ought to just clean up the docs files. (See #425 and associated links)

Jens doesn't seem to like the idea of having more directories in the repo. But maybe if it's hidden, then he won't have to see it. 😉

cycomachead:
Also, if we open up the forums we should just add a link from the README.

tjvr:
> does this replace the default CONTRIBUTING.md top-level file?

It's the same thing; GitHub allows you to have it either in `CONTRIBUTING` or `CONTRIBUTING.md` or `.github/CONTRIBUTING` or...

cycomachead:
Cool.. I know they added issue templates (which would be good at some point), but I guess I didn't realize this also worked.

Jonathan50:
There's a Snap*!* forum?

cycomachead:
> There's a Snap! forum?

It's a WIP - it should be public soon.

Issue #1089, opened 2015-12-29
Go back n layers, tiny misfeature if n > max possible


brianharvey:
Actually I'm not sure if it's a misfeature, maybe it's a bug, since I think it does what I want in Scratch:  If you say "go back 999 layers" the sprite goes as far back as possible.  In Snap! nothing happens.

liam4:
That's probably just a misfeature, but it might just be better to make the number input to a dropdown and add an option "all", i.e. `go back (all v) layers`.

jmoenig:
excellent suggestion, @liam4 !

liam4:
Plus it wouldn't break old projects, unlike `go to [front/back v]` because of adding an input.

jmoenig:
done!

PR #1390, opened 2016-08-26
'go to back' on large num in 'go back' block. Fixes #1207

### LINKED ISSUE #1207: go to back block
There is a `go to front` block but no `go to back` block; if someone tries to emulate a `go to back` block using the `move back ___ layers` w/ a large number, the block will nop if the number of layers is too high.

A simple fix could be to have the `move back ___ layers` block move back `Math.min(totalLayers, requestedLayers)` rather than ignoring requests with a request higher than the total number of layers

brollb:


cycomachead:
Also fix #1089

jmoenig:
Wait! How could I miss this? Oh, I see, @brollb issued this when I was on vacation. Geez, this is nice, thank you!

jmoenig:
removing the *bug* label and adding *blessed*, so I don't forget to pull it :-)

PR #1399, opened 2016-09-03
Russian (minor changes)

joker314:
I made stop become ostonovit' because "stop" is what you might shout to ask somebody to stop, but I don't think it makes grammatical sense when combines with "this script" and so on. It would be understood in both cases though, and I suppose I just made the translation longer, feel free to not merge.

As for e -> yo in the translation of "all", I'm pretty sure that the correct one is what I did, but since the characters look so similar, people often use "e" in both cases.

Also, I am wondering, will the help files have translations at any point?


joker314:
Also there are many more cases of e -> yo, should we change them all?

jmoenig:
Oops, how could I miss this PR? Must have occurred during my summer vacation. Sorry, will pull asap for the next minor release upcoming around Christmas!

PR #1417, opened 2016-09-22
Added check that parent is RingMorph. Fixes #1416

### LINKED ISSUE #1416: Can't unringify empty RingMorph
This is minor but when trying to unringify an empty RingMorph, the following error is thrown:

```
Uncaught TypeError: Cannot read property 'parent' of null @ morphic.js:3377 
```
Also, when clicking on a single, empty RingMorph, "unringify" is an option when it doesn't really make sense as the single RingMorph is not inside a ring.

brollb:
When providing the "unringify" option, it was checking that
the given block has a parent that is a RingMorph. However,
`parentThatIsA` will return the given block if it is the given
type when, in this context, we don't want to check the target of
the ringification.

Also fixed TypeError when unringify-ing an empty RingMorph caused
by a similar issue w/ getting the containing ring of an empty RingMorph

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -2430,7 +2430,7 @@ BlockMorph.prototype.userMenu = function () {
         }
         return menu;
     }
-    if (this.parentThatIsA(RingMorph)) {
+    if (this.parent.parentThatIsA(RingMorph)) {
         menu.addLine();
         menu.addItem("unringify", 'unringify');
         menu.addItem("ringify", 'ringify');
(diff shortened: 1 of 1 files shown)


cycomachead:
This makes sense to me, but that code just looks weird. 😛

jmoenig:
I love your bug reports that come with a fix, Brian! Thank you, I will pull this and integrate it into the next release (which, you know, could take some more weeks).

brollb:
No problem - always happy to contribute!

PR #1419, opened 2016-09-23
Control / ⌘ key + arrow key lets us navigate through words now.

bromagosa:


jmoenig:
Thanks, Bernat! I'll pull it with the next batch of changes.

show0k:
In OSX, it is not the command key which is usually used to navigate throw words, but the alt key !

bromagosa:
Thanks @show0k, I've updated the pull request accordingly...

jmoenig:
Hi, Théo - @show0k - Wow, I didn't even realize the text navigation shortcuts on a Mac, but I still remember the ones I used on Windows for many years :-)

So, the alt-key navigates among words, and the cmd-key goes to the beginning or end of line, right?

show0k:
`So, the alt-key navigates among words, and the cmd-key goes to the beginning or end of line, right?`
Exactly !

liam4:
On the topic of these navigation controls, ctrl-A goes to the beginning of the line (or paragraph, depending on the environment), and ctrl-E goes to the end.

PR #1430, opened 2016-09-27
Automatic indentation for newlines following already indented lines in multiline inputs

liam4:
Sorry for the long title!

[Here's a demo!](https://youtu.be/bXjRz2UaBrM) Note that newlines after an already indented line will be indented as well:

```
foo<enter>
<enter>
  bar<enter>
  <cursor>
```

It works for any multiline input, so for example, it also works for project description inputs:

Project description demo

It's handy for writing long pieces of code. Right now you have to press space at least once for each level of indentation you want in your code.

It _does not_ behave like your text editor, which probably automatically indents things when you press enter after an opening curly brace, or de-indents when you add a closing curly brace, but I do think this is fairly helpful and it's simple enough it can be applied to any multiline text input.

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -5108,7 +5108,18 @@ CursorMorph.prototype.processKeyDown = function (event) {
         if ((this.target instanceof StringMorph) || shift) {
             this.accept();
         } else {
+            var cursor = this.target.root().cursor;
+            var oldSlot = cursor.slot;
+            cursor.goHome();
+            var line = this.target.text.slice(cursor.slot).split('\n')[0];
+            cursor.gotoSlot(oldSlot);
+
+            var indent = 0;
             this.insert('\n');
+            while (line[indent] === ' ') {
+                indent++;
+                this.insert(' ');
+            }
         }
         this.keyDownEventUsed = true;
         break;


morphic.js
```diff
@@ -5108,7 +5108,18 @@ CursorMorph.prototype.processKeyDown = function (event) {
         if ((this.target instanceof StringMorph) || shift) {
             this.accept();
         } else {
+            var cursor = this.target.root().cursor;
+            var oldSlot = cursor.slot;
+            cursor.goHome();
+            var line = this.target.text.slice(cursor.slot).split('\n')[0];
+            cursor.gotoSlot(oldSlot);
```
liam4: Any way to get the current line that's simpler than this? It feels a bit long.

BookOwl:
This will be _really_ nice for working with JSFunctions

cycomachead:
Wait - the project text in the open dialog is _editable_...wow. I never knew that...

jmoenig:
no, @cycomachead , only the text in the *save* dialog is editable.

Anyway, I'm closing this because I keep missing important pull-requests and I don't want to turn multi-line text into a programming IDE.

cycomachead:
Heh it's basically the same dialog. :)  but I never know that either!

brianharvey:
Not for the imminent release, but I vote for this PR eventually.  Especially if a delete just after the enter deletes all the spaces.  I also vote for tab = 4 spaces.

cycomachead:
> I also vote for tab = 4 spaces.

This doesn't do anything with the tab character. It simply makes it so that each new line has the same number of spaces as the one above it.

I don't think you can tab in a text input slot since tab leaves focus...

brianharvey:
I bet we can have multiline input boxes capture tabs.  Other sites let you enter programs. :)

jmoenig:
no way.

brianharvey:
Umm, okay, but could you explain why not?

jmoenig:
I want our input slots to be just that, mostly for story telling, not for programming. No advanced text-coding support, this is a blocks-based IDE. Also, I'm getting increasingly nervous because our recently added and prioritized features exhibit a disproportional code/functionality ratio. Way too much code and way too much complication for way too little value. Snap is starting to feel like MS Windows.

cycomachead:
> I want our input slots to be just that, mostly for story telling, not for programming. No advanced text-coding support, this is a blocks-based IDE.

I think there are plenty of users who want to write text in input slots for literal story telling. This doesn't apply just to writing code, this applies to everywhere someone might want to write more text. (Now if we added support for auto-indentations based on open `{ ( [`, I would agree that's not as necessary.)



> Also, I'm getting increasingly nervous because our recently added and prioritized features exhibit a disproportional code/functionality ratio. Way too much code and way too much complication for way too little value. Snap is starting to feel like MS Windows.

Perhaps we could take a step back and look at the last few years of issues/requests and see where things stand on priorities vs code, but on the whole -- I think we're running into cases of the so called "80 / 20 rule" .

brianharvey:
It would be easier to discuss this with respect to specific features.  What did you have in mind?

In some cases, such as Bernat's recent improvements to text editing, it's true that they don't make something possible that used to be not possible, so there's a sense in which there's no payoff.  On the other hand, I think it's hugely important to meet users' expectations from other software, including other web sites.  So this extra code is part of the cost of using Morphic instead of the DOM.

I could have lived without the ability to type arithmetic expressions into the search bar, but given that we're going to do that, I think the extra code to do it properly is important, not a waste of space.

Hierarchical menus?  I don't think it was that much code; Morphic's existing structure made it really easy, and if I'd felt more comfortable about adding an item to that seven-item list it'd be even less code.  And, apart from my immediate use case, I _love_ hierarchical menus!  I paid money (twice, because MacOS 10.10 broke the old program) for the ability to select a particular preference pane from the Mac menu bar by putting a submenu under the System Preferences item.

And, Michael is right, apart from OOP we already have all the really killer features in Snap<i>!</i>.  From now on it's mostly going to be polishing rough corners, plus adding the things that are killer features for _us_ but probably not for most users, such as self-reflection.  Well, no, plus Akos and the other Brian's net collaboration stuff.  That'll be killer for some real users.

cycomachead:
It might be best to have a separate thread for this but I agree with Brian on these examples. There are other things, like table lists, that I personally see as not as much bang-for-the-code-buck, but others seem to find them useful. :)

In general though, I agree with the point about meeting users' expectations and doing things properly. Regardless of whether it's right, my experience both as a teacher and a user of many programs is that doing something half way can be far more frustrating than doing nothing. Obviously, that doesn't apply to every feature, but I have seen many students drop CS10 or not recommend the course or simply just lose interest due to various Snap! oddities.

bromagosa:
IMO there are still things where we're behind Scratch, and I think those should also be priorities, because they're making kids and teachers stay away from Snap!.

I know we generally aim for an older audience, but there's no harm at all in adding a vector editor, multiple instruments, piano (and other sound controls), webcam blocks (they're [on the way](https://bromagosa.github.io/Snap/snap.html)!), sound editing, backpack and, of course, social stuff. That includes the forum, where this conversation should be happening.

brianharvey:
Oh, +1, you should do all that stuff.  But this isn't a feature request, it's a PR from a volunteer, and after things slow down (if ever!) we should pull it.

JSFunction, about which I was very doubtful when Jens invented it, is turning out to be very popular with two categories of users: (1) hotshot kids, who want to program in a 'grownup' language, but still want to be in our community; (2) adults who want a feature that will never make its way to the top of our priority list, and who can't do it themselves, but can use an implementation from one of the hotshot kids.  Neither of these groups is exactly our core audience, but they're not unimportant either.

@liam4, would it make sense to do this in the multiline library?  @jmoenig, would you hate that more or less than just pulling it?

jmoenig:
I closed this pull-request, because I don't want this feature. Nowadays you can copy & paste text into and out of JS-function blocks and use your favorite text editor with syntax highlighting and auto-complete etc. for editing text code. No need to add that stuff to Snap itself.

PR #1432, opened 2016-09-27
Add check for arguments reporter

dekrain:


dekrain:
If you use variable `arguments` without declare it (for example: `call {(arguments)} with inputs: [1], [2], [3]`, it reports [1, 2, 3]), it returns arguments passed in `call` block.

PR #1449, opened 2016-10-06
Use userSetSpec for var rename. Fixes #1448

### LINKED ISSUE #1448: Renaming variable can result in incorrectly sized block
If you create a variable and rename it to a name that is longer than the current one, the block does not update appropriately. Below I renamed `var` to `dddddddd`:
badsize

brollb:


jmoenig:
Good catch, thanks for catching and fixing this!

brollb:
no problem!

PR #1450, opened 2016-10-06
Fixed minor typo (anser -> answer)

brollb:
This is pretty minor but I noticed it so I figured I would make a PR


jmoenig:
another good find, thanks, Brian! We should appoint you to be Snap's editor-in-chief :-)
(naw, honestly, I really appreciate your diligence!)

PR #1453, opened 2016-10-07
Fixed minor costume editing discrepancy

brollb:
Prior to this change, it was unclear whether `anIDE` was a required argument for the `edit` method of `Costume`. [This line](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/objects.js#L6947) seems to suggest that it is an optional argument; however, that cannot be the case since [this line](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/objects.js#L6941) will fail if `anIDE` is not provided.

This modification makes `anIDE` optional by moving the failing line within the check for `anIDE`.


jmoenig:
Thanks!

PR #1457, opened 2016-10-14
text editing improvements

bromagosa:
Same changes as in morphic.js, plus removed `selectAll()` call upon click.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -7529,7 +7529,6 @@ InputSlotMorph.prototype.mouseDownLeft = function (pos) {
         this.escalateEvent('mouseDownLeft', pos);
     } else {
         this.contents().edit();
-        this.contents().selectAll();
     }
 };
 
@@ -7540,7 +7539,6 @@ InputSlotMorph.prototype.mouseClickLeft = function (pos) {
         this.dropDownMenu();
     } else {
         this.contents().edit();
-        this.contents().selectAll();
     }
 };
 
--- a/morphic.js
+++ b/morphic.js
@@ -1245,6 +1245,11 @@ function fontHeight(height) {
     return minHeight * 1.2; // assuming 1/5 font size for ascenders
 }
 
+function isWordChar(aCharacter) {
+    // can't use \b or \w because they ignore diacritics
+    return aCharacter.match(/[A-zÀ-ÿ0-9]/);
+}
+
 function newCanvas(extentPoint, nonRetina) {
     // answer a new empty instance of Canvas, don't display anywhere
     // nonRetina - optional Boolean "false"
@@ -4986,6 +4991,9 @@ CursorMorph.prototype.initializeClipboardHandler = function () {
         'keydown',
         function (event) {
             myself.processKeyDown(event);
+            if (event.shiftKey) {
+                myself.world().currentKey = 16;
+            }
             this.value = myself.target.selection();
             this.select();
 
@@ -4998,7 +5006,15 @@ CursorMorph.prototype.initializeClipboardHandler = function () {
         },
         false
     );
-    
+
+    this.clipboardHandler.addEventListener(
+        'keyup',
+        function (event) {
+            myself.world().currentKey = null;
+        },
+        false
+    );
+
     this.clipboardHandler.addEventListener(
         'input',
         function (event) {
@@ -5056,28 +5072,47 @@ CursorMorph.prototype.processKeyPress = function (event) {
 
 CursorMorph.prototype.processKeyDown = function (event) {
     // this.inspectKeyEvent(event);
-    var shift = event.shiftKey;
+    var shift = event.shiftKey,
+        wordNavigation = event.ctrlKey || event.altKey,
+        selecting = this.target.selection().length > 0;
+
     this.keyDownEventUsed = false;
     if (event.ctrlKey && (!event.altKey)) {
         this.ctrl(event.keyCode, event.shiftKey);
         // notify target's parent of key event
         this.target.escalateEvent('reactToKeystroke', event);
-        return;
     }
     if (event.metaKey) {
         this.cmd(event.keyCode, event.shiftKey);
         // notify target's parent of key event
         this.target.escalateEvent('reactToKeystroke', event);
-        return;
     }
 
     switch (event.keyCode) {
     case 37:
-        this.goLeft(shift);
+        if (selecting && !shift && !wordNavigation) {
+            this.gotoSlot(Math.min(this.target.startMark, this.target.endMark));
+            this.target.clearSelection();
+        } else {
+            this.goLeft(
+                    shift,
+                    wordNavigation ?
+                        this.slot - this.target.previousWordFrom(this.slot)
+                        : 1);
+        }
         this.keyDownEventUsed = true;
         break;
     case 39:
-        this.goRight(shift);
+        if (selecting && !shift && !wordNavigation) {
+            this.gotoSlot(Math.max(this.target.startMark, this.target.endMark));
+            this.target.clearSelection();
+        } else {
+            this.goRight(
+                    shift,
+                    wordNavigation ?
+                        this.target.nextWordFrom(this.slot) - this.slot
+                        : 1);
+        }
         this.keyDownEventUsed = true;
         break;
     case 38:
@@ -5121,6 +5156,11 @@ CursorMorph.prototype.processKeyDown = function (event) {
         // this.inspectKeyEvent(event);
     }
     // notify target's parent of key event
+    this.target.escalateEvent('reactToKeystroke', event);
+};
+
+CursorMorph.prototype.processKeyUp = function (event) {
+    this.world().currentKey = null;
     this.target.escalateEvent('reactToKeystroke', event);
 };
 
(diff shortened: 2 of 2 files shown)


morphic.js
```diff
@@ -4986,6 +4991,9 @@ CursorMorph.prototype.initializeClipboardHandler = function () {
         'keydown',
         function (event) {
             myself.processKeyDown(event);
+            if (event.shiftKey) {
+                myself.world().currentKey = 16;
```
jmoenig: Oh, I see. We're checking for ```shiftPressed``` on ```mouseDown```, right? Is that why we need this?

morphic.js
```diff
@@ -4998,7 +5006,15 @@ CursorMorph.prototype.initializeClipboardHandler = function () {
         },
         false
     );
-    
+
+    this.clipboardHandler.addEventListener(
+        'keyup',
+        function (event) {
```
jmoenig: Hmm....  I think I can see why we need this, but I'm not happy with crawling up the Morphic tree all the way to the root on every single keystroke. Maybe we can capture the ```World``` in a closure somewhere. Need to think about this. I hate shadowing the DOM. See, Web-Programming is all evil :-)

morphic.js
```diff
@@ -5056,28 +5072,47 @@ CursorMorph.prototype.processKeyPress = function (event) {
 
 CursorMorph.prototype.processKeyDown = function (event) {
     // this.inspectKeyEvent(event);
-    var shift = event.shiftKey;
+    var shift = event.shiftKey,
```
jmoenig: Maybe we can move setting the ```World```'s ```currentKey``` attribute to here. Dunno if that makes it any better...

morphic.js
```diff
@@ -5056,28 +5072,47 @@ CursorMorph.prototype.processKeyPress = function (event) {
...
+        wordNavigation = event.ctrlKey || event.altKey,
+        selecting = this.target.selection().length > 0;
```
jmoenig: little quib: How about getting right of the ```> 0``` part, because if the ```length``` is zero, ```selecting``` will be falsey already, right? :-)

morphic.js
```diff
@@ -5124,6 +5159,11 @@ CursorMorph.prototype.processKeyDown = function (event) {
     this.target.escalateEvent('reactToKeystroke', event);
 };
 
+CursorMorph.prototype.processKeyUp = function (event) {
+    this.world().currentKey = null;
```
jmoenig: wait! Aren't we doing that already in the ```ClipboardHandler```? Maybe we don't do it there and only here? Would that work?

morphic.js
```diff
@@ -5213,7 +5253,7 @@ CursorMorph.prototype.gotoPos = function (aPoint) {
 
 CursorMorph.prototype.updateSelection = function (shift) {
     if (shift) {
-        if (!this.target.endMark && !this.target.startMark) {
+        if (isNil(this.target.endMark) && isNil(this.target.startMark)) {
```
jmoenig: is this because we now somehow have zero as a valid ```startMark``` or ```endMark```?

morphic.js
```diff
@@ -7837,7 +7879,14 @@ StringMorph.prototype.slotAt = function (aPoint) {
             }
         }
     }
-    return idx - 1;
+
+    // see where our click fell with respect to the middle of the char
+    if (aPoint.x - this.left() >
```
jmoenig: Nice! This is cool, thanks!

morphic.js
```diff
@@ -7860,6 +7909,41 @@ StringMorph.prototype.endOfLine = function () {
...
+StringMorph.prototype.previousWordFrom = function (aSlot) {
+    // answer the slot (index) slots indicating the position of the
+    // previous word to the left of aSlot
+    var index = aSlot - 1;
+    
+    // while the current character is non-word one, we skip it, so that
+    // if we are in the middle of a non-alphanumeric sequence, we'll get
+    // right to the beginning of the previous word
+    while (index > 0 && !isWordChar(this.text[index])) {
+        index --;
```
jmoenig: another little quib: Some of the static code analysis tools I use dislike shifting integers and complain about this. Can we change it to something a little bit more explicit, like ```index -= 1; ``` ? Thanks! (And sorry for the hassle)

morphic.js
```diff
@@ -7860,6 +7909,41 @@ StringMorph.prototype.endOfLine = function () {
...
+    }
+
+    // while the current character is a word one, we skip it until we
+    // find the beginning of the current word
+    while (index > 0 && isWordChar(this.text[index - 1])) {
+        index --;
```
jmoenig: same here...

morphic.js
```diff
@@ -7860,6 +7909,41 @@ StringMorph.prototype.endOfLine = function () {
...
+    }
+
+    return index;
+};
+
+StringMorph.prototype.nextWordFrom = function (aSlot) {
+    var index = aSlot;
+    
+    while (index < this.endOfLine() && !isWordChar(this.text[index])) {
+        index ++;
```
jmoenig: ... and here (```index += 1; ```)

morphic.js
```diff
@@ -7860,6 +7909,41 @@ StringMorph.prototype.endOfLine = function () {
...
+    }
+
+    while (index < this.endOfLine() && isWordChar(this.text[index])) {
+        index ++;
```
jmoenig: ... you guessed it... :-)

morphic.js
```diff
@@ -8047,22 +8131,45 @@ StringMorph.prototype.deleteSelection = function () {
 };
 
 StringMorph.prototype.selectAll = function () {
+    var cursor;
     if (this.isEditable) {
         this.startMark = 0;
+        cursor = this.root().cursor;
```
jmoenig: Whoa! Why do we need to move the cursor here? What if we have several ```StringMorph```s or ```TextMorph```s with different selections... Oh. Maybe we never do that... ?

morphic.js
```diff
@@ -8047,22 +8131,45 @@ StringMorph.prototype.deleteSelection = function () {
...
+    if (this.world().currentKey === 16) {
+        this.shiftClick(pos);
+    } else if (this.isEditable) {
         this.clearSelection();
     } else {
         this.escalateEvent('mouseDownLeft', pos);
     }
 };
 
+StringMorph.prototype.shiftClick = function (pos) {
```
jmoenig: very nice! I love this

morphic.js
```diff
@@ -8079,18 +8186,79 @@ StringMorph.prototype.mouseClickLeft = function (pos) {
...
+    // selects the word at pos
+    // if there is no word, we select whatever is between
+    // the previous and next words
+    var slot = this.slotAt(pos);
+
+    if (this.isEditable) {
+        this.edit();
+
+        if (slot === this.text.length) {
+            slot --;
```
jmoenig: ... ```slot -= 1; ```

morphic.js
```diff
@@ -8079,18 +8186,79 @@ StringMorph.prototype.mouseClickLeft = function (pos) {
...
+StringMorph.prototype.selectBetweenWordsAt = function (slot) {
+    var cursor = this.root().cursor;
+
+    cursor.gotoSlot(this.nextWordFrom(this.previousWordFrom(slot)));
+    this.startMark = cursor.slot;
+    this.endMark = cursor.slot;
+
+    while (this.endMark < this.text.length
+            && !isWordChar(this.text[this.endMark])) {
+        this.endMark ++;
```
jmoenig: ... and another shift. I'm sorry, but it's not a big change, right... (I can do it myself, if you prefer)

blocks.js
```diff
@@ -7529,7 +7529,6 @@ InputSlotMorph.prototype.mouseDownLeft = function (pos) {
         this.escalateEvent('mouseDownLeft', pos);
     } else {
         this.contents().edit();
-        this.contents().selectAll();
```
bromagosa: Hi Jens!

Sure, the issue was that this broke the shitf+click selection feature. You can just try to re-add this line and then play with shift click in, for example, a JS block.

morphic.js
bromagosa: Yep, I usually prefer to give boolean values to vars with boolean names though.

morphic.js
bromagosa: My bad! But actually the one needs to stay is the one in `ClipboardHandler`, we can remove this one.

morphic.js
bromagosa: Exactly, and that was breaking selection behavior when you started selecting from the beginning of any text. You can test it out in the [live morphic](http://snap.berkeley.edu/snapsource/morphic.html) page. Just go to the beginning of any text and press shift + →.

morphic.js
bromagosa: haha, yeah no problem, let me change this. Too many languages in my head lately...

morphic.js
bromagosa: You can never select more than one thing at the same time, as the cursor and selection both die whenever a `StringMorph` or a `TextMorph` loses its focus.

Additionally, placing the cursor at the end of the selection here makes the whole funny selection plus arrow key behavior much easier to deal with, as there's only one special case left to take care..

morphic.js
bromagosa: no worries, I'm onto it. Sorry! :)

morphic.js
jmoenig: Hmm... See, I never get these things right. Everything I design ends up wanting to be accessed in (at least) both directions. Maybe we want to add a reference to the cursor to each editable element, so we don't have to do all this crawling. It might make some other things easier, as well. OTOH bidirectional references have a tendency to go out of synch and currupt. Need to think about this...

PR #1464, opened 2016-10-19
Added unique sprite name check on import. Fixes #1463

### LINKED ISSUE #1463: Importing sprite can cause name collisions
When naming a sprite, `ide.newSpriteName` guarantees the name will be unique. However, if we are importing a sprite with the same name as an existing sprite, this check is not performed and can result in multiple sprites with the same name.

brollb:


jmoenig:
another greath catch. Thanks, Brian!

brollb:
no problem!

Issue #1469, opened 2016-10-22
[feature] Add context with function


dekrain:
Generally, you can push context with block, input or name of function.
So, If you want to run custom function, you must do this:
```javascript
var f = new CustomBlockDefinition();
var func = SpriteMorph.prototype.blockForSelector('reportJSFunction');
var caller = SpriteMorph.prototype.blockForSelector('evaluate');
var rep = SpriteMorph.prototype.blockForSelector('doReport');
func.inputs()[0].setContents(['proc']);
func.inputs()[1].setContents(' /* Code here... */ ');
caller.replaceInput(caller.inputs()[0], func);
rep.replaceInput(rep.inputs()[0], caller);
f.body = rep.fullCopy();
process.pushContext(f.blockInstance());
```
but it be fine if support that:
```javascript
process.pushContext(function(proc){ /* A code */ });
```

PR #1470, opened 2016-10-22
Implement context functions

dekrain:
Implementation of #1469

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -683,12 +683,12 @@ Process.prototype.evaluateBlock = function (block, argCount) {
 Process.prototype.evaluateFunction = function (exp) {
     var yields = false, result;
     if (this.isCatchingErrors) {
-            try {
-                result = exp.apply(this.context.receiver, [this, function(){ yields = true; }].concat(this.context.inputs));
-            } catch (error) {
-                this.handleError(error, exp);
-                return;
-            }
+        try {
+            result = exp.apply(this.context.receiver, [this, function(){ yields = true; }].concat(this.context.inputs));
+        } catch (error) {
+            this.handleError(error, exp);
+            return;
+        }
     } else {
         result = exp.apply(this.context.receiver, [this, function(){ yields = true; }].concat(this.context.inputs));
     }
--- a/threads.js
+++ b/threads.js
@@ -633,6 +633,9 @@ Process.prototype.evaluateContext = function () {
     if (exp instanceof BlockMorph) {
         return this.evaluateBlock(exp, exp.inputs().length);
     }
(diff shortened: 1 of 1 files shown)


jmoenig:
I have no idea what this is about. Can we communicate a little bit more verbally? Like I'd love to know more about what it is you're doing, and not just guess from reading code. Otherwise I'm going to have to close all these cryptic contributions, because, frankly, I don't know what they are about.

dekrain:
It implement a new type for expression for context. If it's function, it normally evaluates it as normal block.

liam4:
I'm curious why you can't (or didn't) add `start` as an input right away?

```js
// Push a context. Stuff can be accessed using `proc`, `yield`
// can be used to cause Snap! to yield (e.g. wait-0-secs), and
// all arguments after those first two are inputs added to the
// procedure.
process.pushContext(function(proc, doYield, inptTime, inptStart) {
  if (!(Date.now() - start > time)) {
    doYield()
  }
})

process.context.addInput(1000)
process.context.addInput(Date.now())
```

Also, why would you really need `addInput` at all? Couldn't you do this?

```js
var time = 1000
var start = Date.now()

process.pushContext(function(proc, doYield) {
  if (!(Date.now() - start > time)) {
    doYield()
  }
})
```

This reminds me a bit of scratch-vm.. though I don't quite know how stuff like this works there, either :) -- just answers to those questions should help though, thanks.

Aside from those questions, I guess this is cool.

I don't quite understand the example in #1469, though..

dekrain:
Don't inspect this, this is only example

liam4:
Inspect? Like, look at how it works..?

Isn't that the point of an example?

dekrain:
Ok, I mean: focus mostly on my feature. This is only example

liam4:
But I don't know how your feature works or what it does, since you didn't describe it in the original post - so either the example should explain it, or you should explain it with words :)

(I'd go for well-commented examples explaining feature requests like this.)

dekrain:
It will be fine if expression of context can be function

brianharvey:
Wait, are you inventing continuations?  We already have those...

dekrain:
No, I mean non-blocks expressions (functions)

cycomachead:
This isn't something for Snap<em>!</em> _users_, right? 

In terms of #1469 that's something only people hacking on Snap<em>!</em> would do - so I don't think this belongs in Snap<em>!</em>.

dekrain:
Nope. It's only for manually executing. (`Process.prototype.pushContext` etc.)

PR #1476, opened 2016-10-25
Renamed 'Cancel' to 'Close'

brollb:
This is a pretty minor suggestion about the labeling of the "Cancel" button when importing costumes.

"Cancel" seems to imply that the importing of the selected costumes will be canceled. Since the costume imports cannot be canceled, I thought "Close" was more clear as to the behavior of the button

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -2892,7 +2892,7 @@ IDE_Morph.prototype.importMedia = function (mediaType) {
     dialog.createLabel();
     dialog.addBody(frame);
     dialog.addButton('ok', 'Import');
-    dialog.addButton('cancel', 'Cancel');
+    dialog.addButton('cancel', 'Close');
 
     dialog.ok = function () {
         if (selectedIcon) {


liam4:
+1, especially because the dialog doesn't disappear after importing a costume (which is good, IMO).

brianharvey:
Thank you!

jmoenig:
We currently don't have ```'Close'``` in the translations dictionary, so I'm reluctant to pull this one, because it'll make Snap less accessible to kids in 37 languages...

brianharvey:
Jens, we have to change it to something.  Kids think that if they push that button they won't get the new costume(s) they asked for!  Better a fixable problem in 37 languages than a bad problem in 38 languages!

jmoenig:
This thing was a pull-request that I gave in to integrating into Snap. The right thing is to enable multi-select on the costumes (but please don't let a kid go off an implement it!). In the meantime kids can double-click on costumes to import them right-away, and those costumes are immediately shown because the IDE switches to the costumes tab, so there is no mystery whatsoever. The double-click feature is a "power-user" gimmick, beginners can just select one costume and either press OK or Cancel. I don't see any urgency to change this.

brianharvey:
It's been driving me crazy since that feature came in.  Would it be an okay compromise to close that dialog if you double-click?

bromagosa:
Just a quick and very ugly proof of concept, but what about adding a close button to all DialogBoxMorphs:

image

This doesn't need translation, and every person on the planet knows what a red button with a cross does.

brianharvey:
> every person on the planet knows what a red button with a cross does.

Umm, actually, I think it's only Windows users who see red buttons.  The rest of us sometimes see unadorned X shapes on the regular background.  And sometimes we see three dots in the top left corner, the leftmost of which shows an X if you hover over it.

In any case, Bernat, I think your solution is the tail wagging the dog.  No reason to uglify every dialog just because of a translation problem in one of them.

I reiterate my suggestion that double-clicking a costume should close the window.  Then the word "Cancel" will only mean what actually happens if you single-click and hit Cancel.

cycomachead:
> I reiterate my suggestion that double-clicking a costume should close the window. Then the word "Cancel" will only mean what actually happens if you single-click and hit Cancel.

Yes. "Cancel" means "undo this action" or "don't do what I just tried to do".

Also, "Cancel" is a 1 word translation - we could pretty easily knock out a few of those 37 languages ourselves. (Not by using a translator, but by changing our iPhones in 37 different languages and finding a Cancel button.)

PR #1478, opened 2016-10-27
GitHub integration

wcyuan:
This is more like a request for comment than a pull request.  This change adds Github as a new source for loading and saving snap projects.  It doesn't really make sense as is, since it's not suitable for beginners, but maybe it could become a hidden advanced feature.  Currently, this change doesn't really tap into the power of git (no way to see or access past revisions, or branches or anything).  It also only handles regular passwords (not two factor auth).  With this level of auth, the file size limit is only 1 MB.  

thanks,
Conan

### DIFF
--- a/github.js
+++ b/github.js
@@ -30,7 +30,7 @@
 /*global modules, IDE_Morph, SnapSerializer, hex_sha512, alert, nop,
 localize*/
 
-modules.cloud = '2016-October-16';
+modules.github = '2016-October-16';
 
 // Global stuff
 
@@ -43,6 +43,8 @@ function Github() {
     this.baseUrl = "https://api.github.com";
 }
 
+// Github API
+
 Github.prototype.getProjectList = function (
     callBack,
     errorCall,
@@ -52,7 +54,7 @@ Github.prototype.getProjectList = function (
     var myself = this;
     username = username || "wcyuan";
     repo = repo || "snap-projects";
-    var url = "https://api.github.com/repos/"
+    var url = myself.baseUrl + "/repos/"
         + encodeURIComponent(username)
         + "/"
         + encodeURIComponent(repo)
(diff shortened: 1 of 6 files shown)


dekrain:
WOW !!

jmoenig:
Interesting! Another idea would be to start with generic local git integration, as other IDEs do...

wcyuan:
> Another idea would be to start with generic local git integration, 

Yes, that would be better in many ways, since it would work with any git setup not just Github.  I like going directly to the cloud for students who may be on shared computers, but that might not matter since git is already not for beginners.  

By the way, if you want to try it out, you can use wcyuan.github.io/snap-github

brianharvey:
Imho our users would be much more interested in Google Drive integration, and then in second place Dropbox.

dekrain:
For JSON messages try to use:
```javascript
var json = JSON.parse(xhr.responseText);
function jsonblock(blocks){
var block = new ReporterBlockMorph();
block.setSpec('JSON { %c }');
block.category = 'control';
(new ScriptsMorph()).add(block);
var curr;
if (blocks.length >= 1) {
 block.replaceInput(block.inputs()[0], (curr = blocks[0]));
 blocks.slice(1).forEach(function(blck){
  curr.nextBlock(blck));
  curr = blck;
 });
}
block.drawNew();
return block;
}
function fieldblock(name, value){
var block = new CommandBlockMorph();
block.setSpec('%s : %s');
block.category = 'operators';
(new ScriptsMorph()).add(block);
block.inputs()[0].setContents(name);
if (value && value instanceof ReporterBlockMorph) {
block.replaceInput(block.inputs()[1], value);
} else {
block.inputs()[1].setContents(value);
}
block.drawNew();
return block;
}
function json_to_blocks(data) {
var blcks = [];
for (var k in data) {
 var v = data[v];
 blcks.push(
  fieldblock(
   k, (v && typeof v === 'object') ? json_to_blocks(v) : v
  );
 );
}
return jsonblock(blcks);
}

// Show the block
```
For example:
```
(JSON {
[message] : [YAY!!]
[code]: [200]
[_meta]: (JSON {
[url]: [http://...]
})
})
```

Issue #1481, opened 2016-10-28
[feature] Implement multi-line inputs


dekrain:
Image:
![multi line](https://cloud.githubusercontent.com/assets/19158291/19810707/cd43c3f4-9d2e-11e6-9645-fb7461fd89ae.png)
Fragment of code:

``` javascript
InputSlotDialogMorph.prototype.addSlotsMenu = function () {
    var myself = this; var mlt = false;

    this.slots.userMenu = function () {
        if (contains(['%s', '%n', '%txt', '%anyUE'], myself.fragment.type)) {
            var menu = new MenuMorph(myself),
                on = '\u2611 ',
                off = '\u2610 ';
            menu.addItem('options...', 'editSlotOptions');
            menu.addItem(
                (myself.fragment.isReadOnly ? on : off) +
                    localize('read-only'),
                function () {myself.fragment.isReadOnly =
                         !myself.fragment.isReadOnly;
                         }
            );
            if (myself.fragment.type !== '%n') {
                menu.addItem((mlt ? on : off) + localize('multi-line'), function(){mlt=!mlt;});
            }
            return menu;
        }
        return Morph.prototype.userMenu.call(myself);
    };
};
```

jmoenig:
yep, the only reason I've not yet added this feature is that mulit-line text currently doesn't support the brown-dot placeholder notation for spaces.

dekrain:
I can make this, it's a moment...

jmoenig:
happy to announce that Snap5 now has an option for built-in multi-line input slot support, albeit without showing spaces as brownish dots. Thank you! Closing...

PR #1482, opened 2016-10-28
Add multiline option to input spec fragment

dekrain:
Implements #1481


dekrain:
Hellooo?

dekrain:
@jmoenig ?

jmoenig:
I've considered even making everything multi-line in Snap, but currently TextMorphs don't support the brownish dot for white-space, which Snap needs to distinguish empty input slots (and thus implicit parameters) from whitespace. Closing.

PR #1491, opened 2016-11-01
Remove the manifest file

cycomachead:
As of today (Nov 1, 2016) appcache is removed from the HTML5 spec.
Currently, the appcache isnt being used in Snap! and so there arent and changes in removing it.
We should look at other options for offline access for which there are many.


cycomachead:
Tag #900

jmoenig:
Yep, I agree. Thanks, Michael. You know, my favorite solution would be an installable app just like @bromagosa 's Snap4Arduino. Wouldn't that solve it all?

bromagosa:
That'd solve it all, yes :)

Michael has an Electron-based Snap<i>!</i> version. We could build on that one.

cycomachead:
Yeah, everything seems to work fine so far - I just need to figure out the autoupdater stuff. Which shouldn't be terribly difficult. :) I haven't extensively tested everything, but the basics work and the stuff I would expect to break (like libraries, and the image security exceptions and saving files) all work fine.

Then of course, there's all the little bits to polish it and make things look like a real app, but that'll come over time I guess.

https://github.com/cycomachead/snap-app

jmoenig:
Awesome, this is really exciting Michael!

brianharvey:
I pulled this since it doesn't change any code anywhere.  Why did they remove the feature?  Is there an officially approved other way to do it?

cycomachead:
> Why did they remove the feature? Is there an officially approved other way to do it?

I think it's because the current version was a bit difficult to work with and never got traction. (It didn't scale super well to larger apps).

There's going to be a new version sometime: https://w3c.github.io/manifest/

Current browsers would still support the old version for a while, but given that we never got it fully working and there are other issues (caching), it seems worth persuing other options.

Desktop/native app wrappers (like the Snap!-app repo) are nice because they "de-sandbox" the app and we could take advantage of native features (like "open in Snap!" controls for XML files, or better saving and opening from the desktop.) We do lose the advantage of having sites automatically work offline but there was some debate about how useful this was. (You would never know whether a site worked offline until you tried it.)

bromagosa:
No only that, but you also get rid of all CORS issues in native app wrappers.

cycomachead:
> Not only that, but you also get rid of all CORS issues in native app wrappers.

How could I forget that! Yes, that's a nice advantage!

brianharvey:
Okay, I'm in.  Do the apps made by Snapp<i>!</i> update themselves when on the net?  That's the big thing we need.  And then a menu item in the Snap<i>!</i>-logo menu "Download standalone app."

cycomachead:
You mean, does the desktop app auto-update?

Not yet, but that's what's holding me back from releasing it. Other's have built out the infrastructe so it's totally ddoable.

PR #1503, opened 2016-11-08
Control+Z mapped to undrop

bromagosa:


cycomachead:
LGTM, this will definitely be useful.

Now we just need to find a good way to surface keyboard shortcuts in menu actions. :)

liam4:
Tadaaa!

:tada:

jmoenig:
Not *quite* sure whether it's right to map ```undrop``` to the ```undo``` keyboard shortcut just now, or whether to implement the full ```undo``` feature first. Let me think about this some more...

cycomachead:
That's fair. I guess it depends if I undrop would be a component of a general undo feature?  it seems like most people think of undrop as "partial undo" so if things change in the future it would just become a better version of Undo mapped to the same keys. 

-- 
Michael Ball
From my iPhone
michaelball.co

jmoenig:
yes, "undrop" is only a small part of a general "undo". For example, I want "undo" not to place the last dropped block back in to the "hand", but to slide it back to from where it was picked up. Also, there have to be - really unlimited - levels to undo, at least per session.

jmoenig:
Ah, I see that this has been superseded by my last commits...

PR #1524, opened 2016-11-23
Corrections for the Danish translation

hjek:
I should add some more corrections, but particularly _if ... else_  → _hvis ... sonst_ sounds more German than Danish.

### DIFF
--- a/lang-dk.js
+++ b/lang-dk.js
@@ -2,7 +2,7 @@
 
     lang-dk.js
 
-    German translation for SNAP!
+    Danish translation for SNAP!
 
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks! I'll pull this for the next release, which might take a few weeks (maybe as long as Christmas, so please don't despair, I've got this ^^)

hjek:
Nice! There are still a few bits of German lingering around the Danish translation that I can probably fix in time for that release then.

dekrain:
Shouldn't it have [i18n] label?

PR #1535, opened 2016-11-29
Russian, while vs until

rewlad:
"пока" means "while";
there's no proper short word for "until";
so we can either use "до" ~ "before" or "пока не" ~ "while not";


jmoenig:
thank you very much! I'll pull this and integrate it for the next upcoming minor release, which, however, might take even up to a few weeks. Stay tuned, I'm not forgetting this PR (that's why I labelled it green just now). :-)

PR #1546, opened 2016-12-05
Costumes tab name to Costumes/Backgrounds

jguille2:
A suggestion.

Adapt 'costumes' tab name to real name (Costumes for sprites and Backgrounds for stage).

Thanks,

Joan

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -1280,7 +1280,9 @@ IDE_Morph.prototype.createSpriteBar = function () {
         tabColors,
         null, // target
         function () {tabBar.tabTo('costumes'); },
-        localize('Costumes'), // label
+        localize(this.currentSprite instanceof SpriteMorph ?
+			'Costumes' : 'Backgrounds'
+			),
         function () {  // query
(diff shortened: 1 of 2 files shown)


jmoenig:
Good point, thanks for the pull-request, @jguille2 !

jguille2:
This second commit adapts "rename costume" to the context (costume for sprites, background for stage).

I would add these two new strings ('Backgrounds' and 'rename background') to translation files... but I think we need to arrange the translation system (actually not synchronized). I'll open another issue... and maybe I can help doing some work in this theme.

jguille2:
This third commit (the last :) ) to adapt stage messages to turn pen trails into costumes/backgrounds.

Shorter strings messages would be better to reduce the number of translations but of course I've adapted to existing messages. In any case, two more strings to put into translations files.
We will talk about the translation system in #1551, and I'll update translations files with these 4 strings and with the rest of the pending work.

Thanks!!

jmoenig:
Excellent, thanks @jguille2 !

Issue #1553, opened 2016-12-08
Deleting custom block definition creates extra input box, causing other bugs


snorthway:
Related in some way to #1372 but I didn't see anything about an extra input box in that discussion. Apologies if I missed something there that makes this a duplicate.

Steps:

1. Make a normal script
<img width="120" alt="screen shot 2016-12-08 at 11 05 50 am" src="https://cloud.githubusercontent.com/assets/2825499/21017323/ff5cdad8-bd36-11e6-8038-e0d839bed529.png">

2. Create a custom block (note that this works regardless of whether there's any code in the custom block)
<img width="573" alt="screen shot 2016-12-08 at 11 06 10 am" src="https://cloud.githubusercontent.com/assets/2825499/21017357/16f45450-bd37-11e6-8ac0-a988ec2a0a9f.png">
<img width="615" alt="screen shot 2016-12-08 at 11 06 19 am" src="https://cloud.githubusercontent.com/assets/2825499/21017359/1954d7ba-bd37-11e6-8a6d-ee3dc9748996.png">

3. Put the custom block at the end of your script
<img width="135" alt="screen shot 2016-12-08 at 11 06 28 am" src="https://cloud.githubusercontent.com/assets/2825499/21017371/24de4e04-bd37-11e6-93d4-86b763f5d416.png">

4. Right click the block you just put in and select "delete block definition"
<img width="550" alt="screen shot 2016-12-08 at 11 06 38 am" src="https://cloud.githubusercontent.com/assets/2825499/21017385/30776340-bd37-11e6-89f0-38d068bd2d74.png">

5. An extra input box appears at the end of the last block.
<img width="123" alt="screen shot 2016-12-08 at 11 06 42 am" src="https://cloud.githubusercontent.com/assets/2825499/21017400/35ac3430-bd37-11e6-8c0d-c7d76947ca61.png">
<img width="145" alt="screen shot 2016-12-08 at 11 06 53 am" src="https://cloud.githubusercontent.com/assets/2825499/21017426/4a98cc1e-bd37-11e6-9729-3d5badabbbe8.png">

6. BONUS: if you put a block inside the extra input box, you ~rip a hole in the space-time continuum~ can see a few other bugs.

For example, I accidentally dragged a "pick random" into the box. When trying to remove it, it just made this huge shadow.
[...]

PR #1554, opened 2016-12-08
Fix extra inputs appearing when custtom blocks are deleted

jguille2:
Fixing #1553 

We need to amend [this change](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/commit/ff5f9a7601d1b37868ab6ccc5b79fd9b78362411#diff-5f5bb19933ecde89bb86b19173cdef9eL2531) made in October to improve Visible stepping feature.
It causes the appearance of extra inputs after custom block definitions deleting.

I have rewritten the reporterBlock condition.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -2546,9 +2546,10 @@ BlockMorph.prototype.deleteBlock = function () {
             }
         });
     }
-    if ((this.parent instanceof BlockMorph)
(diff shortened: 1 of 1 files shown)


jmoenig:
Thank you, @jguille2 ! I appreciate your quick solution to this issue!

PR #1558, opened 2016-12-09
Localization for 'undrop' and 'redrop' options

jguille2:
Hi,
Working in catalan translation I've seen new 'undrop' and 'redrop' options are not translatable. File lang-de.js are updated with these two strings, but code needs the explicit 'localize()' function.
Thanks!

Joan


jmoenig:
good catch, thanks @jguille2 !

Issue #1551, opened 2016-12-07
Updating translation strings in lang files from code


jguille2:
I see lang-de.js is not synchronized with code strings.

I can use lang-pt.js like a template (with 248 extra strings) to achieve a better translation... but I know it is not the right way.

I can try working with this to bring something useful. First thoughts:

  - We need a **lang.js** as a template for translations. Only with the English strings. Easier to update from any developer, adding and updating code strings.
  - What would be the correct strings order?  Themes or categories seems best... but we use the same strings in different places. We could order them **alphabetically**.
  - We can add **comments** to string lines. We can annotate the file (or files) where they appear, or we can annotate the category (tabs, editor, menu...)

In any case, I offer to work on this topic and follow your indications.

Joan

DyslexicAwe:
@jguille2 Wow, that would be an awesome Seasons Holiday gift for all translators. Until now, changing an existing lang-_LL_.js was not an easy task and sometimes it just didn't work and I had no clue why it didn't work.

jguille2:
First of all, I try to develop a proposal and wait for your feedback... then I can do the job according to your instructions.

Excuse my non-first-class English language :(. Luckily, this is an asynchronous issue (I can use more time to do my best)... and I trust your patience to  fix this (my) poor bug :))

## Goals
Two main goals for our translation system:

  - Make it easier to build and update translations.
    - Build a simple system.
    - Keep translation files synchronized with code.
    - Allow and encorage the maximum number of languages (with full or partial translations).

  - Allow developers to change strings in code when refining or improving features... without being tied to an excessively rigid structure.

## Proposals

  1. Build a **lang-en.js** as the master template. English to English translation

    - All developers and helpers could directly update it, with the code changes.
    - Easier to spread changes. Partial translations will be copies from this template (with all the strings) that can be easily updated (only editing text).
    - English language could be refined without code changes (English code to English lang). This will not be the standard way, but an easy possibility. Some times, our free translations are more refined than the English one (chained to the system).

  2. lang-en.js (and the other lang files) can have all the **strings in alphabetical order**.

    - I like a thematic distribution (menu, blocks...) but strings can live in different scenarios.
    - Easier to developers to use existing strings, avoid collisions...
    - Maybe one line per string is better (although they'll be longer)
    - We can comment lines to add info. Source file, function... or only a counter of uses (to help developers to delete lang strigs after code modifications).

  3. Make an **updater script** to run it locally.

[...]

PR #1561, opened 2016-12-13
Catalan language update

jguille2:
Three commits updating lang-ca file (2016-12-12)
1. Update with lang-de
2. Add lang-pt strings (245 extra strings)
3. Add more 26 strings

We keep talking in #1551 about synchronization strategies. The 'extra' 26 strings are already (some of them) in different languages... We will continue there...

Thanks,

Joan

### DIFF
--- a/lang-ca.js
+++ b/lang-ca.js
@@ -289,7 +289,7 @@ SnapTranslator.dict.ca = {
 
     // motion:
     'Stage selected:\nno motion primitives':
-        'Escenari seleccionat:\nno hi ha primitives de moviment\n'
+        'Esteu a l\'Escenari:\nno hi ha primitives de moviment\n'
             + 'disponibles',
 
     'move %n steps':
@@ -675,11 +675,11 @@ SnapTranslator.dict.ca = {
     'Import...':
         'Importa...',
     'file menu import hint':
-        'carrega una llibreria de projecte\no de blocs exportada, un vestit\no un so',
+        'Importa projectes, blocs,\nimatges o sons',
 
 
     'Export project as plain text...':
-        'Exporta el projecte en text pla...',
+        'Exporta el projecte...',
     'Export project...':
         'Exporta el projecte...',
     'show project data as XML\nin a new browser window':
@@ -687,11 +687,11 @@ SnapTranslator.dict.ca = {
     'Export blocks...':
         'Exporta els blocs...',
     'show global custom block definitions as XML\nin a new browser window':
-        'mostra les definicions de blocs personalitzats\nen format XML en una altra finestra del\nnavegador',
+        'exporta els blocs personalitzats que triis\nen un arxiu en format XML',
     'Unused blocks...':
         'Blocs no utilitzats...',
     'find unused global custom blocks\nand remove their definitions':
-        'busca blocs personalitzats globals\nno utilitzats i esborra\'ls',
+        'cerca els blocs personalitzats globals\nque no s\'utilitzan per poder esborrar-los',
     'Remove unused blocks':
         'Esborra blocs no utilitzats',
     'there are currently no unused\nglobal custom blocks in this project':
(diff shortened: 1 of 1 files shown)


lang-ca.js
```diff
@@ -924,9 +930,11 @@ SnapTranslator.dict.ca = {
     'add comment':
         'afegeix un comentari',
     'undrop':
-        'recupera bloc',
+        'desfés',
```
bromagosa: @jguille2 "undrop" does not work in the same way as "undo". "undrop" just recovers the latest dropped block, hence "recupera bloc", whereas "undo" undoes whatever action was perform last, hence "desfés". It is very important to keep the distinction in the translation, they are separate actions with different outcomes.

lang-ca.js
```diff
@@ -924,9 +930,11 @@ SnapTranslator.dict.ca = {
...
     'undo the last\nblock drop\nin this pane':
-        'recupera l\'últim bloc\nque s\'hagi llençat',
+        'desfés l\'últim moviment\nde blocs',
```
bromagosa: same here, the original translation conveys better what the action does. "Undrop" does not "undo" the latest block movement, that's what "undo" does. In any case, maybe it should have been phrased "recupera l\'últim bloc\nque s\'hagi deixat anar".

lang-ca.js
```diff
@@ -924,9 +930,11 @@ SnapTranslator.dict.ca = {
...
+    'redrop':
+        'refés',
```
bromagosa: And here. "Redrop" is not the same as "Redo".

lang-ca.js
jmoenig: @bromagosa the distinction between "undrop" and "undo" has been blurred in the upcoming release. Basically "undrop" is now *almost* a full "undo", with the only limitation that it only works for blocks and scripts. I've been keeping the name "undrop" only because @brollb is working on a much, much better and way more general "undo" mechanism that truly deserves this name. For translation purposes I actually don't mind going straight to "undo" and "redo".

lang-ca.js
jmoenig: Yeah, same here. Now "undrop" actually *does* "undo" the latest block movement, even in an animation fashion and for unlimited levels :-)

lang-ca.js
jmoenig: ... and "redrop" also moves the latest block "undrop" back again, and also undeletes any deleted block, regardless of whether it was dragged and dropped on the palette or deleted via the context menu. In essence, the upcoming release already has a better "real" undo/redo mechanism than most existing blocks environments out there.

bromagosa: Oh I see! I stand corrected then :)

I thought you meant to keep undrop/redrop along with @brollb's undo/redo, but if there's only to be one single mechanism then @jguille2's translation makes perfect sense.

jmoenig: Yeah, we're going to exchange my new "undrop" mechanism with @brollb 's "real undo" once it's ready for prime time. But since @brollb 's mechanism is much more refined I wanted to make it clear that my hacky kludgy one isn't yet the real thing by choosing a different name :-)

jmoenig:
@bromagosa do you otherwise agree @jguille2 's suggested changes? In this case I'd go ahead and pull this for the upcoming Xmas release. Thanks!

bromagosa:
@jmoenig yep, of course! :)

jmoenig:
Thanks!

lang-ca.js
jguille2: Hi, @bromagosa! Adding new 'redrop', I changed 'undrop' consciously. @jmoenig has explained better than I can do in English :))
I've changed other strings with a free translation... because original English string maybe is not the right explanation now (features changing...). I prefer tot talk about this in #1551 because is more general about translations synchronization system. I'll try to make a specific proposal...

PR #1567, opened 2016-12-20
Finish the experimental math expression to blocks translator

BookOwl:
As @brianharvey asked, I made the experimental math parser follow PEMDAS and be left-associative.

Screenshots:

untitled


tjvr:
This is in conflict :( Can you try rebasing onto the latest master?

objects.js
```diff
@@ -2640,87 +2640,135 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
...
+                token = findNumber();
+                tokens.push({ type: "number", val: token});
+            } else if (/[+\-*/%]/.test(expressionString[0])) {
+                tokens.push({ type: "operator", val: expressionString[0]});
+                expressionString = expressionString.slice(1);
+            } else if (expressionString[0] === "(") {
+                tokens.push({ type: "lp", val: "("});
+                expressionString = expressionString.slice(1);
+            } else if (expressionString[0] === ")") {
+                tokens.push({ type: "rp", val: "("});
```
tjvr: Why `val (` here?

objects.js
```diff
@@ -2640,87 +2640,135 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
...
+                    while (tokens[0].type !== "rp") {
+                        args.push(createASTFromTokens(tokens));
+                        if (tokens[0].type === "rp") {
+                            break;
+                        }
+                        if (tokens[0].type !== "comma") {
+                            throw "No comma between arguments";
+                        }
+                    }
+                    tokens.splice(0, 1);
```
tjvr: Why not `tokens.shift()` instead? :)

tjvr:
This looks plausible to me, FWIW :)

objects.js
BookOwl: Oops, typo. It doesn't really matter, but I'll fix it.

objects.js
BookOwl: Because MDN said that you use splice to remove elements. My JS is kind of rusty.

BookOwl:
Yeah, conflicts are gone!

objects.js
BookOwl: I know, but it also returns it. Since I'm not doing anything with the result, I thought that splice was better. I'll change it if you think that shift is better though.

BookOwl: I changed it just for you, @tjvr

brianharvey:
Thanks, Matthew, for the programming, and Tim, for the code review!

I don't quite understand the second of three screenshots in the first post on this page, though.  How did 5*7/x turn into 5*7*x?  Or was that the teeny typo?

BookOwl:
Nope, that's a different bug. Thanks for pointing it out.

BookOwl:
Fixed

objects.js
```diff
@@ -2640,87 +2640,137 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
...
+            var next = "";
+            while (expressionString.length > 0 && /[A-Za-z_0-9]/.test(expressionString[0])) {
+                next += expressionString[0];
+                expressionString = expressionString.slice(1);
+            }
+            return next;
+        }
+        function findNumber() {
+            var next = "";
+            while (expressionString.length > 0 && /\d|\./.test(expressionString[0])) {
```
cycomachead: Our strings are probably not long enough to make a huge difference, but `.slice()` allocates a new string, so it seems better to just use an index.

Also why check `.length > 0` when `expressionString` is false if empty.

objects.js
```diff
@@ -2640,87 +2640,137 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
...
+                next += expressionString[0];
+                expressionString = expressionString.slice(1);
+            }
+            return parseFloat(next);
+        }
+        while (expressionString.length > 0) {
+            expressionString = expressionString.trim();
+            if (/[A-Za-z_]/.test(expressionString[0])) {
```
cycomachead: I feel like all regex's should be defined and given a semantic name, otherwise it's difficult to read why they are used.

objects.js
```diff
@@ -2640,87 +2640,137 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
...
-                inputs[operator ? 1 : 0] += ch;
+            } else {
+                throw "Invalid expression";
             }
         }
-        if (operator) {
-            return [operator, already, format(inputs[1])];
+        function precedence1(tokens) {
+            var exp = precedence0(tokens);
+            if (tokens.length === 0) {
```
cycomachead: Typically this would just by `!tokens`

jmoenig:
Thanks! Let me know when you feel it's ready for heavy testing.

Can you make it support negative numbers? Currently your parser turns negative numbers into separate nodes subtracting the number from zero. While this is kinda correct mathematically it makes the resulting formula much more complex than it needs to be.
Thanks!

objects.js
```diff
@@ -2640,87 +2640,135 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
     // highly experimental Christmas Easter Egg 2016 :-)
     var ast;
 
-    function parseInfix(expression, operator, already) {
-        // very basic binary infix parser for arithmetic expressions
```
cycomachead: This comment should be retained, though in an updated form, I think - including what happens when an error is thrown.

objects.js
```diff
@@ -2640,87 +2640,135 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
...
-                        return expr;
-                    }
-                    break;
-                }
-                expr += ch;
+    function tokenize(expressionString) {
+        var tokens = [];
+        var token;
+        expressionString = expressionString.trim();
+        function findIdent() {
```
cycomachead: Maybe I'm daft, but I am confused by the name `Indent`. What is meant by Indent? It seems to be anything that's text?

objects.js
```diff
@@ -2779,7 +2829,7 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
 
     if (expressionString > 100) {return null; }
```
cycomachead: More for @jmoenig, but I also feel like these magic numbers should be given meaningful names. :)

Also I'm pretty sure this should be `expressionString.length`?

jmoenig: Oh sure, right!

jmoenig: I just fixed that in the repo

jmoenig: Thanks, Michael!

jmoenig:
Hey, @BookOwl , thanks for pointing out the glitch in my code in the ```round``` case! I just pushed a fix which now causes conflicts :-(

How about if you simply exchange my whole ```reporterize``` function with your own one, that way we can keep both in the source for a while without running into conflicts?

jmoenig:
Another question to consider is whether we should add a ```raisedTo``` operator/reporter primitive to Snap...

objects.js
tjvr: Personally I find that harder to follow — then I have to look elsewhere to work out what it's doing! But that's just me :-)

objects.js
tjvr: Ident is short for identifier. You've added an N :-)

objects.js
tjvr: Nope, only if tokens were a list. The empty list is not falsey in JS! This isn't Python... :-)

objects.js
BookOwl: Oops, another typo. I'll fix it.

objects.js
BookOwl: I am using an index...
I'll remove the .length check

objects.js
BookOwl: I'll change this since Jens agrees with @cycomachead

BookOwl:
@jmoenig, I can't make `-x` not be `0 - x` in the general case because Snap doesn't have a negate block, so `-(4 * 7)` would have to be turned into `(0 - (4 * 7))`, and I just had numbers do the same thing because I didn't have to do any extra work.

objects.js
BookOwl: Which function should I put the comment in?

jmoenig:
right, but you can make negative numbers, such as -42, and just use them as inputs. No need to turn -42 into (0 - 42). Make sense?

BookOwl:
Yes

BookOwl:
Should I add the boolean operators in my version of the parser/block generator?

BookOwl:
I made the negative numbers parse correctly.

BookOwl:
I resolved the conflicts be just telling git that my version is what should be kept.

objects.js
liam4: Typically this would just be !tokens.length from what I've seen in other code bases. But I prefer `.length === 0`.

objects.js
```diff
@@ -2856,7 +2838,7 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
         return block;
     }
 
-    if (expressionString.length > 100) {return null; }
+    if (expressionString > 100) {return null; }
```
dekrain: You removed fix of bug.

BookOwl: Oops, thanks for pointing that out

brianharvey:
@jmoenig #raisedto: All these years we've been resisting adding as primitives things that can usefully (as a learning experience) be written in Snap<i>!</i> itself.  Perhaps we should revisit that policy, but not for this reason -- it'd be the tail wagging the dog.

In this case we probably want a full math library with ^ as just one piece, optionally loadable.

brianharvey:
@BookOwl #Booleans: Bearing in mind that the original motivation for this whole feature was pasting in formulas from published work, if it's a math paper rather than a computer program, Boolean symbols mean something more like assert(...) than like a computation.  I suppose it'd be possible for ∧ and ∨ to have one meaning while & and | have a different meaning.  But I still think we should live with arithmetic for a while before we start planning extensions.

BookOwl:
@brianharvey, I'm asking because @jmoenig has already added it to his version of the parser.

objects.js
cycomachead: I don't know if it matters much, but probably around createAST since that's doing essentially the same work.

objects.js
cycomachead: That may have been autocorrect. :) In any case, then it should be `findIdentifer` We're not saving that many characters, and my first guess was actually "Identity" but that didn't make sense.

objects.js
cycomachead: What it _is_ doing and what it is _trying_ do are two separate things. The reason to name things is because they help convey programmer intent, not just function. :) When I look at this I question whether or not `-` can be in a variable name or whether this is a complete set of characters. I completely forgot to ask that question before....

objects.js
cycomachead: OMG Github just lost my comment. :(

Anyway... You're right! I've been not writing enough JS lately....

I was suggesting `!tokens.length` then, because that's more common in Snap!'s code too. But I don't have a strong preference...

(In Snap! .length is used as a boolean about 60-70 times, and `=== 0` only 15, according to my basic grepping.)

cycomachead (commented):
I actually do think this is quite clean, but stuffing a whole infix parser inside objects.js feels wrong to me me. I'm not trying to like stake a claim on this, but I do feel like we've passed a point of what is comprehensible in some of our files. :-)

cycomachead:
While we are on the topic of design things:
I find `5 * ` to generate `5 x 0` confusing. I will like the 0 should be unfilled until I type something there.

BookOwl:
`5 *` doesn't generate `5 x 0`, or anything else. If you want an empty input slot, you write `5 * _`

brianharvey:
Oh, well, if Jens has it, by all means.  Might as well have the relational operators too.  Of course this adds more levels of binding strength. :P

cycomachead:
> 5 * doesn't generate 5 x 0, or anything else. If you want an empty input slot, you write 5 * _

Oh, interesting. On Jens' version it does:
screen shot 2016-12-21 at 10 48 53 pm

I think this is kinda what I would expect, modulo the 0 being there.

BookOwl:
@jmoenig, do you think that the boolean and relational operators should be in here?

jmoenig:
@BookOwl thanks for the commits! What's the consensus here? Is this now ready for prime-time, or would you like to keep testing and refining it? (Hey, no rush whatsoever, this is coming along just fine). Which operators would you like to change or add?

BookOwl:
I just added pick random, so unless you want me to add the boolean and relational operators, I think that this is ready for you to test.

objects.js
```diff
@@ -2634,96 +2634,149 @@ SpriteMorph.prototype.searchBlocks = function (
...
-                    level -= 1;
-                    if (!level) {
-                        return expr;
-                    }
-                    break;
-                }
-                expr += ch;
+    function tokenize(expressionString) {
+        // very basic tokenizer for arithmetic expressions
+        var tokens = [];
```
cycomachead: These should be:
```js

var a = 1,
    b = 2,
    c = 3;
```
formatted (to be consistent with Snap<em>!</em>)

Also, I really feel like `IDENT` should just be `IDENTIFIER` since we're not really short on characters.

objects.js
```diff
@@ -2634,96 +2634,149 @@ SpriteMorph.prototype.searchBlocks = function (
...
+        var DIGITS = /\d|\./;
+        var IDENT_START = /[A-Za-z_]/;
```
cycomachead: Is this being used to match variable names? If so it seems like we should greatly expand what it accepts, though that does significant complicate things if it were to match spaces and operators, but perhaps we can ignore that for now...

cycomachead:
When I start with a negative number, there seems to be an infinite loop. 

I was trying stuff with the expression `5 * -10 + 3` if I did certain things to `-10` like `-10.3.3.3` or `-10.1xxx` sometimes the block generated would be (5 x -10) + 3 and sometimes the +3 would be left off. At some point I deleted something and then got into another infinite loop.

cycomachead:
Hmm, consider this example. I'm not sure what's right but I'm highlighting it for others to think about:
"-11.0.0.0"
Jens' version turns this in to `() - (11.0.0.0)` where that's a variable/orange blob.
This version turns it into `-11` into whatever block you'd be in, the rest of the "0.0.0" is just dropped.

Jens' version turns "5 -11-11" into 5 - 11 -11, which this version is 5 - 11. 

Ok, some of these are small issues, but it doesn't seem too hard for students to run into weird cases saying "what the heck is going on?" (in both versions). I'd be really useful to have some documentation about a couple of the edge cases with some expectations for what they do.

objects.js
BookOwl: Yes it does, but that regex is just for the start of variable names. The rest of a variable can also contain digits.

BookOwl:
The reason that "-11.0.0.0.0" gets turned into "-11" is because that is how [JS's `parseFloat` function works](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat)

objects.js
BookOwl: OK.

BookOwl:
`5*-10.3-3` having the `-3` get left off _is_ a bug.

What input is causing infinite loops?

jmoenig:
this is coming along nicely, thanks! I keep stumbling across glitches and improvements in my much simpler version, haha! This is fun...

brianharvey:
The Official Right Thing about minus signs is that `<expr>-<expr>` and `<expr> - <expr>` are dyadic subtraction, but `<expr> -<expr>` is monadic negation (and therefore has a higher binding strength than anything other than parentheses).  Of course `<operator> - <expr>` is monadic regardless of spacing.  And monadic negation of a literal nonnegative number should just be rendered as a minus sign as part of the number itself.

My vote is that a malformed numeric literal like 3.0.0 should be flagged as an error (in Snap<i>!</i>, not just in the text-to-Snap<i>!</i> translation) but that a literal that starts with a digit but includes letters, such as `2nd player`, should be a legal variable name.  But I don't think there's an Official Right Thing about that.

cycomachead:
> The reason that "-11.0.0.0.0" gets turned into "-11" is because that is how JS's parseFloat function works

IMO this is a bug. That means (if I understand correctly) that anything that looks like it starts with a number, but could be a variable gets treated as a number. But it is at least currently consistent with how Snap! handles numbers that are malformed. To me though someone trying to type '11-11-11' is probably indicating a bunch of subtractions.

As far as infinite loops: I don't remember the specific example other than just starting off with `-`. But it didn't take too much playing around to get it to happen. (It might not have been an infinite loop, but basically Safari just hung and I couldn't inspect things so that's my best guess.)

objects.js
cycomachead: (Apparently, I never hit post on this 'review'...you've already addressed some of the comments so sorry for the duplication. 

Sorry for not being clear, what I meant was an explicit index variable. (Maybe `position` or something, that is just an int.)
Then each loop would simply do `position += 1` and checking `expressionString[0]` would be `expressionString[position]`.

I don't know if it really makes a difference though, but I imagine with a long string and parsing on every change it could be noticeable especially on slower computers.

cycomachead: It might also make sense to set `expressionString[0]` to an explicit variable, but I don't know if there's much difference. (in readability or perf.)

cycomachead (commented):
Sorry this was supposed to be published a few days ago... This is the cost of having too many tabs.

jmoenig:
So, again asking the general consensus about whether this pull request is now ready to be integrated? @BookOwl , do you consider your work to be ready for prime time?
Thanks!

BookOwl:
@jmoenig, I want to fix it so that `1-1` is correctly parsed into `(1) - (1)` first.

jmoenig:
@BookOwl okay, cool!

BookOwl:
@jmoenig, I think that this is ready

jmoenig:
Thanks!

objects.js
```diff
@@ -2634,100 +2634,164 @@ SpriteMorph.prototype.searchBlocks = function (
...
+                }
+                tokens.shift();
+                exp = [op.val, exp, precedence1(tokens)];
+            }
+            return exp;
         }
-        return format(inputs[0]);
     }
-
     function blockFromAST(ast) {
```
jmoenig: Can you explain why your ```blockFromAST()``` function is different from my version? Is it because I've been changing it after you forked it?

BookOwl: The only difference is that I haven't added the boolean operators.

jmoenig: Ah, I see. So we can just use my ```blockFromAST()``` function on your ASTs, and you could add the Boolean operators to your parser, right? Or, if that's too much trouble we first parse it your way, and - if that doesn't result in anything useful - mine :)

BookOwl: It should be pretty easy to add the boolean and relational operators, I just need you to tell me what precedence they should have.

jmoenig: Good idea, and good question! Least precedence, I think. But I'm sure @brianharvey knows all about it... :)

objects.js
```diff
@@ -2783,10 +2801,11 @@ SpriteMorph.prototype.reporterize = function (expressionString) {
             '*': 'reportProduct',
             '/': 'reportQuotient',
             '%': 'reportModulus',
-            'random': 'reportRandom'
+            'random': 'reportRandom',
+            'round': 'reportRound'
```
cycomachead: This confuses me. It might well be the right place, but `round` is monadic.

BookOwl: I moved round there because even though it is monadic, it is a different type of block from the other monads, and it is easier to treat it as a diadic function. I should change the comments to make this clear

objects.js
brianharvey: https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#Operator_precedence

BookOwl: Is it OK to use & and | for "and" and "or"?

brianharvey: Oh, there's a question.  People who program in infix expect && and || for the Boolean operators, so you should certainly accept them (so that copy-and-paste works).  It might be okay to accept & and | _also._  On the other hand, we really should add bitand and bitor to the primitive dyadic operators someday.

BookOwl: I was just asking because single character operators are easy.

brianharvey: I understand.  I guess I'm saying "no," at least not _instead of_ && and ||.

BookOwl: That's fine. I've already got && and || working, I just need to get ! (for not) to work

BookOwl: It's working! 😄

BookOwl:
OK, so the boolean and relational operators have been added! 😄 Here are the supported operators and their precedence, from lowest to highest:
```
&&, ||
>, =, <
+, -
*, /
! (a unary operator)
```

objects.js
BookOwl: Done

dekrain:
Soon 100 comments!

jmoenig:
Thanks, @BookOwl , this is looking great now! I'll pull it for the next - upcoming - release, probably sometime next week (the pull, not the release ^^)

BookOwl:
Thanks! :D

jmoenig:
Okay, looking at this again I'm noticing a few issues. I'll leave this open and figure out how to integrate it, but perhaps not in the upcoming release due in the next days. Hang on!

BookOwl:
Is there anything that I can fix?

jmoenig:
It's gotten more complicated and a lot bigger than what I was hoping for. I was hoping for something more minimalistic, because I'm basically reluctant to add so much code for a full-fledged parser to ```objects.js```. Perhaps I'll decide to add a new file for parsing stuff to the Snap code base, which is where this could end up in.

BookOwl:
OK

brianharvey:
In response to

> In any case, I don't know what we can fix that one b/c that's how JS displays numbers. Scientific notation is was what users will see if they were to get that as the result of a math block in Snap!, so it's not totally unexpected.

(posted in #1591 by mistake):

Bignums FTW!!!!

BookOwl:
How would bignums work with the rest of Snap!?

brianharvey:
We have a bignum library, which Jens won't let me install as a library because he says 4.1 will break it.  Not just bignums, but the complete Scheme numeric tower.  It replaces the arithmetic primitives and then everything just works.
bignum
rat
imag

BookOwl:
I don't think that this should rely on having an optional library, especially if Jens doesn't want the library added.

brianharvey:
Oh, sorry, I didn't mean to suggest that you should write code around bignums, it's just that someone said, on that other thread, that nothing can be done about switching representations because that's what JS does, and I'm providing a counterexample.

BookOwl:
@jmoenig, would this be more manageable if I broke out the expression parsing and translating code into a new file named `math_translator.js` or similar, and have that file expose a single function that took a string containing a math expression and returned a block, and changed `reporterized` to call that function?

jmoenig:
No, I'm not a fan of IIFEs. Let's keep this around until I'm less under fire to include features and change things for the urgently needed release (the release is urgend, because Chrome 55 breaks Snap's music feature, which the release will fix). Eventually I'll get around to find a good way to integrate this.

micahcowan:
Right now, function-call identifiers get looked up in a list of `reportMonadic` variants, and then in a mapping of operator names, which get passed to `SpriteMorph.prototype.blockForSelector()`.

Are there any thoughts on, if those two checks fail, trying the key/identifier directly in blockForSelector? i.e., in this block of code from `blockFromAST()`:

```javascript
        } else { // The function or operator has its own block
            block = SpriteMorph.prototype.blockForSelector(selectors[key]);
            inps = block.inputs();
        }
```

the `block =` line becomes something like:

```javascript
            block = SpriteMorph.prototype.blockForSelector(selectors[key] || key);
```

?

(For reference, the specific line I'm referring to is line 2873 of `objects.js` as of `fd2d8df8`)

Making that single-line change "just works" for me, to open things up to use expressions like: `xPosition() / 2 + 10`: 
image

with the caveat that it currently cheerfully allows you to use command blocks in your expressions as well (`turnLeft(15) + 10`, for instance:) 
image

But that could be remedied with a quick additional check for appropriate block types...

I guess one potential issue with this is that not all reporter blocks are actually available via `blockForSelector`; for instance, AFAICT there's no way to get the "tempo" reporter in this way, or (say) "item N of list L". But since we're well into "power user" territory with this feature anyway (though I'm very excited by how much time this can save for power users!), maybe that can just be a caveat. Or maybe it's an excuse to create more tags to get at the core selectors?

Which still leaves the fact that custom reporters won't be selected in this fashion. Which is probably fine, for the same "power user" reasoning?

Anyway, obviously this is a "nice to have", not a "this should really be a part of this", and especially if I'm missing something that makes this much more complex than a simple one-line change + block-type check, this could either become a separate feature request after the current code has been pulled, or just an interesting idea that didn't get implemented. :)

cycomachead:
Hmm. I have mixed feelings on going down that route too far. On the one hand, being able to pop simple reporters in place is immensely handy, and I do feel like things should work properly if they are deployed. However, this starts to sound just like a text form of Snap! Which I'm pretty sure we all do not want kids to be using. (And believe me they will try...). 

I wonder if it makes sense to re-think some aspects of using the current text input mechanism to allow building an expression up left to right so that you could do something like insert a block in a slot and then put a new block "under" it so you could build an expression left to right instead of middle out. (I hope that made sense. That was thinking out loud, not recommending that we do such things necessarily.)

-- 
Michael Ball
From my iPhone
michaelball.co

brianharvey:
We _do_ want to support "a text form of Snap<i>!</i>," namely the keyboard editor.  We should consider how the math translator fits into that larger project some time after the immediate coming release.

BookOwl:
@micahcowan, that's really neat! I think that I'll add it in, if there are no objections from anybody else.

PR #1568, opened 2016-12-21
Langs update

jguille2:
Hi!
I'm working in #1551, making some scripts for an easier translation synchronization (I hope tomorrow I'll share more info), and I found some things on the way...
  - We have 39 languages (38 lang-xx translations and English). Chinese lang_packs codes are not according to ISO standards. According with cch (Traditional Chinese translator), I've updated them: **zh_CN for "Simplified Chinese"** (displaying "简体中文") and **zh_TW for "Traditional Chinese"** (displaying "繁體中文").
  - I've deleted duplicate translation strings in Catalan and Portuguese. Now, I note **Snap has 800 translatable text strings**. 531 are noted in lang-de, 244 more added in lang-pt, and 25 more in lang-ca.

We continuous in #1551 for this sync...

Thanks,

Joan

### DIFF
--- a/lang-ca.js
+++ b/lang-ca.js
@@ -1122,9 +1122,9 @@ SnapTranslator.dict.ca = {
     'Command\n(C-shape)':
         'Comanda\n(en forma de C)',
     'Any\n(unevaluated)':
-        'Qualsevol\n(sense evaluar)',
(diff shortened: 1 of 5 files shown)


jmoenig:
Thanks!

PR #1570, opened 2016-12-21
Report infix parser errors

dekrain:


cycomachead:
FWIW, I think we should have a separate discussion about how we report errors for this new feature.

IMO, we are beginning to conflate two distinct ideas in one text box because it's convenient. But I think we also have to careful how we present errors, because in the ideal sense we actually show students where the error is occurring. Otherwise, the math parser uses symbols that students could use in their own blocks and it's possible they could see errors when not even trying to get a parsable expression.

PR #1580, opened 2016-12-26
Add BHs Eisenbergification library

cycomachead:


cycomachead:
We really should try to keep the website state up to date with this repo. Otherwise things get out of sync...

brianharvey:
Oh...  the library isn't totally finished (have to add the per-sprite things) and that led me into a detour inventing hierarchical menus... so no hurry about pulling this.  But thanks.

cycomachead:
Ah, well, while it's in progress, the setting should also distinguish between serialized settings and local browser settings somehow.

I think it might also be cool if there were one menu, but it dynamically between between boolean and text / number / etc inputs. :) The `of` block does this.

jmoenig:
Michael is right, in general please let's *never* extend Snap by deploying any experimental changes directly to the production environment, but to the *dev* folder and to Github *first*, so we can review and test it.

brianharvey:
okay.

dekrain:
I have proposition. If there are project name setting, I'd like to add `isCloudProj` flag.

brianharvey:
Hmm, there's a "source" variable whose value is "local" or "cloud" or "examples" so I guess that could be used.  Why do you want it?

cycomachead:
> Hmm, there's a "source" variable whose value is "local" or "cloud" or "examples" so I guess that could be used. Why do you want it?

No, please don't expose this. Not that it would necessarily because things but that would mess with the default saving mode and I think it's best to keep people saving to the cloud as often as possible.

brianharvey:
Oh, I thought he wanted to read it, not write it.  Writing it would be wrong because, you know, it wouldn't be _true!_  The project came from where it came from.

dekrain:
Yes, I wanted option in `setting []` block

jmoenig:
I can't even look at this without getting myself all worked up into fits of rage!

Just to recap, we *did* agree I'd let you have your settings library, but only on the *minimal* conditions that it would *never* cause any *permanent* changes to a user's global settings. That is, things like language, zoom level for blocks, flat design, table support, etc. etc. can *never, ever* be saved in the browser. The user *must* be able to reload Snap and get *their* settings back, not something that some other project caused.

There are other settings that are almost equally infuriating, *especially* those that relate to meta-gestures such as switching to and from presentation mode that I *never* want to see in any project. We need to have a long and serious discussion about these, and I need to have the last word on whether we advertise those to users in the libraries browser, or whether you just use them in your own projects.

But the first and foremost priority is to make sure that no global change is made permenent by a script.

jmoenig:
Wait! That's already in the repo, right? We're going to close this one.

cycomachead:
Is it? I thought it was on the website but not the repo. 

-- 
Michael Ball
From my iPhone
michaelball.co

PR #1581, opened 2016-12-27
hierarchical menus

brianharvey:
Implemented in morphic.js (several changes to MenuMorph and
MenuItemMorph); used for custom block menu inputs in blocks.js

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -7533,7 +7533,7 @@ InputSlotMorph.prototype.setContents = function (aStringOrFloat) {
 InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
     var choices = this.choices,
         key,
-        menustack = [],
+        menuStack = [],
         menu = new MenuMorph(
             this.setContents,
             null,
@@ -7556,10 +7556,12 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
                 menu.addLine();
         // } else if (key.indexOf('§_def') === 0) { 
         // menu.addItem(choices[key].blockInstance(), choices[key]);
-            } else if (key.charCodeAt(key.length-1) == 0x25ba) {
+            } else if (key.charCodeAt(key.length - 1) == 0x25ba) {
                 // Submenu
+				// 0x25ba = Unicode "BLACK RIGHT-POINTING POINTER"
+				// (not to be confused with "black right-pointing TRIANGLE")
                 menu.addItem(key, choices[key]);
-                menustack.push(menu);
+                menuStack.push(menu);
                 menu = new MenuMorph(
                                      this.setContents,
                                      null,
@@ -7567,23 +7569,24 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
                                      this.fontSize
                                  );
             } else if (key.charCodeAt(0) == 0x25c4) {
-                var oldmenu=menustack[menustack.length-1],
-                    items=oldmenu.items;
-                items[items.length-1][6] = menu;
+				// "BLACK LEFT-POINTING POINTER"
+                var oldmenu = menuStack[menuStack.length - 1],
+                    items = oldmenu.items;
+                items[items.length - 1][6] = menu;
                 oldmenu.addChild(menu);
-                menu = menustack.pop();
+                menu = menuStack.pop();
             } else {
                 menu.addItem(key, choices[key]);
             }
         }
     }
 
-    while (menustack.length > 0) {
-        var oldmenu=menustack[menustack.length-1],
+    while (menuStack.length > 0) {
+        var oldmenu=menuStack[menuStack.length - 1],
             items=oldmenu.items;
         items[items.length-1][6] = menu;
         oldmenu.addChild(menu);
-        menu = menustack.pop();
+        menu = menuStack.pop();
     }
 
     if (menu.items.length > 0) {
--- a/blocks.js
+++ b/blocks.js
@@ -2551,7 +2551,7 @@ BlockMorph.prototype.deleteBlock = function () {
         });
     }
     if (this instanceof ReporterBlockMorph &&
-			((this.parent instanceof BlockMorph)
+                        ((this.parent instanceof BlockMorph)
             	|| (this.parent instanceof MultiArgMorph)
             	|| (this.parent instanceof ReporterSlotMorph))) {
         this.parent.revertToDefaultInput(this);
@@ -7533,6 +7533,7 @@ InputSlotMorph.prototype.setContents = function (aStringOrFloat) {
 InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
     var choices = this.choices,
         key,
+        menustack = [],
         menu = new MenuMorph(
             this.setContents,
             null,
@@ -7553,13 +7554,38 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
         if (Object.prototype.hasOwnProperty.call(choices, key)) {
             if (key[0] === '~') {
                 menu.addLine();
-            // } else if (key.indexOf('§_def') === 0) {
-            //     menu.addItem(choices[key].blockInstance(), choices[key]);
+        // } else if (key.indexOf('§_def') === 0) { 
+        // menu.addItem(choices[key].blockInstance(), choices[key]);
+            } else if (key.charCodeAt(key.length-1) == 0x25ba) {
+                // Submenu
+                menu.addItem(key, choices[key]);
+                menustack.push(menu);
+                menu = new MenuMorph(
+                                     this.setContents,
+                                     null,
+                                     this,
+                                     this.fontSize
+                                 );
+            } else if (key.charCodeAt(0) == 0x25c4) {
+                var oldmenu=menustack[menustack.length-1],
+                    items=oldmenu.items;
+                items[items.length-1][6] = menu;
+                oldmenu.addChild(menu);
+                menu = menustack.pop();
             } else {
                 menu.addItem(key, choices[key]);
             }
         }
     }
+
+    while (menustack.length > 0) {
+        var oldmenu=menustack[menustack.length-1],
+            items=oldmenu.items;
+        items[items.length-1][6] = menu;
+        oldmenu.addChild(menu);
+        menu = menustack.pop();
+    }
+
     if (menu.items.length > 0) {
         if (enableKeyboard) {
             menu.popup(this.world(), this.bottomLeft());
--- a/morphic.js
+++ b/morphic.js
@@ -7682,19 +7682,21 @@ MenuMorph.prototype.unselectAllItems = function () {
     this.changed();
 };
 
-MenuMorph.prototype.popup = function (world, pos) {
+MenuMorph.prototype.popup = function (world, pos, submenu) {
     this.drawNew();
     this.setPosition(pos);
     this.addShadow(new Point(2, 2), 80);
     this.keepWithin(world);
-    if (world.activeMenu) {
+    if (world.activeMenu && !submenu) {
         world.activeMenu.destroy();
     }
     if (this.items.length < 1 && !this.title) { // don't show empty menus
         return;
     }
-    world.add(this);
-    world.activeMenu = this;
+    if (!submenu) {
+        world.add(this);
+        world.activeMenu = this;
+    }
     this.world = world; // optionally enable keyboard support
     this.fullChanged();
 };
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
     this.selection = aMenuItem;
 };
 
+MenuMorph.prototype.mouseLeave = function () {
+    if (this.parent instanceof MenuMorph) {
+        // this is a submenu so we vanish when lose focus
+        var mouses = this.world.hand.allMorphsAtPointer();
+        if (contains(mouses, this.parent)) {
+            var items = mouses.filter(function (morph)
+                                      {return morph instanceof MenuItemMorph});
+            if (items[0].doubleClickAction != this) {
+                this.destroy();
+            }
+        } else {
+            this.destroy();
+        }
+    }
+}
+            
+
 MenuMorph.prototype.destroy = function () {
+var myself = this;
+
     if (this.hasFocus) {
         this.world.keyboardReceiver = null;
     }
+
+    if ((this.parent instanceof MenuMorph) && this.world) {
+        // If I'm a submenu, un-highlight the item pointing to me.
+        var mouses = this.world.hand.allMorphsAtPointer(),
+        items = this.parent.children.filter(function (child) {
+            return (child instanceof MenuItemMorph) &&
+            (child.doubleClickAction instanceof MenuMorph) &&
+            !(contains(mouses, child)); });
+        items.forEach(function (item) {item.image = item.normalImage;
+                                       item.changed(); });
+}
+
+    // Also kill my submenus if any.
+    var triggers = this.children.filter(function (each) {
+        return (each instanceof MenuItemMorph) &&
+                (each.doubleClickAction instanceof MenuMorph);
+    });
+    triggers.forEach(function (item) {item.doubleClickAction.destroy();});
     MenuMorph.uber.destroy.call(this);
 };
 
@@ -9403,16 +9442,34 @@ MenuItemMorph.prototype.mouseEnter = function () {
         this.image = this.highlightImage;
         this.changed();
     }
+    if (this.doubleClickAction instanceof MenuMorph) {
+        var w=this.parentThatIsA(WorldMorph);
+        this.doubleClickAction.popup(w,
+                                     new Point(this.parent.bounds.right()-10,
+                                     w.hand.position().y-10),
+                                     true);
+        if (!contains(this.parent.children, this.doubleClickAction))
+            this.parent.addChild(this.doubleClickAction);
+    }
     if (this.hint) {
         this.bubbleHelp(this.hint);
     }
 };
 
 MenuItemMorph.prototype.mouseLeave = function () {
-    if (!this.isListItem()) {
+    if (this.doubleClickAction instanceof MenuMorph) {
+		if (!contains(this.world().hand.allMorphsAtPointer(),
+                     this.doubleClickAction)) {
+        this.doubleClickAction.destroy();
+        this.image = this.normalImage;
+        this.changed();
+}
+    } else if (!this.isListItem()) {
         this.image = this.normalImage;
         this.changed();
     }
+
+
     if (this.hint) {
         this.world().hand.destroyTemporaries();
     }
@@ -9435,7 +9492,11 @@ MenuItemMorph.prototype.mouseMove = function () {
 
 MenuItemMorph.prototype.mouseClickLeft = function () {
     if (!this.isListItem()) {
-        this.parent.destroy();
+        var topmenu = this.parent;
+        while (topmenu.parent instanceof MenuMorph) {
+            topmenu = topmenu.parent;
+        }
+        topmenu.destroy();
         this.root().activeMenu = null;
     }
     this.trigger();
--- a/morphic.js
+++ b/morphic.js
@@ -9479,8 +9479,10 @@ MenuItemMorph.prototype.mouseDownLeft = function (pos) {
     if (this.isListItem()) {
         this.parent.unselectAllItems();
         this.escalateEvent('mouseDownLeft', pos);
-    }
-    this.image = this.pressImage;
+        this.image = this.pressImage;
+    } else if (!(this.doubleClickAction instanceof MenuMorph)) {
+        this.image = this.pressImage;
+	}
     this.changed();
 };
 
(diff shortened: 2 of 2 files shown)


brianharvey:
Woo hoo!  I didn't destroy everything by branching and making the PR!

Am I going to hell because I added the "&& this.world" at morphic.js line 7847 without really understanding why it's needed? (That is, I don't understand why it would ever not be a (the) WorldMorph.)

dekrain:
```JavaScript
Morph.prototype.world = function() {
if (this instanceof WorldMorph)
  return this;
if (this.parent)
  return this.parent.world();
return null;
};
```

blocks.js
```diff
@@ -7553,13 +7554,38 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
         if (Object.prototype.hasOwnProperty.call(choices, key)) {
             if (key[0] === '~') {
                 menu.addLine();
-            // } else if (key.indexOf('§_def') === 0) {
-            //     menu.addItem(choices[key].blockInstance(), choices[key]);
+        // } else if (key.indexOf('§_def') === 0) { 
+        // menu.addItem(choices[key].blockInstance(), choices[key]);
+            } else if (key.charCodeAt(key.length-1) == 0x25ba) {
```
cycomachead: This should be a constant and given a name and a comment explaining what it is/looks like.

morphic.js
```diff
@@ -7682,19 +7682,21 @@ MenuMorph.prototype.unselectAllItems = function () {
     this.changed();
 };
 
-MenuMorph.prototype.popup = function (world, pos) {
+MenuMorph.prototype.popup = function (world, pos, submenu) {
```
cycomachead: Is this intended to be a boolean value? If so it needs to be documented, but it seems like it would be really nice to move away from functions that have boolean arguments. They're a code smell.

blocks.js
cycomachead: Also spacing around the `-`

blocks.js
```diff
@@ -7553,13 +7554,38 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
+                menu.addItem(key, choices[key]);
+                menustack.push(menu);
+                menu = new MenuMorph(
+                                     this.setContents,
+                                     null,
+                                     this,
+                                     this.fontSize
+                                 );
+            } else if (key.charCodeAt(0) == 0x25c4) {
+                var oldmenu=menustack[menustack.length-1],
```
cycomachead: Spacing around the `=`. Names should also be camelCase (except classes which are PascalCase)

I also think `oldMenu` is unclear. It seems like we still want it later, so maybe it's the `containerMenu` or something.

blocks.js
```diff
@@ -7553,13 +7554,38 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
+                    items=oldmenu.items;
+                items[items.length-1][6] = menu;
```
cycomachead: I don't think you can easily do something about this, but accessing only item 6 of an array is troublesome. Loads of room for error, and it's not very communicative.

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
     this.selection = aMenuItem;
 };
 
+MenuMorph.prototype.mouseLeave = function () {
+    if (this.parent instanceof MenuMorph) {
+        // this is a submenu so we vanish when lose focus
```
cycomachead: When who/what loses focus?

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+        var mouses = this.world.hand.allMorphsAtPointer();
+        if (contains(mouses, this.parent)) {
+            var items = mouses.filter(function (morph)
```
cycomachead: All variables should be declared at the top of the function. Also the filter function should be 
```js
.filter(function () {
     return;
}
```

(it can also be extract into a named function `isMenuMorph` and then you'd do `.filter(isMenuMorph)` which looks much nicer.

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+                                      {return morph instanceof MenuItemMorph});
+            if (items[0].doubleClickAction != this) {
```
cycomachead: Does this actually work? I didn't think you could compare objects like that in JS.

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+
 MenuMorph.prototype.destroy = function () {
+var myself = this;
+
     if (this.hasFocus) {
         this.world.keyboardReceiver = null;
     }
+
+    if ((this.parent instanceof MenuMorph) && this.world) {
+        // If I'm a submenu, un-highlight the item pointing to me.
```
cycomachead: This actually seems unnecessary. In macOS (and to my knowledge Windows) when selecting / hovering over a submenu the parent menu item _also_ stays selected. I think that's to indicate a) where the menu came from, and b) if you un-hover of the submenu then it will still be visible.

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+        var mouses = this.world.hand.allMorphsAtPointer(),
+        items = this.parent.children.filter(function (child) {
```
cycomachead: Assuming this function sticks around I think `items` needs to be made more explicit. It's hard to parse that filter function.

I'm also beginning to not like the name `mouses` as it implies a bunch of courses and not morphs. Maybe it could just be `morphsAtPointer` or `morphsAtMouse` or mouseMorphs` or something.

Also `items` should be indented 4 more spaces.

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+            return (child instanceof MenuItemMorph) &&
+            (child.doubleClickAction instanceof MenuMorph) &&
+            !(contains(mouses, child)); });
```
cycomachead: If the opening `{` is at the end of a line, then the closing `}` should be at the start of it's own line.

morphic.js
```diff
@@ -9403,16 +9442,34 @@ MenuItemMorph.prototype.mouseEnter = function () {
...
+    }
     if (this.hint) {
         this.bubbleHelp(this.hint);
     }
 };
 
 MenuItemMorph.prototype.mouseLeave = function () {
-    if (!this.isListItem()) {
+    if (this.doubleClickAction instanceof MenuMorph) {
+		if (!contains(this.world().hand.allMorphsAtPointer(),
```
cycomachead: This doesn't need to be indented so much.

morphic.js
```diff
@@ -9403,16 +9442,34 @@ MenuItemMorph.prototype.mouseEnter = function () {
         this.image = this.highlightImage;
         this.changed();
     }
+    if (this.doubleClickAction instanceof MenuMorph) {
+        var w=this.parentThatIsA(WorldMorph);
```
cycomachead: Spacing.

morphic.js
```diff
@@ -7818,10 +7820,47 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+        items = this.parent.children.filter(function (child) {
+            return (child instanceof MenuItemMorph) &&
+            (child.doubleClickAction instanceof MenuMorph) &&
+            !(contains(mouses, child)); });
+        items.forEach(function (item) {item.image = item.normalImage;
+                                       item.changed(); });
+}
+
+    // Also kill my submenus if any.
+    var triggers = this.children.filter(function (each) {
```
cycomachead: Declare `triggers` above, but this `filter` looks nice!

morphic.js
```diff
@@ -9434,13 +9493,24 @@ MenuItemMorph.prototype.mouseMove = function () {
 };
 
 MenuItemMorph.prototype.mouseClickLeft = function () {
-    if (!this.isListItem()) {
-        this.parent.destroy();
-        this.root().activeMenu = null;
+	if (!(this.doubleClickAction instanceof MenuMorph)) {
+        if (!this.isListItem()) {
```
cycomachead: Indent more.

brianharvey:
Thanks, I'll look at those in more detail later today, but I'm glad only one of them looks like a potential bug -- although I'm pretty sure it does work.

I don't think you (and Jens) are ever going to get me to love camel case.  (In good programming languages, identifiers are case-insensitive anyway. :P)  Certainly not for very-local variables like "oldmenu" which clearly means "remembering the old value of menu"!  So it would be downright misleading to capitalize the m.

I thought I'd fixed all the spacing issues but I guess I missed a couple.  The root of the problem is that in my country tabs always mean eight spaces.

brianharvey:
@dekrain I don't think Jens would want me to make unnecessary changes to the WorldMorph code.

dekrain:
No, I just wrote the world function's code.
I meant to tell `this.world` is a function.

blocks.js
brianharvey: Spacing, OK.  Added comment on Unicode name.  Not convinced about putting a constant definition screensful away for something only used once in the code.

blocks.js
brianharvey: Spacing, done.  Just to make you happy I changed menustack to menuStack.  As I said before, I think it would be *wrong* to capitalize the M in oldmenu.

blocks.js
brianharvey: Yes, I agree.  This bothered me, too!  Take it up with Jens...

morphic.js
brianharvey: Comment added.  No idea what you have against Boolean args.  Okay, not if there are five of them in one procedure.

morphic.js
brianharvey: This menu, of course.  Is that really not clear from the fact that we're in MenuMorph.mouseLeave()?

morphic.js
brianharvey: Going along with your brace fetish since I don't have a competing brace fetish of my own.  Remind me what's wrong with local declarations?

blocks.js
cycomachead: I see two things that should be constants, but the point about them being constant isn't about use but clarity. Why say `0x25ba` is there's something like `SUBMENU_START`or who knows what. But a comment works too, since it is only used once.

blocks.js
cycomachead: I don't get why. CamelCaseing is about readability. Or if we really want, we can snake_case things, but that's probably not a good idea.

morphic.js
brianharvey: Aha, you're right!  Luckily the menu immediately gets recreated if we moved the mouse back to its MenuItemMorph.  So I eliminated that if, and all the nasty local variables you complained about as a side effect.  :-)

morphic.js
brianharvey: No, you're misunderstanding.  This is when _leaving_ the submenu, not when entering it.  If you just mouse away from the submenu and it disappears, the highlighting of its MenuItem should be undone too.

OTOH I'm not sure why this works; if you can't check object equality, how does "contains" work?  Or is the "this.world" that I don't understand somehow work around the comparison not working?  I'd be delighted if you could explain that!

cycomachead:
Well the rest of Snap! Doesn't use them so there's that. That's the main reason I suggested it. 

I think Snap / Jens doesn't use them because Crockford. So that's not helpful. 

I believe Crockford doesn't like them because a local declaration implies that a variable is only locally available but in JS things aren't scoped that way. Variables are available for the rest of the function. 

-- 
Michael Ball
From my iPhone
michaelball.co

morphic.js
cycomachead: Well blame Armando for this habit. But there's a few things:

Boolean rags typically indicate that there should be two functions. A booolean parameter can't indicate anything about what it might control. So now if you later read a call to popup (which should be popUp..) you won't know what the flag is doing.

morphic.js
brianharvey: A bunch of courses would be "mousses"! :)
Okay, so, I figured this out, I think -- the comparison was indeed failing, but its only purpose was optimization; what's really happening is that _all_ the MenuItems are being un-highlighted, and the one we're actually pointing at then gets re-highlighted.

So most of what you didn't like above is gone now.

morphic.js
brianharvey: Done.

morphic.js
cycomachead: The second half of the sentence is missing a noun so it looks funny. ( you can blame PNH for instilling proper sentences.)

morphic.js
brianharvey: ok

blocks.js
liam4: Don't snake_case things unless you're programming in Python (or languages where snake_case is appropriate).

I find it *usually* best to go with whatever the language constants do (if the designers have blessed you with consistent naming patterns!) – in JavaScript, variables are (usually) capitalizedLikeThis (e.g. `'my string'.toUpperCase()`, `document.createElement()`).

morphic.js
brianharvey: righto, all done.

morphic.js
liam4: Local variable declarations – using `var` – are scoped to the function, so it usually makes more sense to declare the variable at the top of the function:

```js
function foo(arg) {
  console.log(x); // the value undefined
  console.log(y); // throws a reference error

  if (arg === 3) {
    var x = 123;
  } else {
    var x = 15;
  }

  console.log(x); // 123 or 15
}
```

See how `x` had the value `undefined` but when we tried to get the value of `y`, we got a reference error? This becomes a real problem in a case like this:

```js
function foo(x) {
  return function() {
    var y = x + 2

    // later in the logic..

    if (Math.random() > 0.5) {
      var x = 15 * 7 * 1.5
      y += x
    }

    return y
  }
}
```

So if we call `foo(3)`, we get a function that would *theoretically* return either `5` (`y = x + 2`) or `162.5` (`y` (5) `+= 15 * 7 * 1.5`), but we *actually* get `NaN`! That's because our later declaration of `var x = 15 * 7 * 1.5` overshadowed the earlier declaration of `x` (as an argument to `foo`), so when we did `var y = x + 2` we actually did `var y = undefined + 2`, which is `NaN`. (Any later math we might do to `NaN` also gets us `NaN`; that's just a property of `NaN`.)

So if we move the variable declarations to the top, we end up getting this:

```js
function foo(x) {
  return function() {
    var x // <--- 'var x' moved here
    var y = x + 2

    // later in the logic..

    if (Math.random() > 0.5) {
      x = 15 * 7 * 1.5 // <--- no var
      y += x
    }

    return y
  }
}
```

..and then it's clear that `x` is `undefined` when we add it in `y = x + 2`, so we've figured out the problem.

Convoluted examples aside, hopefully this makes at least a little bit of sense :)

IMO this is one of the biggest problems with JavaScript, but lately it's gotten better since you can use `let` and `const`, which are scoped to the *block* they're in:

```js
function foo(arg) {
  console.log(x); // throws a reference error

  let x;
  if (arg === 3) {
    x = 123;
  } else {
    x = 15;
  }

  console.log(x); // 123 or 15
}
```

morphic.js
liam4: Perhaps an issue of inconsistent tabs and spaces and such? I think the rest of Snap<em>!</em> uses spaces (four spaces being a single level of indentation).

morphic.js
cycomachead: Thanks, Liam. I thought I wrote a much simpler response via email but I don't see it. 

(Now if only we could use some ES6 stuff in Snap! that would be cool.)

blocks.js
cycomachead: I was being sarcastic with snake case. Lol. 

Even if you have poor descisions it's good to write consistent code or at least declare which code isn't going to follow conventions and why. 

I really ought to revive my ESlint branch.

brianharvey:
> because Crockford

Funny, I just delivered a rant about that on the Scratch forum. :)

I don't see why we should deny ourselves the wonderful documentation virtue of declaring the variable as close as possible to its use, so the reader doesn't have to go frantically searching for it, just because JS is badly designed.  It doesn't do any _harm_ to have the declaration JIT, unless you're trying to use the same name for two local variables in the same procedure, in which case nothing is going to save you from having bad code!

So, I think you're supposed to click something to say if you think I fixed your problems or not...

brianharvey:
@dekrain Oh.  But MenuMorph overloads that definition with one whose value is a WorldMorph, not a Function.  :(

dekrain:
Ouch, my wrong.

morphic.js
brianharvey: Second paragraph now obsolete, see below

morphic.js
brianharvey: Well, I agree that a comment (which I added) helps.  But I disagree about the two-procedures.  Abstraction!  With two procedures, when you change one of them and forget to change the other, things will go wrong.

morphic.js
brianharvey: Yeah, as I said, where I come from tab always means eight spaces, but I set my preferences to use four spaces so there shouldn't be any new ones of these.

blocks.js
brianharvey: The other thing that you think should be a constant is Jens's, and is in a comment anyway.  :)

blocks.js
brianharvey: In a language in which case in identifiers is significant, "oldMenu" means "the old value of Menu."  What I want is "the old value of menu," which is different, and can only reasonably be spelled "oldmenu."

morphic.js
brianharvey: :)

Very annoying that when you type a colon, github fills half your screen with a menu of suggestions, none of which is the one you 95% of the time mean!  :(

morphic.js
brianharvey: @liam4 Anyone who uses names such as x and y in programming (other than in a procedure that plots graphs of functions) deserves whatever happens, regardless of where declared.  Come up with an example in which someone using self-documenting variable names will be burned that way.

morphic.js
cycomachead: Yeah, and makes sense. I think I just missed the "un-" before. I think the cleaned up version looks much better. :)

morphic.js
cycomachead: Well that suggests that the two procedures should be re-using the same underlying but not exposed function. 
I think the main argument is that a boolean typically indicates a function is trying to do two different things. In this case the two different things are related, but I still think the use case of `.popup(x, y, true)` isn't clear. (Unless you define `showSubMenus = true`, and you use that... [also x and y are placeholders since those are bad names for anything other than coordinates, pretty much.])

cycomachead: Also now that I think about it:

Should a subMenu be a menu whose parent is a `MenuMorph`? Otherwise isn't the parent an IDE or some other morph? In that case we could ditch the parameter an define `MenuMorph::isSubMenu()` and use that?

morphic.js
```diff
@@ -9403,16 +9439,34 @@ MenuItemMorph.prototype.mouseEnter = function () {
         this.image = this.highlightImage;
         this.changed();
     }
+    if (this.doubleClickAction instanceof MenuMorph) {
+        var w = this.parentThatIsA(WorldMorph);
```
cycomachead: This should be `world` (I should have caught that earlier...)
Reason: Other functions that need a world reference use `world`

morphic.js
liam4: I wholly agree that it's not that much of an issue in practice, though at least you know the semi-reasoning behind people yelling at you about the issue!

morphic.js
brianharvey: Yes, a submenu is a menu whose parent is a menu.  The trouble is that popup is first called when it doesn't have any parent at all; see MenuItemMorph::mouseEnter.

Oh... is the contains in mouseEnter not going to work?  Because then the main menu's children will have this submenu a million times.

morphic.js
brianharvey: Right you are.  I had seen this.world used for the world in MenuMorph, so I tried that and of course it didn't work in MenuItemMorph, and I hadn't yet discovered this.world(), so I fell back on something I was confident about.  Fixing...

morphic.js
brianharvey: :)

morphic.js
```diff
@@ -9434,13 +9490,24 @@ MenuItemMorph.prototype.mouseMove = function () {
...
+            topmenu.destroy();
+            this.root().activeMenu = null;
+        }
+        this.trigger();
     }
-    this.trigger();
 };
 
+MenuItemMorph.prototype.mouseDoubleClick = function () {
+    if (!(this.doubleClickAction instanceof MenuMorph))
```
cycomachead: I thought I commented on this but there should always be braces around one-line `if`s.

Otherwise this happens: http://www.dwheeler.com/essays/apple-goto-fail.html

jmoenig:
Okay guys, now we have 3 branches of Snap in the repo, that's not what I want. Please remove these branches and issue pull-request to the single main Snap branch. Then I'll review.

morphic.js
brianharvey: Fine.  I do always do that, but didn't in this case because the IF is the entire body of the procedure!  But okay, just for you.

cycomachead:
The branches _are_ to the master branch they are just from different branches. And I think mine can be deleted since the PR was merged. 

There shouldn't be anything wrong with having more branches. The master branch is still the default and what all PRs are made against unless someone specifically chooses not to. 
-- 
Michael Ball
From my iPhone
michaelball.co

jmoenig:
Thanks for the changes, Brian!

jmoenig:
Wait! Why did you close this? Didn't it just work from your other fork? Jeez, I suck at Git!

brianharvey:
Git automatically closed the PR when I deleted the branch.  But I made a new PR from my new fork.  Should be okay.

PR #1587, opened 2016-12-28
Hierarchical Menus

brianharvey:
Changing merge request to be from different fork as per Jens


jmoenig:
Thanks, Brian!

So, my first impression as a user is that your submenus feel "right". I like how the item that triggers the submenu to pop up remains selected and I like how when the mouse leaves the submenu only the submenu goes away. I'm also pleased how submenus can be arbitrarily nested and still feel alright to me as a user. This is nice! I was afraid that in - literal - edge cases things would feel awkward to users, for example when you expand a cascade of submenus that are bigger than the primary one in a location where they completely overlay each other (say, in the World's bottom right corner). But your mechanism solves these cases nicely! In my - very few - test cases so far I never got stuck in a situation where I couldn't get back again.

In general, however, I'm suspicious about submenus. To me it feels that whenever I want a submenu that's an indicator that I really want something else, probably a better widget than a menu, but since menus are easy to create that's what I end up using (I'm totally guilty of this myself, just look at all the stuff in Snap that you need a context menu for - they're quick'n easy to set up ^^).

Especially for custom drop-downs in blocks I've very conscientiously decided not to make the input slot options dialog box resizable, and even to make it rather small to begin with, so people use options sparingly. I know that there's an unresolvable conflict between wanting as few blocks as possible and not wanting big or even hierarchical menus. Although I'd probably opt for more expressive blocks rather than trying to cram a whole class hierarchy browser into a single nested menu in a single input slot of a single block. But, as I've conceded, it's a conflict that cannot be resolved, so we need to find the right balance, and nested menus might be a part of that.

I love how you pull this off with so little code changes! I'm also fine with not introducing a separate subclass of MenuItemMorph for this - whatever you feel more comfortable with is fine here, I think. But I'm wondering whether in the end we might want to consider a more thorough redesign of the Morphic Menu system. The current one is really quite minimalistic (just as I like it), but introducing nesting also opens up a whole Pandora's box of issues to consider, just to mention a few:

* Should we align all right-arrows in all menu items into a virtual column?
* Perhaps, now that we're considering submenu indicators (the right arrow/triangle character) we also want to support other "adornments", like keyboard-shortcuts (and put them into the same column).
* With submenus we could also think about making the layout configurable. Right now, menus are always columns, perhaps there could also be a "row" layout orientation so we can use them as menu bars (this could be especially interesting if the top-level menu items are not textual but pictorial 
* would be perhaps want a choice of different submenu indicators other than a right-arrow? Perhaps even none at all?

There are also some loose ends:

* submenus are currently not keyboard accessible
* the api needs to be more expressive for programmers to use outside of custom drop-downs.

The part I really, really hate the most is adding nesting-syntax to the input options dialog. That text-field needs to be the utmost simplistic thing ever, but with nesting syntax you're beginning to turn it into a scripting-language thingy, and the the right-arrow left-arrow syntax is clearly horrible. This way to specify drop-down menus has been a kludgy hack all along, because the hardware folks needed it for their Arduino pins etc. Now that you're introducing custom-drop-downs big time requiring the users of a *blocks based* system to hack around in a complicated *text* syntax in a minimalitic bare-bones text input field (without syntax highlighting, no error console, no auto-complete etc.) is taking it way too far, I think. It's beginning to create a text mark-up language for block parts, and that can't be right. (I know, I know, I've started this, but that doesn't mean I like it or I have to agree to even further extent this kludge).

Instead of inventing a text language for block-input drop-downs we should probably add dynamic drop-downs that can be specified with ... blocks, i.e. reporters. The *right* way would be - I think - to do this ontop of a generic object system that gives us key-value dictionaries for free, and I really think it's worth doing that *before* thinking about or even implementing a more complex hierarchical drop-down specification syntax. I know you've also wanted this all along, and I think this should probably now be bumped up on our priority list. OOP and dynamic drop-downs (which probably also require "system" processes that can't be killed by the user, but that's another story...)

Of course, those dynamic menus might also require or encourage nesting, so this pull-request is useful anyway. And yes, until we have a better system we might as well add syntax to the options, and just tell people that it's a loose end... Ah, now I regret putting in this feature. See, this what happens whenever I agree to add a feature *you* come along and turn it into a DSL :-)

I'll look some more into the code tomorrow, but overall I like what I've seen so far. You've done a good job at imititating Snap's souce-code style :-D

Oh, yeah, you're right about the doubleClickAction kludge, of course. Instead of overloading that one why don't you just use the action slot (so the "action" can be a MenuMorph)? Oh wait! there are (at least) two modes in which menus can be used, and the one you're interested in is the "selection" mode, in which the "action" is the Menu's target, right? In this case you might want to add another explicit slot to MenuItemMorph, I guess. Okay, this reminds me to test both usage modes...

Anyway, this is coming along nicely. Thanks!

brianharvey:
Thanks!  I'm glad you like it.

I had thought about using > and < instead of ► and ◄, but was worried that someone would want an item called <expr> and things would go wrong.  I think the really right thing (well, okay, the sorta right thing) is for me to learn how to add a couple of small buttons to the dialog to generate those two characters.

I also thought about right-adjusting the arrowhead in the menu, but decided that that's a can of worms that can wait until the second release. :-)  Probably not really a can of worms, I just don't know where to begin -- there's all that code to figure out the width of the menu that I didn't want to read.

You're right, the whole interface is very texty; I was surprised you did it that way.  (Reminded me of the old, old BYOB procedure prototypes!)  It should totally be redesigned.  But I don't want to wait that long to get the sprite attribute blocks up and running so Paul can read pen properties and things like that.

Keyboard shortcuts would be awesome!  But even more important, I'd like users to be able to add buttons with menus to the toolbar, which is the other half of Eisenbergification.  (The actions of user menus would be Snap<i>!</i> scripts, of course.)  And, just so you don't blow your top, I'm not suggesting that users should be able to modify the existing menus, just add their own!  So no malware problem.  :-)

Right, keyboard accessibility.  I'll look at that.  Should be straightforward; you navigate as in any menu, and when you select a submenu item it opens the submenu.

I'm not sure what you mean, really, about the API.  Making the submenu is exactly the same as making any menu.  I suppose I could write a procedure that takes a MenuItem and a (sub)Menu and makes the former point to the latter.

But, we should do a release this week, and it should include submenus!  Nobody will object if they get better later.  Imho.

jmoenig:
Don't be silly, this isn't going into this week's release, we didn't even test it yet. Geez, we have *users*!

Two things: I'm fuming mad at your library that lets projects change the settings of those who just play the projects, especially settings such as language, blocks zoom, layout design. Fuming mad, actually, short of disabling the JS feature altogether tonight to protect users. This is the most serious disagreement we've had for years, I'm *not* kidding! The other: If you ever let a project or a script change the Snap GUI, by adding buttons and menus, this *will* be the end of my work on Snap!

brianharvey:
Did you get my email in which I proposed a compromise about the settings?

And, could you please explain why it's bad for users to be able to add buttons to the toolbar?  I get that you don't like the idea but it would really help me if you explained why.

cycomachead:
On the actual issues:

Are you only testing this on block menus? It would be nice w/ PR's is there was an example included so we didn't have to guess. :) 

> I had thought about using > and < instead of ► and ◄, but was worried that someone would want an item called and things would go wrong. I think the really right thing (well, okay, the sorta right thing) is for me to learn how to add a couple of small buttons to the dialog to generate those two characters.

IMO, there should be buttons OR we just find a way to build a submenu from nested lists (though that's a bigger change...). Currently even using `~` causes a few bugs/edge cases, and I definitely agree with Jens that we shouldn't start creating cryptic text systems to use. If a menu is a list of KVPs, then I think any key whose value is a list (or dictionary) should automatically have a submenu. 

I also still really find that adding a boolean argument to `popup()` to be confusing. It definitely impairs the readability of popup and seems super easy to leave off. Why not allow a `MenuMorph` to be able to detect whether or not it is a submenu? It seems like this is easy enough to achieve by having a property `MenuMorph.isSubMenu = true` or by checking if `menu.parent instanceof MenuMorph`?

---

On things unrelated to this PR: It'd be nice if some of the discussion about what belongs in Snap! happened somewhere other than just email. :)

brianharvey:
Apparently the menu morph is also used for list table views, and so I tried that to make sure it wasn't broken, but the issue there is negative testing, i.e., making sure that the list doesn't start looking like submenus.  I haven't tried adding submenus to non-user menus, though.  Perhaps I should address Jens's concern about the API first.

I would be totally happy to use lists to set up menus!  I was trying to make the minimal possible change to get submenus working, though, and I figured if ~ for separators is okay, so is this.  But you are pulling on the end of a long cord, because the _really_ right thing is to associate a _function_ with the input slot, taking the other input values as its inputs, and reporting an association list as you say.  And there should be reporters that report the systemy things that go into primitive menu slots, e.g., broadcasts.  My thinking was that all that stuff can wait for 4.1, and for 4.0.10 (I guess that's what's next) the text-based interface would do.  Does that make sense?

About the Boolean input, as I said, at the time popup is called the menu doesn't have a parent yet, so it doesn't know it's a submenu.  I suppose it would be possible to rearrange the sequence of events so that it would; I'll look into that tomorrow.

Thanks.

brianharvey:
Oh, P.S., a lot of discussion happens on the forum, and also there's a probably-way-out-of-date developers email list.  But in this particular case I think what you're missing is that Jens prefers not to yell at me in public.  :-)

cycomachead:
> Perhaps I should address Jens's concern about the API first.

yeah, I think that's why I was a little confused at first. I wasn't quite sure what to begin testing. But I think we could take the similar object structure to what current exists to define the pull-down menus in byob.js

> I would be totally happy to use lists to set up menus! I was trying to make the minimal possible change to get submenus working, though, and I figured if ~ for separators is okay, so is this.

Yeah, that's why I didn't complain about this at first. 😄 But since Jens said something similar, I figured I'd mention it. It's not a problem (IMO) to make multiple smaller changes first adding support for displaying submenus and then adding support for creating them.



> I suppose it would be possible to rearrange the sequence of events so that it would; I'll look into that tomorrow.

Ah, I think I missed that comment in all the discussion, but yeah I think rearranging might make sense. I can't see a situation where you'd need to create a submenu before knowing the parent. (I mean, logically as someone consuming an API.)



> Oh, P.S., a lot of discussion happens on the forum, and also there's a probably-way-out-of-date developers email list. But in this particular case I think what you're missing is that Jens prefers not to yell at me in public. :-)

I think I'm on _most_ of the email lists. But mostly I think some more public discussion wouldn't be terrible (especially if we ever find a way to hear teacher's comments about what they're doing), but also we can probably all do with a little less yelling at each other. :)

morphic.js
```diff
@@ -7682,19 +7682,25 @@ MenuMorph.prototype.unselectAllItems = function () {
     this.changed();
 };
 
-MenuMorph.prototype.popup = function (world, pos) {
+MenuMorph.prototype.popup = function (world, pos, submenu) {
```
jmoenig: this could probably be done without having to introduce the optional ```submenu``` flag, don't you think?

morphic.js
```diff
@@ -9434,13 +9490,25 @@ MenuItemMorph.prototype.mouseMove = function () {
...
+            }
+            topmenu.destroy();
+            this.root().activeMenu = null;
+        }
+        this.trigger();
     }
-    this.trigger();
 };
 
+MenuItemMorph.prototype.mouseDoubleClick = function () {
```
jmoenig: overloading ```doubleClickAction``` seems like a kludge, there must be a better way

morphic.js
```diff
@@ -9403,16 +9439,34 @@ MenuItemMorph.prototype.mouseEnter = function () {
...
+            this.parent.addChild(this.doubleClickAction);
+    }
     if (this.hint) {
         this.bubbleHelp(this.hint);
     }
 };
 
 MenuItemMorph.prototype.mouseLeave = function () {
-    if (!this.isListItem()) {
+    if (this.doubleClickAction instanceof MenuMorph) {
```
jmoenig: I find this very confusing to read. I think it would be clearer if you'd just add a property named ```submenu``` or so, instead of overloading a property that's meant for double clicks

morphic.js
```diff
@@ -7818,10 +7824,40 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
 
+MenuMorph.prototype.mouseLeave = function () {
+    if (this.parent instanceof MenuMorph) {
+        // this is a submenu so we vanish when lose focus
+		this.destroy();
+    }
+}
+            
+
 MenuMorph.prototype.destroy = function () {
```
jmoenig: something about this method seems overly complicated. What's all the scanning for? Can't you just destroy the menu and be done with? Why does this code need to be extended at all?

morphic.js
```diff
@@ -7818,10 +7824,40 @@ MenuMorph.prototype.select = function (aMenuItem) {
     this.selection = aMenuItem;
 };
 
+MenuMorph.prototype.mouseLeave = function () {
+    if (this.parent instanceof MenuMorph) {
+        // this is a submenu so we vanish when lose focus
+		this.destroy();
```
jmoenig: hmm... I guess this kinda makes sense. But MacOS leaves the submenu visible as the mouse leaves it. Since submenus are just submorphs (right?) can't we just leave them around until something else gets focus? Of course, I can see how this might lead to situations where a big submenu is in front of the primary one and doesn't let you get back...

morphic.js
```diff
@@ -9403,16 +9439,34 @@ MenuItemMorph.prototype.mouseEnter = function () {
         this.image = this.highlightImage;
         this.changed();
     }
+    if (this.doubleClickAction instanceof MenuMorph) {
+        var w = this.world();
```
jmoenig: please declare all local variables in a single ```var``` .statement at the top of the scope, i.e. the function, not inside code blocks that don't have their own scope. Also, ```w``` could be spelled out ```world``` :)

morphic.js
```diff
@@ -9422,8 +9476,10 @@ MenuItemMorph.prototype.mouseDownLeft = function (pos) {
     if (this.isListItem()) {
         this.parent.unselectAllItems();
         this.escalateEvent('mouseDownLeft', pos);
-    }
-    this.image = this.pressImage;
+        this.image = this.pressImage;
+    } else if (!(this.doubleClickAction instanceof MenuMorph)) {
+        this.image = this.pressImage;
```
jmoenig: in the case of menu items I think ```highlightImage``` is the correct one.

brianharvey:
Thanks.  I'll respond in detail (mainly by fixing things) tomorrow; it's 12:30am here and I'm too groggy to code.  :-)

morphic.js
cycomachead: It's not possible without some refactoring apparently (see 2 comments ago from BH), but he said he'll look into it. :)

morphic.js
```diff
@@ -7818,10 +7824,40 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+}
+            
+
 MenuMorph.prototype.destroy = function () {
+var myself = this,
+    subitems = function(menu) {
+        return menu.children.filter(function (child) {
+                    return (child instanceof MenuItemMorph) &&
+                           (child.doubleClickAction instanceof MenuMorph);
+			} )
```
cycomachead: Here? Shouldn't there just be no space? (the one at the end of the return should be there.

I actually thought that putting one here would be a parse error. You can't put `;` in argument lists, right?

blocks.js
```diff
@@ -7553,13 +7554,41 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
         if (Object.prototype.hasOwnProperty.call(choices, key)) {
             if (key[0] === '~') {
                 menu.addLine();
-            // } else if (key.indexOf('§_def') === 0) {
```
cycomachead: > (I guess we can actually remove these two lines, since they've been commented out for ages anyway)

oh, that would make me very happy. 

But BH says his indents look fine to him. It looks like he's using actual tab chars...

blocks.js
```diff
@@ -7553,13 +7554,41 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
-            //     menu.addItem(choices[key].blockInstance(), choices[key]);
+        // } else if (key.indexOf('§_def') === 0) { 
+        // menu.addItem(choices[key].blockInstance(), choices[key]);
+            } else if (key.charCodeAt(key.length - 1) == 0x25ba) {
```
cycomachead: IMO, the idea of using something not easily type-able was the correct one, since `{ }` would mess up dropdowns for variable selection, like `~` does.

morphic.js
cycomachead: Ah, yes, I agree that parent menus should always be visible. If we need to, we can do something clever like showing menus on the left is necessary. 

Actually...should RTL languages do that anyway? (Do we support RTL at all?)

morphic.js
brianharvey: Done.

morphic.js
brianharvey: Leaving this for now per phone call.

morphic.js
```diff
@@ -7818,10 +7824,40 @@ MenuMorph.prototype.select = function (aMenuItem) {
     this.selection = aMenuItem;
 };
 
+MenuMorph.prototype.mouseLeave = function () {
+    if (this.parent instanceof MenuMorph) {
+        // this is a submenu so we vanish when lose focus
+		this.destroy();
+    }
+}
```
brianharvey: Oh fine.  I point out that JSHint gives about 100 complaints, of which about five were mine.  A certain amount of the pot calling the kettle African-American going on here. :(

morphic.js
brianharvey: I'm pretty sure this is all necessary.  At least, everything I added was a response to something not working; I guess I can't promise for sure that this was the best place to solve the problems.

morphic.js
```diff
@@ -7818,10 +7824,40 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
+            
+
 MenuMorph.prototype.destroy = function () {
+var myself = this,
```
brianharvey: Yeah JSHint told me.  :)

morphic.js
brianharvey: Can't we just start over and rewrite Snap<i>!</i> in Scheme?  When you guys can't agree about the correct syntax there's something wrong with the underlying syntax spec.

morphic.js
```diff
@@ -7818,10 +7824,40 @@ MenuMorph.prototype.select = function (aMenuItem) {
...
         this.world.keyboardReceiver = null;
     }
+
+    if ((this.parent instanceof MenuMorph) && this.world) {
+        // If I'm a submenu, un-highlight the item pointing to me.
+		subitems(this.parent).forEach(function (item) {
+            item.image = item.normalImage;
+            item.changed();
+        } );
+    };
```
brianharvey: yeah got it

morphic.js
brianharvey: As it turns out the variable was used only once, so I flushed it altogether.  Some other time I'll discuss with you why one-letter names other than x are perfectly fine when correctly used. :)

morphic.js
brianharvey: Maybe later.  (I got a late start today -- someone kept me up until 4am -- so I don't want to start breaking something that's used in a zillion places.)

morphic.js
brianharvey: Nope, not here.

morphic.js
brianharvey: See above.

blocks.js
brianharvey: It's a long, sad story.  I meant to type C-J to start a new line properly indented, but instead hit M-J because my hand slipped, so all the spacing of this "paragraph" was lost and I had to recreate it.  It's better now.

blocks.js
brianharvey: I have used braces as agreed last night, even though I really agree with Michael.  :/

blocks.js
```diff
@@ -7553,13 +7554,41 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
+				// (not to be confused with "black right-pointing TRIANGLE")
+                menu.addItem(key, choices[key]);
+                menuStack.push(menu);
+                menu = new MenuMorph(
+                                     this.setContents,
+                                     null,
+                                     this,
+                                     this.fontSize
+                                 );
+            } else if (key.charCodeAt(0) == 0x25c4) {
```
brianharvey: Ditto.

blocks.js
```diff
@@ -7553,13 +7554,41 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
+				// "BLACK LEFT-POINTING POINTER"
+                var oldmenu = menuStack[menuStack.length - 1],
```
brianharvey: Fine.  Picking my battles...  :)

blocks.js
```diff
@@ -7553,13 +7554,41 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
+                oldmenu.addChild(menu);
+                menu = menuStack.pop();
             } else {
                 menu.addItem(key, choices[key]);
             }
         }
     }
+
+    while (menuStack.length > 0) {
+        var oldmenu=menuStack[menuStack.length - 1],
```
brianharvey: ok

brianharvey:
Okay, hereby requesting pull.

blocks.js
```diff
@@ -7554,24 +7556,26 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
         if (Object.prototype.hasOwnProperty.call(choices, key)) {
             if (key[0] === '~') {
                 menu.addLine();
-        // } else if (key.indexOf('§_def') === 0) { 
-        // menu.addItem(choices[key].blockInstance(), choices[key]);
-            } else if (key.charCodeAt(key.length - 1) == 0x25ba) {
+         // } else if (key.indexOf('§_def') === 0) { 
+         //     menu.addItem(choices[key].blockInstance(), choices[key]);
+            } else if (key.charAt(key.length - 1) == '{') {
```
cycomachead: Aside from the rest of JSHinting, this should just be an index, but I don't think JSHint can catch that.

byob.js
```diff
@@ -292,11 +292,17 @@ CustomBlockDefinition.prototype.inputOptionsOfIdx = function (idx) {
 };
 
 CustomBlockDefinition.prototype.dropDownMenuOf = function (inputName) {
-    var dict = {};
+    var dict = {},
+	    gensym = 1;
```
cycomachead: This should be a real word. No need to save characters. :-) In this case I'm not sure what it means.

(This is one habit that's been drilled into me the past few moths, a bit painfully, but I think it definitely ends up helping out.)

Otherwise, this little algorithm could use a comment a why/when we need to check of uniqueness.

byob.js
```diff
@@ -292,11 +292,17 @@ CustomBlockDefinition.prototype.inputOptionsOfIdx = function (idx) {
...
     if (this.declarations[inputName] && this.declarations[inputName][2]) {
         this.declarations[inputName][2].split('\n').forEach(function (line) {
-            var pair = line.split('=');
-            dict[pair[0]] = isNil(pair[1]) ? pair[0] : pair[1];
+            var pair = line.split('='),
+			    unique = pair[0];
+			if (unique == '~' || unique == '}') {
+				unique = unique.concat(gensym.toString());
```
cycomachead: Unless @jmoenig disagrees with this, I think this line can be simplified `unique += gensym` does exactly the same thing, but it might be better to be a little more explicit. At the very least `.toString()` is unnecessary.
```js
 > a = '~'
'~'
> a += 1
'~1'
> 
```
(For all the quirks of JS coercion, a string + an int is a string, is a consistent rule.)

blocks.js
```diff
@@ -7553,13 +7556,43 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
-            // } else if (key.indexOf('§_def') === 0) {
-            //     menu.addItem(choices[key].blockInstance(), choices[key]);
+         // } else if (key.indexOf('§_def') === 0) { 
+         //     menu.addItem(choices[key].blockInstance(), choices[key]);
+            } else if (key.charAt(key.length - 1) == '{') {
+                // Submenu
+                // 0x25ba = Unicode "BLACK RIGHT-POINTING POINTER"
+                // (not to be confused with "black right-pointing TRIANGLE")
+                menu.addItem(key.substring(0,key.length - 1).
+							 concat(String.fromCharCode(0x25ba)),
```
cycomachead: Also, why use Morphic's `concat` and not just `String::concat` or `+`.  Also I think you should be able to simply write `'\u25BA' ` instead of calling fromCharCode. (And the `\u` is a pretty clear unicode indicator.

Would it make sense to call localize around this character? I don't know if any other language would want it to be different?

blocks.js
```diff
@@ -7553,13 +7556,43 @@ InputSlotMorph.prototype.dropDownMenu = function (enableKeyboard) {
...
+							 choices[key]);
+                menuStack.push(menu);
+                menu = new MenuMorph(
+                                     this.setContents,
+                                     null,
+                                     this,
+                                     this.fontSize
+                                 );
+            } else if (key.charAt(0) == '}') {
+                // "BLACK LEFT-POINTING POINTER"
```
cycomachead: This comment doesn't belong here?

blocks.js
brianharvey: Oh right.  It was all harder back when I was using untypable characters.  I'll clean that up.

As for the many JSHint complaints, as I said earlier, maybe 2% of them were mine; the rest are (presumably) Jens's code.

blocks.js
cycomachead: Also can we _please_ add some space between the last character and the pointer. Any amount will do, but it looks that we're likely to see menus that look like this, and I'm just going to be driven mad when I notice it. (and users shouldn't have to worry about proper spacing.
```js
> 'a' + '\u25ba'
'a►'
```

byob.js
brianharvey: ok

blocks.js
brianharvey: I guess.  I hate + for concatenation, especially in a language that allows looseness about strings of digits vs. numbers.

PS Does 5*'a' give you 'aaaaa'?

blocks.js
brianharvey: Yeah, left over.  I kinda left it there to remind myself where we changed those characters just in case we decide to change back.

blocks.js
cycomachead: I see a lot in the new code that was written, I can point them about, but maybe I should just setup an appropriate tool to do it. @snap-bot will be happy to comment when he's ready. 😉 

(Also the unicode character you're using can totally be embedded in our file, but it's probably safer to actually refer to the value, rather than use the actual glyph.)

blocks.js
cycomachead: Nope, though, it really should... 
```js
> 5 * '1'
5
> 5 * 'a'
NaN
```

cycomachead: I guess the concat isn't so bad after I realized what it was applying to. `.concat(' \u25BA')` would be pretty clear.

brianharvey:
done, again.

blocks.js
brianharvey: Too late, I changed it to +.

About the spacing, on the list for later is to right-justify the arrow.

cycomachead: Yeah, I don't have strong feelings about this one. :) (I think + is nice to read, but concat is definitely more type safe. 

> About the spacing, on the list for later is to right-justify the arrow.

oh this will be excellent. Then I'll go through and add keyboard shortcut icons. 😁

morphic.js
```diff
@@ -7519,6 +7519,10 @@ MenuMorph.prototype.addLine = function (width) {
     this.items.push([0, width || 1]);
 };
 
+MenuMorph.prototype.addSubmenu = function (key, submenu) {
+	this.addItem(key, key, null, null, null, null, null, submenu);
```
cycomachead: LOL. But also cringe. this for @jmoenig:
 But I'm using this as a plug to eventually have an API that uses named parameters / objects instead of just positional arguments! Making anything that can be null / optional a key would greatly improve readability and learnability of morphic. :)

morphic.js
```diff
@@ -9403,16 +9443,33 @@ MenuItemMorph.prototype.mouseEnter = function () {
         this.image = this.highlightImage;
         this.changed();
     }
+    if (this.doubleClickAction instanceof MenuMorph) {
+        if (!contains(this.parent.children, this.doubleClickAction)) {
+            this.parent.addChild(this.doubleClickAction);
+		}
+        this.doubleClickAction.popup(this.world(),
+                                     new Point(this.parent.bounds.right()-10,
```
cycomachead: How were 10 and 4 picked? Are they what looks good, or are they also defined elsewhere?

brianharvey: Looks good.

morphic.js
cycomachead: well the resulting source needs to be in JS, but if we have a compiler then that seems fine. :)

morphic.js
liam4: >an API that uses parameters/objects instead of just positional arguments

ES6, of course:

```js
function myComplicatedFunction({
  foo = 'default foo',
  bar = 123,
  kaz = 3 * 7
} = {}) {
 console.log(foo, bar, kaz)
}

myComplicatedFunction()
myComplicatedFunction({foo: 75})
```

brianharvey:
On 12/30/16 4:16 PM, Michael Ball wrote:
> well the resulting source needs to be in JS, but if we have a compiler then that seems fine.

I don't understand this comment at all. :(

brianharvey:
@jmoenig You were going to do keyboard support for submenus but it was really easy so I did it!  Right arrow on a MenuItem that points to a submenu enters the submenu.  ESC while in the submenu kills it and returns focus to its parent.

jmoenig:
Excellent, thanks!

jmoenig:
(I've got a generalized shortcut-alignment/layout mechanism for menus working in my dev version, working that also supports your submenu arrow-indicators by right-alignim them and even support icons)

liam4:
>> well the resulting source needs to be in JS, but if we have a compiler then that seems fine.
>
>I don't understand this comment at all. :(

I think he was just saying that Snap<em>!</em> still needs to be able to run in a web browser, so you'd need to use a program to turn your new Scheme code into JavaScript code (since browsers can't run Scheme code).

brianharvey:
@liam4 Oh, I was proposing that we rewrite all the world's browsers to use Scheme instead of JS.  Or I suppose we could just write a Scheme interpreter in JS; that'd be easier.

brianharvey:
@jmoenig Awesome!  Is there anything else I promised the other night that I haven't done?

morphic.js
cycomachead: ES6 would be cool, but we could do all that I wanted with just type checking arguments in plain es5. :)

Oooh I just called es5 plain. The web is so much nicer than it used to be!

brianharvey: > The web is so much nicer than it used to be!

Yeah, maybe in another century or two it'll get to be as nice as the pre-Web Internet.

bromagosa:
@brianharvey [enjoy](http://www.biwascheme.org/)! :)

brianharvey:
@bromagosa Awesome!

jmoenig:
Thanks! I'm working on a bunch of extensions (support for right-aligned shortcuts and submenu indicators, only popping up submenus if the mouse actually hovers over an item (similar to tool tips), keeping submenus open when the mouse leaves the menu and more consistent keyboard control. I'll pull this togehter with those extensions later.

PR #1591, opened 2016-12-30
Preview Libraries Before Importing

cycomachead:
So, the libraries menu is getting a bit untenable for me, and the names are no longer helpful to what the libraries do / contain.

This creates a dialog that lets you preview a library before importing.

http://cycomachead.github.io/snap/ to play with it.

Screenshots:
screen shot 2016-12-30 at 12 13 31 am
This isn't how things should be, but I'm not sure I understand why...
screen shot 2016-12-30 at 12 37 24 am

screen shot 2016-12-30 at 12 37 52 am


gui.js
```diff
@@ -6265,6 +6229,341 @@ ProjectDialogMorph.prototype.fixLayout = function () {
...
+    SpriteMorph.prototype.categories.forEach(function (category) {
+        blocksList.forEach(function (definition, idx) {
+            if (definition.category !== category) {return; }
+            if (category !== previousCategory) {
+                y += padding;
+            }
+            
+            previousCategory = category;
+            blockImage = definition.templateInstance().fullImage();
+            blockContainer = new Morph();
```
cycomachead: Something around this line is a problem and I don't know why.

cycomachead: Previous attempt without setting the width and height looked like this:
screen shot 2016-12-29 at 11 38 17 pm

gui.js
```diff
@@ -6265,6 +6229,341 @@ ProjectDialogMorph.prototype.fixLayout = function () {
...
+
+    // Create a copy of the array; we'll remove blocks as we draw them.
+    blocksList = this.cachedBlocks(libraryKey).slice(0);
+
+    if (!blocksList.length) {return; }
+    // populate palette, grouped by categories.
+    this.initializePalette();
+    x = this.palette.left() + padding;
+    y = this.palette.top();
+    SpriteMorph.prototype.categories.forEach(function (category) {
```
bromagosa: Try to do it like this:

`    SpriteMorph.prototype.categories.forEach(function (category) {
        blocksList.forEach(function (definition, idx) {
            if (definition.category !== category) {return; }
            if (category !== previousCategory) {
                y += padding;
            }
            
            previousCategory = category;
            blockImage = definition.templateInstance().fullImage();
            blockContainer = new Morph();
            blockContainer.setExtent(
                new Point(blockImage.width, blockImage.height)
            );
            blockContainer.setPosition(new Point(x, y));
            myself.palette.add(blockContainer);

            blockContainer.image.getContext('2d').drawImage(
                blockImage,
                0,
                Math.round((blockContainer.height() - blockImage.height) / 2));

            y += blockContainer.fullBounds().height() + padding;
            blocksList.pop(idx);
        });
    });`

jmoenig:
Wow. This is a brilliant idea and really, really nice, Michael!

gui.js
cycomachead: Thanks! I'll check that out later today! Redrawing the whole image makes sense.

gui.js
```diff
@@ -6265,6 +6229,341 @@ ProjectDialogMorph.prototype.fixLayout = function () {
...
+                y += padding;
+            }
+            
+            previousCategory = category;
+            blockImage = definition.templateInstance().fullImage();
+            blockContainer = new Morph();
+            blockContainer.contents = blockImage;
+            blockContainer.image = blockImage;
+            blockContainer.setPosition(new Point(x, y));
+            blockContainer.setExtent(
```
Hardmath123: Ah, this has happened to me before. The solution was to set the extent before setting the image (i.e. move these three lines up before line 6477).
screen shot 2016-12-30 at 10 03 58 am

```
diff --git a/gui.js b/gui.js
index 140e2da..e9289ac 100644
--- a/gui.js
+++ b/gui.js
@@ -6474,12 +6474,12 @@ LibraryImportDialogMorph.prototype.displayBlocks = function (libraryKey) {
             previousCategory = category;
             blockImage = definition.templateInstance().fullImage();
             blockContainer = new Morph();
-            blockContainer.contents = blockImage;
-            blockContainer.image = blockImage;
-            blockContainer.setPosition(new Point(x, y));
             blockContainer.setExtent(
                 new Point(blockImage.width, blockImage.height)
             );
+            blockContainer.contents = blockImage;
+            blockContainer.image = blockImage;
+            blockContainer.setPosition(new Point(x, y));
             myself.palette.add(blockContainer);
             y += blockContainer.fullBounds().height() + padding;
             blocksList.pop(idx);
```

cycomachead:
Oh thanks hardmath!
That sorta makes sense I guess and I feel like that's the more "intended" method than directly calling drawImage. 

Sadly reordering my calls was not one of the permutations I tried but I'm glad it's an easy fix. :)
-- 
Michael Ball
From my iPhone
michaelball.co

brianharvey:
+1, I love this.  Does the image scroll if there are more blocks than fit?  Would be great to have this pulled asap.

brianharvey:
Here you go:

[LIBRARIES.txt](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/files/679456/LIBRARIES.txt)

gui.js
cycomachead: Thanks! This is the type of fix I was hoping to find. :)

More for curiosity, but do you (or @jmoenig) know why this is the case? I'm more curious than anything.

Hardmath123: Yes, @cycomachead, I believe it's because JS clears <canvas>es when you forcibly resize them. At least, that was the intuition that led me to this solution! :-)

cycomachead: I hadn't realized that, but it makes sense. Thanks!

cycomachead:
Ok, I just made a few minor fixes and have updated the LIBRARIES list. 

This should be ready to go / be reviewed. :)

jmoenig:
Thanks, Michael! It might take me a while to get around reviewing this pull request, but I'm planning to include this in the next release. Hang on!

cycomachead:
No problem!

I also just rewrote the layout code so that the blocks viewer is now elastic and will grow/shrink if the dialog is resized. I think the code is much better, but there's a minor thing I couldn't get right. (See code comments).

gui.js
```diff
@@ -6266,6 +6230,335 @@ ProjectDialogMorph.prototype.fixLayout = function () {
...
+        this.body.setExtent(new Point(
+            this.width() - this.padding * 2,
+            this.height()
+                - this.padding * 3 // top, bottom and button padding.
+                - titleHeight
+                - this.buttons.height()
+        ));
+
+        this.listField.setExtent(new Point(
+            200,
```
cycomachead: I wanted this to be `this.listField.width()` since it's already be seen (when the list field is created) but that didn't work out.

gui.js
```diff
@@ -6266,6 +6230,332 @@ ProjectDialogMorph.prototype.fixLayout = function () {
...
+            this.body.width() - this.listField.width() - thin,
+            100
+        ));
+        this.palette.setExtent(new Point(
+            this.notesField.width(),
+            this.body.height() - this.notesField.height() - thin
+        ));
+
+        this.listField.setPosition(this.body.position());
+        this.palette.setPosition(new Point(
```
cycomachead: It occurs to me that I could write these another way, which I almost prefer but don't see done in Snap<em>!</em>:

```js
        this.palette.setPosition(this.listField.topRight().add(new Point(thin, 0)));
        this.notesField.setPosition(this.pallete.bottomLeft().add(new Point(0, thin)));
```

It would be cool if there were `addX` and `addY` functions. :)

jmoenig: well, two observations: Your "other way" looks perfectly fine to me, I'm somewhat astonished that you didn't find any other occurrences of that style, because it's totally what I was thinking of at the time :-)

Yeah, there is the shortcut that you can just add a number (as opposed to a Point) to another Point, but this will add it to both x and y. I was imagining more of that back then...

PullJosh:
**Minor observation:** When entering extremely large numbers (e.g. 33333333333333333333333), the number will be placed within the block in sci notation (3.333333333333333e+22). While this isn't necesarily a bad thing, it *is* inconsistant with the behavior observed when entering numbers into a block directly (what you type is what you get.)

image

That being said, this will probably never *actually* matter.

BookOwl:
@PullJosh, what does that have to do with previewing libraries?

cycomachead:
I think it's just the wrong PR. :)

In any case, I don't know what we can fix that one b/c that's how JS displays numbers. Scientific notation _is_ was what users will see if they were to get that as the result of a math block in Snap<em>!</em>, so it's not totally unexpected.
(Well it can be fixed, but I don't know if it's worth it.)

And I should have posted this on the other PR...ooops.

PullJosh:
> @PullJosh, what does that have to do with previewing libraries?

Oops! Wrong PR, sorry. :P

BookOwl:
Figured. :P

jmoenig:
Michael, do you consider this pull-request ready for review and integration? I'm asking because I'm considering integrating it into the impending release - which shoujld happen end of next week. If you feel it's ready I'll spend the time reviewing and integrating it. If you'd rather keep working on it for some more time I'm fine with that, too. In this case it could go into the next release. But if you feel comfortable with it, I'd like to integrate it. What do you think?

jmoenig:
My other question is, did you test this in a "live" environment, perhaps in a some dev folder on s.b.e? Just curious as to how ready this is. Thanks!

cycomachead:
I always do my testing on a local server, unless I need cloud access. 

I think it's ready to be reviewed. 

Though actually I did notice that when I fixed scrolling the loading message no longer appears centered. I can fix that today, but I still think this should go in the next release. :)

-- 
Michael Ball
From my iPhone
michaelball.co

cycomachead:
Ok, minor tweaks made (including one typo I caught from 'simplifying' the positioning logic...).

Should be ready now! :)

jmoenig:
Thanks! I'll check it for inclusion in the upcoming release

jmoenig:
This look good to me, I'll pull it later today!

jmoenig:
Wow, this really came out beautiful and elegant, and I love how you integrated this complex feature seamlessly into Snap's UI Design. Thank you, @cycomachead !

cycomachead:
Thank you!

PR #1593, opened 2016-12-31
Leap

brianharvey:


brianharvey:
@jmoenig Just in case you weren't notified about it, I pulled changes to three libraries just now.  I hope that's okay...

jmoenig:
Please don't just pull stuff before we both agree on it. Especially while we're *that* close to a release and while we're working on rushing in new feature that you demand elsewhere. This *is* production software and you're treating it as pre-alpha-just-hack-around stuff. Thanks.

brianharvey:
Sorry, I thought (1) I was in charge of libraries, and (2) these are really minor changes.  You can revert if you want.

liam4:
Well, you should at least mention that you did also change some other things, as described by the commit titles! :)

PS, you can request someone to 'review' your pull request when you make it (in the same place you'd set a label). But I don't think that's terribly necessary unless the PR gets ignored for a week or so (Jens seems to be pretty quick at getting to issues!).

jmoenig:
@brianharvey yes (1) you are, and (2) these are :)
I'm just very nervous because you've set out to write libraries that change the Snap IDE, so I can no longer be sure...

brianharvey:
@liam4 Huh?  I changed three libraries.  And I pulled from master, which I guess wasn't strictly necessary but isn't a change.

liam4:
Well, yes, but the title only says "Leap" when you really also changed the setrgb and pen trails libraries.

brianharvey:
Yeah, well, if I knew how to use git I might have changed the title. :(

cycomachead:
The title can just be changed by the "Edit" button on GitHub. :)

Anyway PRs don't get deployed automatically and it's not like "master" is what people are currently using. Not that we couldn't have more of a review process but hey, this wasn't directly pushed to s.b.e as far as I can tell!

brianharvey:
No, this PR is about really minor improvements (e.g., replace a text input slot that only accepts a few choices with a menu).  I'll wait for the next release, which will presumably include it since I pulled it.  :)

PR #1483, opened 2016-10-29
Little Bits of Cleanup

cycomachead:
I collected these overtime - finally getting around to a PR.

FWIW, there are no functional changes here. All minor code things. There are two "big" ones:

* "SNAP!" turned into "Snap!" to reflect the decision sometime ago to not shout the name everywhere.
* `= function() { nop() };` to just `nop;` since they're equivalent. (Well, it's one less function call, but I don't think that matters.)

The others align code comments and clean up some misspellings in context menus.

PR #1596, opened 2017-01-02
Catalan updating

jguille2:
Sync translation strings from code (first in catalan translation... soon in a master template and in all translation files).
-First, I want to sync last changes in code (I saw someone in Michael's PR #1483)
-Then I saw these strings were "news" (they were not in translation files)
-So I've added all developers menu strings to translation files

Note ( @cycomachead  , @brianharvey ) I suggest you use 'git rebase' before merging complex branchs (when these branchs have resynchronized with master, or when they have reverted changes). If not, it's too difficult to trach progress. Thanks

### DIFF
--- a/lang-ca.js
+++ b/lang-ca.js
@@ -2083,5 +2083,37 @@ SnapTranslator.dict.ca = {
 	'development mode...':
 		'mode de desenvolupament',
 	'about morphic.js...':
-		'sobre morphic.js...'
+		'sobre morphic.js...',
+	//Make a Morph
+	'make a morph':
+		'crea un \'morph\'',
+	'rectangle':
+		'rectangle',
+	'box':
+		'caixa',
+	'circle box':
+		'caixa circular',
+	'frame':
+		'marc',
+	'scroll frame':
+		'marc amb desplaçament',
+	'handle':
+		'nansa',
+	'string':
+		'cadena',
+	'Hello, World!':
+		'Hola Món!',
+	'speech bubble':
+		'bafarada',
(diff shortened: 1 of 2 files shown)


cycomachead:
LGTM.

This should be an easy merge anyway since the changes to morphic are pretty isolated. :)

jguille2:
Hi Michael ( @cycomachead ),

Excuse my offtopic... (I think we don't need another issue for this comment) but I try to explain...

If you take a look at [rebase docs](https://git-scm.com/book/en/v2/Git-Branching-Rebasing), you will see the benefits of this feature. I suggested to use this in  local complex branches. This keeps a linear, logical history for our master branch.

The example:
 - If you see  PR #1483, you can inspect changes in this PR view.
 - But working with code, if you take a look at the [PR commit](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/commit/3c8eb522bfddbe31a0f069f099a8f71ea852aae0), you will see that these 'little bits' are 85 commits with 1012 changed files! Impossible to follow history in this view!
 - I know there are tools to see the branch (parallel) working... but code history shows a linear commits history... and then, local commits will appear mixed with 'origin master' commits and seems to make no sense. For exemple we see an SNAP! -> Snap! change in the linear history... but you know precedent commits have "Snap!"... because it's not linear and it's reverting an older commit.

``
So, my suggestion: when branches last a long time or when they have their own merges (to synch with master) or when they reverted their own changes  Then, it's better making a rebase before they merge into master.
``

Issue #1049, opened 2015-12-07
Support Renaming for Upvars, Formal Params, Script Vars


ddgarcia:
We want our kids to refactor their code as they mature, and one of the advantages Snap! has over text languages is that if you rename a block, it propagates to every instance of usage. Because it's easy and instantaneous, kids are almost encouraged to do it. That's great.

However, the same cannot be said of formal parameters, upvars and script variables.  I had a script variable 'n' used 10 times or so, and wanted to change its name to something better, which was "recursion level". It took me 10 minutes to drag the new name to all the expressions that used to have "n" in it, and I actually missed a few. There's no equivalent of emacs "search and replace", so we need the system to help us here.  In the case that there's already a variable of the same name already, it could just warn you to go ahead or cancel with the rename.  It's also important to remember to change the name in the "set/change _name_ to _value" blocks.

That we don't support this makes users not want to refactor these names, because the effort is so high.

cycomachead:
For reference:
#46 

I agree this is needed. :)

jmoenig:
So, this is essentially a tool request. I agree that more tools would be nice, e.g. show all references to a block or a variable, all implementors and callers of a function, message, event etc.

cycomachead:
No, I don't think it's a tool, I think the idea is that it would be _automatic_ that you don't even need a tool.

Like renaming a script var from "a" → "name" would just find all "a" in that scope and replace them with "name", just like renaming a block does.

jmoenig:
Yeah, see, that part I don't like at all, because it somehow conveys the misconception, that the block somehow *is* identical with the variable, whereas I personally use variables out of their scope all the time, often even creating variable references even *before* declaring the variable. The problem with doing more and more things automatically is that we're taking such possibilities away. That's why I claim it should be tools, if at all. Then users can choose for themselves if they actually want to rename all references to a variable or just this particular reference or definition. Same with delete etc. See, automatisms are all nice and well, except often enough it's not what users really want. So, by introducing ten-dollar words and - questionable and even downright ideologically burdened, I believe - concepts like "refactoring" (how does that even work without built-in unit testing?) we also need to impose all kinds of "best practices". And *that's* something that I personally abhor for students.

That said, tools would be nice... :-)

brianharvey:
[...]

Issue #49, opened 2013-04-16
"show definition" command for variables


nathan:
I propose that right-clicking on a variable reporter or setter block (CHANGE or SET) would present the user with a "show definition" command. If it refers to a variable that has not been declared in its lexical scope, the command would be hidden. When the command is selected, the following would occur:
- If the variable to which the block refers is a script-local variable, the SCRIPT VARIABLES block in which it is declared would be scrolled into view.
- If the variable to which the block refers is a formal parameter to a block, the block prototype would be scrolled into view.
- If the variable is a global or sprite-local variable, the variables palette would be shown and the variable reporter corresponding to the variable would be scrolled into view.

PR #1599, opened 2017-01-04
Variable refactorer

bromagosa:
In any variable declaration, be it a global one, sprite local one, script local one or upvar, a contextual menu lets you rename all instances of that var in its context.


bromagosa:
Hi Jens, feel free to squash these when pulling, in case you don't want to have 16 different commits for a single feature. Otherwise, I can create a new branch, squash these into a single commit, and send a new pull request from there. Your call! :)

blocks.js
```diff
@@ -3000,6 +3068,129 @@ BlockMorph.prototype.codeMappingHeader = function () {
...
+        stage = ide.stage,
+        oldWatcher = receiver.findVariableWatcher(oldName),
+        oldValue, newWatcher;
+
+    new DialogBoxMorph(
+            this,
+            function (newName) {
+                var definer;
+                
+                if (this.parent instanceof SyntaxElementMorph) {
```
cycomachead: Shouldn't this be `myself.`? Or am I misreading this?

cycomachead:
Tagging #49 #1049

jmoenig:
Thanks, Bernat! I don't think I mind all the commits, as long as I can browse the changes nicely (which I can).

blocks.js
bromagosa: I think it doesn't matter much, as `DialogBoxMorph`s execute their action in the context of the specified target:

    if (typeof this.target === 'function') {
        if (typeof this.action === 'function') {
            this.target.call(this.environment, this.action.call());
        } else {
            this.target.call(this.environment, this.action);
        }
    } else {
        if (typeof this.action === 'function') {
            this.action.call(this.target, this.getInput());
        } else { // assume it's a String
            this.target[this.action](this.getInput());
        }
    }

So, when you instantiate one with:

    new DialogBoxMorph(this, [...])

The target is going to be the context where it's been instantiated.

@jmoenig did I get that right?

jmoenig:
Ah, I forgot to test something earlier, and unfortunately it does turn out to be a problem: Inheritance. Inherited sprite-local variables seem to be handled well, but 

(1) they still allow me to rename them in the palette, and then fail for any operation that I try.

(2) Also, when I refactor the parent's / ancestor's sprite-local variable, the descendant's / child's references to the inherited var does not get refactored.

I think (1) could be addressed by simply disabling any renaming of inherited variables in the palette, and (2) probably fits into the "not too greedy" spec, in other words, it's okay, I think.

I think I can take care of (1) myself, so don't bother, Bernat!

jmoenig:
otherwise this looks really nice, I'm going to pull it now. Thanks!

bromagosa:
Oh that's right! I never thought to check inherited vars, sorry about that :(

jmoenig:
no problem, Bernat, this might actually be a good thing, because I'm still working on OOP. Don't worry!

brianharvey:
I wouldn't say it's a showstopper, pre-4.1, but I vote that it has to be fixed in the full glorious 4.1 release.

blocks.js
cycomachead: I thought it should, since we've been playing with this, but I didn't really understand why. :)

brianharvey:
I would argue that the right thing is to assume that the most likely case is that the user isn't going to change the inheritance, or at least that the user who _does_ change the inheritance is taking on the responsibility to check on the implications of doing so.  If you assume that the user _is_ going to change the inheritance, you are for sure breaking the inheritance relationship for this variable right now.

And, arguing that the user might later change the inheritance applies just as much to doing it using the gui!  So I don't think you can claim this as an argument against Eisenbergification.

jmoenig:
it's an argument against renaming references to potentially same-named sprite-local variables downstream the descendants-line

brianharvey:
Oh, I see.  So if the user does that interactively, I guess we have to warn the user and ask whether they intend the renaming to cause the children to inherit, or keep the child's variable independent.

And yes, that complicates doing it programmatically.  You could assume keep independent unless the block has an additional optional input True.

jmoenig:
I think there's a viable solution: Per default don't rename any references to potentially inherited variables, i.e. only either global for globals or per single sprite for sprite-locals. This is what Bernat's code does right now. In order to refactor a variable across an inheritance chain the user can edit each sprite and temporarily *shadow* the to-be-refactored inherited variable. *Then* it can be renamed, and afterwards it can be deleted again, *if* the new name corresponds to another inheritable slot.

brianharvey:
That's so complicated!  Especially if the sprite has 100 children (think Breakout), and then maybe there are grandchildren, etc.  This is the sort of task that computers can do much better than people can.

I think the measure of "minimality" or "non-greediness" has to be based on the user's likely intent.  So, if the child has a reference to the parent's variable, and we rename the parent's variable, saying to do it _everywhere,_ I claim the likely intent is to include references in the child.

The tricky case is renaming in the parent _to_ a name that already names an independent variable in the child.  I don't see _any_ plausible use case for this, and so I think it's most likely that the user has made a mistake, not realizing that the new name conflicts with something in a child.  And yet, it's still implausible that the user wants to make a change _only_ in the parent.  So I think we should just refuse to do this, or at least put up a warning and require confirmation.

Because, suppose we do change it _only_ in the parent.  What happens to the references to the old name in code in the child?  They are now references to nothing.  That can't be the user's intention!

DyslexicAwe:
I will describe a similar idea that is not exactly about renaming, though. It is about on-the-fly and on-the-spot creation of variables when user happens to change his mind and wants to replace some value in input slot of a block (for example, move _ block) with a variable, but sparing him the hassle of changing category tab from whichever category it is opened at the moment (for example, movement) to variables category and then pressing the button create a variable, and then setting it to a certain value and going back to the previous category tab, instead the idea is to allow for "variablification of values from input slots" as a friendlier and quicker contextualized user action. How? Well, the contextual menu when a user right-clicks on the input slot in the block would include the "variablify" option-a submenu in fact which lists all existing variables to choose from and an additional option - create a new variable the value of which will be "on-the-fly" set to the value found in the input slot. Here a picture instead of words (as one picture worths 1k words) 
variablify

brianharvey:
I'm not convinced about variablify for two reasons: (1) creeping featurism, for something that's not so hard to do now; (2) more important, this would encourage people _not_ to include a SET block in their initialization code.  And that's not fixable, because we wouldn't want to make guesses about where to put it.

DyslexicAwe:
Thank you for constructive criticism, Brian. I see I gave a wrong impression that laziness was my motivation but it was a pedagogical motivation after all, (my parents, both my mum and dad, were teachers and many afternoons/weekends I have to watch them carefully reading huge piles of tests when I wanted their attention to be on me as kid, my dad is now retired/mum dead for some time, I must be projecting this image on you, here, hm) teaching the _principle of abstraction_ by using variables instead of literal values in input slots. Having hopefully corrected the wrong impression, (although I am aware that first impression is very hard to correct) I will try showing there is a solution for the "unfixable" problem. I'd say that in 90% of all user cases, a SET block is placed in [continuous dichotomous](http://www.statisticshowto.com/dichotomous-variable/) distribution, thus reducing the "informed guesses" to the table below: 

option A | option B
------------ | -------------
at the top of the script| just before the block in question
(initialization placement)|  (just in time) 

Move it up or down the script as soon as you placed it in either of the two positions you can choose from as a user. Viola! You spared the user (a kid) what is (according to teachers) "not hard to do" but (according to my inner kid) _"changing category tab from whichever category it is opened at the moment (for example, movement) to variables category and then pressing the button 'create a variable', and then setting it to a certain value and going back to the previous category tab"_ will unnecessary take away some of the joy and beauty (pun intended) from the learning process.

brianharvey:
Still not convinced, but instead of hijacking this thread, why don't you move your message to a new issue and we can discuss it there.

DyslexicAwe:
> instead of hijacking this thread, why don't you move your message to a new issue and we can discuss it there

I give up on the idea, so no need to open it as a new issue.

PR #1605, opened 2017-01-06
Run flags

brianharvey:
Jens has been editing gui.js too, and I'm confused about what's
supposed to be there, so I downloaded the latest dev version and put my
edits there.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -327,8 +327,28 @@ IDE_Morph.prototype.openIn = function (world) {
         this.inform('Snap!', motd);
     }
     */
+
     function interpretUrlAnchors() {
         var dict;
+	function doDictFlags() {
+            if (dict.editMode) {
+                myself.toggleAppMode(false);
+            } else {
+                myself.toggleAppMode(true);
+            }
+
+            if (!dict.noRun) {
+                myself.runScripts();
+            }
+
+            if (dict.hideControls) {
+                myself.controlBar.hide();
+                window.onbeforeunload = nop;
+            }
+            if (dict.noExitWarning) {
+                window.onbeforeunload = nop;
+            }
+	};
         if (location.hash.substr(0, 6) === '#open:') {
             hash = location.hash.substr(6);
             if (hash.charAt(0) === '%'
@@ -349,6 +369,7 @@ IDE_Morph.prototype.openIn = function (world) {
             }
         } else if (location.hash.substr(0, 5) === '#run:') {
             hash = location.hash.substr(5);
+	    hash = hash.slice(0,hash.indexOf("&"));
             if (hash.charAt(0) === '%'
                     || hash.search(/\%(?:[0-9a-f]{2})/i) > -1) {
                 hash = decodeURIComponent(hash);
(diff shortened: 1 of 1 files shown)


brianharvey:
@jmoenig Here's my PR.  I did test it, but not with projects directly in the URL, because I don't know how to make one of those.  Do we still do that?

I hope this is mergeable with the correct version of yours...

cycomachead:
> @jmoenig Here's my PR. I did test it, but not with projects directly in the URL, because I don't know how to make one of those. Do we still do that?

There's a hidden setting by shift clicking. If by "we" you mean Jens, then maybe. Otherwise, I don't think so! :)

cycomachead:
> so I downloaded the latest dev version and put my
> edits there.

It shouldn't matter what Jens is doing. You can just work off your branch / fork from Github. 
And you should be able to keep things up to date with just `git merge master`

jmoenig:
Overall this looks okay to me. What's weird is that your pull-request identifies a bunch of additions that I've already committed earlier, so the diffing against the current version on Github is somewhow off, it's not clear to me, which changes are yours and which were already there, or rather why some seems to be made *again* and are marked as changes, even though they must already be part of the current version on Github. This is confusing and undermines my trust in Github...

gui.js
```diff
@@ -329,6 +329,26 @@ IDE_Morph.prototype.openIn = function (world) {
...
+                myself.toggleAppMode(true);
+            }
+
+            if (!dict.noRun) {
+                myself.runScripts();
+            }
+
+            if (dict.hideControls) {
+                myself.controlBar.hide();
+                window.onbeforeunload = nop;
```
jmoenig: this is one of the changes I've made some time ago. I don't understand why it appears in this chageset. Something is wrong about the version diffing here. I share your feeling that it should be just ```null``` or even ```delete window.onbeforeunload;```, but - strangely - when I tried those I got errors.

cycomachead:
The diff is because I think Brian copied from s.b.e and not from an existing branch so it looks like stuff was undone. 

Otherwise some of the existing stuff is different because it's been reordered. 

-- 
Michael Ball
From my iPhone
michaelball.co

gui.js
brianharvey: I just cut-and-pasted this code from the #present branch into a separate procedure so I could use it in two places.  I didn't change any details.

brianharvey:
Oh!  Yes at one point in debugging I got to an unrecoverable mess (putting band-aids on bugs just as I always tell students not to do) so copied from s.b.e/dev which I thought was current with the repo.  Sorry about that.  It should be pullable without problem though.

jmoenig:
Hmm... I think I can see all the relevant changes, so ... yeah ... I guess I can pull it and then check line by line myself. You know, this short before the release, it would really help if you could somehow issue a "clean" pull-request, so that in case a glitch surfaces later (which has been known to happen in the past, haha) the changes are well-documented.
Can you do that? If not, let me know and I'll pull it anyway...
Thanks!

jmoenig:
Okay, screw it, I'm pulling this anyway. Next time please issue a proper and clean pull-request.

jmoenig:
... and for god's sake please lint your code!

jmoenig:
Oh, man, this is awful. I'm not talking about spaces vs. tabs or double-nested functions instead of using a parameter. Please, please, pretty please check our variables and their scopes! ```idx``` is a *global* variable in your code because you're not declaring it using a ```var``` statement. You know, this is where *any* linting tool will help. No use complaining about JavaScript's stupidities, but this is precisely why I'm not keen on code contributions that claim urgency over diligence or even respect for Snap's code.

I'm trying to fix the worst things myself now. But please ... I was planning on other things this week, like, some serious testing etc., not this!

brianharvey:
Sorry, sorry!  That was very bad of me.

PR #1609, opened 2017-01-08
Fixes #1607 - fix var refactor for rings

### LINKED ISSUE #1607: "rename all" doesn't penetrate into rings
This issue relates to the handy new "rename all..." feature on variables, introduced from pull request #1599 "Variable refactorer".

This issue appears in the latest master (as of ad9c4fe69c), and also in the original pull at 5117cd4a40

## To reproduce:

1. Create a new global variable, "foo"
1. Add a couple blocks to the editing area: a "set" block of foo, and a "run" block on a "set" block of foo within a ring: 
image
1. Right-click on the "foo" global, and choose "rename all...", rename to bar.

## Expected results:
All instances of the global variable "foo", including those found within the ring, should be renamed.

## Actual Results:
The instance of "foo" inside the ring is left alone and not renamed: 
image

micahcowan:
I _think_ this is all that's necessary to fix #1607?

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -668,7 +668,7 @@ SyntaxElementMorph.prototype.refactorVarInStack = function (oldName, newName, is
     // Rename all oldName var occurrences found in this block stack into newName
     // taking care of not being too greedy
 
-    if (this instanceof RingMorph
+    if ((this instanceof RingMorph && contains(this.inputNames(), oldName))
             || (!isScriptVar && this.definesScriptVariable(oldName))) {
         return;
     }


jmoenig:
yes, at first sight this looks good to me. How about you, @bromagosa ?

bromagosa:
Yep, I don't know why I didn't add this clause, but I swear I thought it was there :sweat_smile:

Thanks!

jmoenig:
Oh! Maybe it was me who broke it when I rearranged your code, @bromagosa ... Anyway, I'll pull this fix later, thanks, @micahcowan !

Issue #1594, opened 2017-01-01
Switch media list files to JSON


cycomachead:
While JSON has a little more syntax the current file format, it would be quicker to parse (native parser) and we wouldn't have to worry about any edge cases that come up if we want to have descriptions that contain newlines and other things.

PullJosh:
JSON is also easier to read, for what it's worth.

jmoenig:
Can you tell me which problem this solves? Is there an issue with newlines in our XML format? If so, we should fix it, right?

cycomachead:
I'm talking about the LIBRARIES / COSTUMES / SOUNDS etc files. The current files are a really loose new line and tab delimited format. It was better than scraping and parsing the HTML response pages, but there's definitely been a couple cases where a line doesn't correctly contain a tab character, and now BH gave me verbose descriptions for the new libraries, and those could be more nicely formatted.

(Of course, I could (in theory) adapt the 'parser' for the current format to handle more newlines or escapes, but it seems simpler and less error prone to use JSON.)

jmoenig:
Ah, now I see. Thanks!

BookOwl:
I created a PR for this in #1621

PR #1621, opened 2017-01-16
Switch media list files to JSON

BookOwl:
I have converted the media list files to JSON as requested in issue #1594 by @cycomachead. This greatly simplifies `IDE_Morph.prototype.parseResourceFile`, while at the same time making it more robust and probably faster.


cycomachead:
Why make these be under `resources`?

It seems better to simply have each file contain an array. 

Also, I wasn't sure whether it was best to duplicate the keys every time or just store items as arrays.

BookOwl:
For some reason I thought that arrays weren't allowed to be the top level thing in JSON, I'll change it.

BookOwl:
@jmoenig?

jmoenig:
I don't have any strong opinions about this. @cycomachead ? @brianharvey ?

Oh, wait, maybe I do, sorta.

I don't think @cycomachead's little parsing function for his current format is too complicated at all, it's basically very straightforward and just a very few lines of code. OTOH I don't much mind JSON, either. The part I don't like so much about JSON is that it requires the name of each field in every instance of that field. That way data files become exponentially bigger the more data they hold. JSON is cool for heterogeneous data, but for a table structure in which every record has the exact same structure I've never understood why people love JSON. Also, all this syntax in the data file increases the risk for bugs. I guess people love JSON because ... they love JSON, not because there's an actual benefit in the case of a table.

Is that too harsh?

brianharvey:
No strong opinions.  Michael's #1594 argues for this on robustness grounds.  Whatever.

cycomachead:
For the JSON structure, we don't need to duplicate the property names. 

We can use do:
```json
[ 
    [ 'fileName', 'title', 'description'],
    ...
]
```
And then map over arrays to add properties. And we can even make the  `description` optional for fields that don't have one.

Either way, I still support JSON because there's been a number of times that LIBRARIES has been edited to have spaces instead of tabs delimiting something.

jmoenig:
this is superseded by Eric Hakenholz's other pull request, so we'll close this one. Thanks!

PR #1625, opened 2017-01-18
Improve library loading message

cycomachead:
So it now says "loading preview [\n] [library name]"


jmoenig:
Nice, thanks!
I'll pull this along with a few other little fixes I'm working on...

PR #1628, opened 2017-01-19
fix typo in cachedInputs

cycomachead:
This shouldn't affect anything, but now there won't be an extra (unused) property lying around.


jmoenig:
Haha! Who knew?
Great catch, thank you!

PR #1629, opened 2017-01-19
Error Handling improvements

cycomachead:
This turns on the assertions for list elements. Cue discussion on that...

Otherwise there are a couple other things:
* I have sane defaults for _not_ raising errors for the list primitives when no data is provided. This is the most "Scratch-like" IMO, and no-ops for the commands make sense because if you have null data it seems like you clicked the block rather than are trying to use it in a script. 
* improves the error display if you have a list of types by joining with ' or '
* Adds a new error `SnapError`. This isn't used yet, but my intention is for us to `throw new SnapError` instead of `throw new Error` which will also showing a message to distinguish between messages we write and JS errors, which we can preface with more verbose text. 
* has the split reporter use the assert method, instead of its own implementation.

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -1575,26 +1575,26 @@ Process.prototype.reportNewList = function (elements) {
 };
 
 Process.prototype.reportCONS = function (car, cdr) {
-    // this.assertType(cdr, 'list');
+    this.assertType(cdr, 'list');
     return new List().cons(car, cdr);
 };
 
 Process.prototype.reportCDR = function (list) {
     if (!list) {return; }
-    // this.assertType(list, 'list');
+    this.assertType(list, 'list');
     return list.cdr();
 };
 
 Process.prototype.doAddToList = function (element, list) {
     if (!list) {return; }
-    // this.assertType(list, 'list');
+    this.assertType(list, 'list');
     list.add(element);
 };
 
 Process.prototype.doDeleteFromList = function (index, list) {
     var idx = index;
     if (index === '' || !list) {return; }
-    // this.assertType(list, 'list');
+    this.assertType(list, 'list');
     if (this.inputOption(index) === 'all') {
         return list.clear();
     }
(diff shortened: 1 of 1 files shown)


threads.js
```diff
@@ -171,6 +171,21 @@ function invoke(
     return proc.homeContext.inputs[0];
 }
 
+// SnapError ///////////////////////////////////////////////////////////
+/*
+    A SnapError is a wrapper around a standard JS Error message.
```
jmoenig: good idea, I like it!

jmoenig:
this is okay, I guess. Let me think about the ramifications and when it would be best to active this...

cycomachead:
I'm not totally convinced about the type assertions. That's the only thing that I think might be worth not doing. 

I also have a plan and a (I think) a decent way for only doing type checking _after_ an error has occurred that will help us present much nicer messages to users overall. I'll hopefully make a PR with that in the next day or so assuming I get some bugs worked out. The gist though is that we can use the input slots in most cases to tell us the types we are interested in. 

-- 
Michael Ball
From my iPhone
michaelball.co

jmoenig:
I think some recent changes in the dev branch have taken care of the most pressing type-error and error-message issues here. I'm going to leave this "under review" while we're working on integrating the other stuff first.

brianharvey:
As a contribution to that review:  The other week I saw a Piazza message from a teacher who'd gotten the error message "TypeError big.to.LowerCase is not a function."  I was thinking this would be a counterexample to your claim that most type errors were under control, but when I looked it up, it turned out that the source of this error (which came from using a list as input to a block that expects text strings, so backwards from the usual case) was a JSFunction custom block I had written for use in BJC.  So, I should have written my block to check the input types.

I wonder whether we want to make it easier for authors of JSFunction blocks by providing automatic type checking at the JS interface if the inputs were typed in the prototype?  Probably not...  I guess I should just add to the JSFunction documentation that it's the user's responsibility to do any needed type checking.

jmoenig:
Good idea, @brianharvey , to make it easy for authors of JS-functions to assert expected input types. We already have it in

```Process.prototype.assertType = function (thing, typeString)```

authors need only to call it. Remember that the last input to any JS-function is always the current process.

cycomachead:
> I wonder whether we want to make it easier for authors of JSFunction blocks by providing automatic type checking at the JS interface if the inputs were typed in the prototype? Probably not... I guess I should just add to the JSFunction documentation that it's the user's responsibility to do any needed type checking.

I think #1744 mostly accomplishes this by automatically doing a check when an error occurs.

brianharvey:
Ah, groovy.

cycomachead:
Closing because we handle list errors, and finishing up #1744 will cover the other changes.

Issue #1634, opened 2017-01-21
Bug? Set list2 to list1 makes no copy but makes list2 same object als list1


Costor:
Seems that a bug in list assignment has been introduced some hours ago:  
At least code that worked fine until yesterday does no longer work:

Setting list2 to list1 does not copy contents of list1 into list2, but list2 becomes alias of list1.  Everything done to list2 is done to list1. 

See example (Press green flag, press space or 'a' key):

http://snap.berkeley.edu/snapsource/snap.html#present:Username=costor&ProjectName=SetListBug

Or is it a new feature (i.e. object assignment of lists?), and we need to explicitly assign copies of list1?

Kind regards, Costor

jmoenig:
It's Math, not a bug.

set POTUS to @realDonaldTrump assigns the actual guy to the presidency, not a copy of him. 

I wish Americans had known this. Really.

Von meinem iPhone gesendet

dekrain:
![memory](https://cloud.githubusercontent.com/assets/19158291/22177118/5abd474c-e018-11e6-998e-7beba830aeda.png)

Costor:
Sorry, I hadn't checked this thoroughly beforehand. I had realized that < all but the first of list > is copying instead of giving a reference to the rest of the list, and also there is no < copy list > primitive, so I had come to think that assignment would also be copying. Treating lists/arrays like atomic variables on asigment (i.e. copying) also seemed natural in a language for kids who have no notion of a pointer.

brianharvey:
All but first of doesn't copy if the list is a linked list.  And if it's an array, the copy is linked, so that the _next_ all but first of doesn't have to copy.

The object of the exercise is that if, for a given list, you are consistent in using it functionally or imperatively, in at most one copy it becomes linked or array, respectively, and then everything happens in constant time.

micahcowan:
So... personally, I find this very dismaying. Differing performance is one thing, but differing behavior, especially due to reasons that may not be under the author's control (e.g., a blocks-library author has no control over how the list passed to it was used prior to passing) is very surprising behavior for a language. At least when, in both cases, they look, taste and smell the same as far as any standard blocks are concerned. There's no (builtin) block for "is this a linked list or an array?", and if there were (or one used a non-builtin version), having to use it would feel very much like a kludge, and a weakness in the language.

[...]

PR #1643, opened 2017-01-27
Fix for #1634: convert to linked list in-place.

micahcowan:
Proposed fix for #1634. With correct author this time. :)

### DIFF
--- a/lists.js
+++ b/lists.js
@@ -143,18 +143,14 @@ List.prototype.cons = function (car, cdr) {
 
 List.prototype.cdr = function () {
     var result, i;
-    if (this.isLinked) {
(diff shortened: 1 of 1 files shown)


jmoenig:
Sorry, I'm not convinced. If you use ```all but first``` an an arrayed list I think it's alright to expect you want a copy of the list without the first element. That's how Smalltalk does it, and thus it must be right :-)

The whole issue is overrated, imo. You're right about libraries. Libraries should make sure to distinguish between linked and arrayed lists, and it's very easy to check this, and I'm sure there must a library that has such a block already. The list primitive for MAP that are already in the source code (and hidden in dev mode) already does this.

PR #1646, opened 2017-01-30
Add [scratchblocks] export

tjvr:
Good morning! :)

This is a small PR to add a "stringify..." option to the menu. This opens a dialog box containing [scratchblocks](https://wiki.scratch.mit.edu/wiki/Block_Plugin/Syntax) code, which can be copy/pasted into the [Scratch Forums](https://scratch.mit.edu/discuss/topic/219464/). (You can also try the [test page](https://scratchblocks.github.io), but since Snap! already has a "script pic" feature, the test page is less useful!)

Uploading pictures to the Snap! forums is quite time-consuming. Copy/pasting scratchblocks code is quicker, and might help to facilitate discussion on the User thread.

I've partly done this as an exercise in writing code in a Jens-approved style; please let me know what you think @jmoenig! :-)

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -2547,7 +2547,7 @@ BlockMorph.prototype.userMenu = function () {
         'userDestroy'
     );
     menu.addItem(
-      'stringify...',
+      'forum code...',
         function() {
             var code = myself.stringify();
             window.prompt('scratchblocks code for you to copy and paste', code);
--- a/blocks.js
+++ b/blocks.js
@@ -1852,6 +1852,34 @@ SyntaxElementMorph.prototype.isEmptySlot = function () {
     return false;
 };
 
+// SyntaxElementMorph to "scratchblocks"
+
+SyntaxElementMorph.prototype.stringify = function () {
+    var nb = this.nextBlock && this.nextBlock(),
+        result;
+    result = this.parts().map(function(child) {
+        if (child.stringify) {
+          return child.stringify();
+        } else if (child instanceof StringMorph) {
+          return child.text;
+        } else {
+          return ''; // should never happen
+        }
+    }).join(' ') + this.stringifyCategory();
+    if (nb) {
+        result += '\n' + nb.stringify()
+    }
+    return result;
+};
+
+SyntaxElementMorph.prototype.stringifyCategory = function () {
+    // private. answers with scratchblocks category specifier
+    return ' :: ' + ({
+        'lists': 'list',
+        'other': 'grey',
+    }[this.category] || this.category);
+};
+
 // SyntaxElementMorph speech bubble feedback:
 
 SyntaxElementMorph.prototype.showBubble = function (value, exportPic) {
@@ -2516,6 +2544,14 @@ BlockMorph.prototype.userMenu = function () {
     menu.addItem(
         "delete",
         'userDestroy'
+    );
+    menu.addItem(
+      'stringify...',
+        function() {
+            var code = myself.stringify();
+            window.prompt('scratchblocks code for you to copy and paste', code);
+        },
+      'generate `scratchblocks` code for the Scratch Forums'
     );
     menu.addItem(
         "script pic...",
@@ -5057,6 +5093,17 @@ ReporterBlockMorph.prototype.determineSlotSpec = function () {
     return null;
 };
 
+// ReporterBlockMorph to "scratchblocks"
+
+ReporterBlockMorph.prototype.stringify = function () {
+    var inner = ReporterBlockMorph.uber.stringify.call(this);
+    if (this.isPredicate) {
+        return '<' + inner + '>';
+    } else {
+        return '(' + inner + ')';
+    }
+};
+
 // ReporterBlockMorph events
 
 ReporterBlockMorph.prototype.mouseClickLeft = function (pos) {
@@ -5617,6 +5664,14 @@ RingMorph.prototype.fixBlockColor = function (nearest, isForced) {
     RingMorph.uber.fixBlockColor.call(this, nearest, isForced);
     slot.fixLayout();
 };
+
+// RingMorph to "scratchblocks"
+
+RingMorph.prototype.stringifyCategory = function () {
+    // force rendering as a 'ring' shape.
+    return RingMorph.uber.stringifyCategory.call(this) + ' ring';
+};
+
 
 // ScriptsMorph ////////////////////////////////////////////////////////
 
@@ -6673,6 +6728,18 @@ ArgMorph.prototype.isEmptySlot = function () {
     return this.type !== null;
 };
 
+// ArgMorph to "scratchblocks"
+
+ArgMorph.prototype.stringify = function () {
+    if (this.isHole) { // empty hole in ring
+        return this.isPredicate ? '< >' : '( )';
+    } else if (this.type === 'list') {
+        return '[ ]'; // scratchblocks renderer does not have a "list" symbol
+    }
+    return ''; // should never happen
+};
+
+
 // CommandSlotMorph ////////////////////////////////////////////////////
 
 /*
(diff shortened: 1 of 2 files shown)


jmoenig:
Thank you!!

I *love* this idea, it's wonderful to be able to convert Snap scripts into text-code that can then be pasted into the Scratch Forums. Did you know that one of my very first contributions to Scratch (way back when, I think it must have been v 1.2) was the ability to convert scripts to XML, and my example was always pasting blocks into the forums? 

This looks quite pullable. Just one question: Would it be possible - and perhaps even make sense - to just use Snap's built-in "Codification" mechanism for this, instead of hard-coding the Scratch-Forum block syntax into the Snap sources? That way we could offer any kind of stringification, and externalize it, too. The only new feature we'd have to add would be to pop-up an alert-box letting you copy the generated code out of Snap. 

Does that make sense at all?
Thanks!

tjvr:
I'm glad you like it! :-)

I thought about using codification to implement this, but I don't think it would make sense. Here's why: the scratchblocks syntax very closely mirrors the Morphic structure of the blocks. This PR is able to be so small because of this. It makes more sense to think about scratchblocks as an alternative "rendering" of the block morphs, than as a language that you can codify to! This means that I don't have to hard-code all of the blocks, in fact.

So while it's a neat idea, I don't think it would make sense to combine scratchblocks and codification. Unless codification works very differently than I thought!

> Did you know that one of my very first contributions to Scratch (way back when, I think it must have been v 1.2) was the ability to convert scripts to XML

You know, I do remember that! It was called Chirp, wasn't it? :)

jmoenig:
Ah, okay, your explanation does make sense yes. Awesome, I'm going to pull this next up. Thank you so much!

tjvr:
Great! I'm very pleased :D Thanks Jens!

BookOwl:
@tjvr, https://scratch.mit.edu/discuss/post/2423593/ https://scratch.mit.edu/discuss/post/2423636/

tjvr:
thanks, fixed :)

blocks.js
```diff
@@ -2518,6 +2547,14 @@ BlockMorph.prototype.userMenu = function () {
         'userDestroy'
     );
     menu.addItem(
+      'stringify...',
+        function() {
+            var code = myself.stringify();
+            window.prompt('scratchblocks code for you to copy and paste', code);
```
jmoenig: okay, how about replacing this with a Morphic dialog box? Now that Morphic supports copy & paste of text, we can do this nicer :-)

cycomachead: We should make a generic one that has some text about _how_ to copy and paste. C&P is only supported via keyboard shortcuts so it's non-obvious to many. 

Anyway, there's a mostly correct dialog in #1084

blocks.js
```diff
@@ -2518,6 +2547,14 @@ BlockMorph.prototype.userMenu = function () {
         'userDestroy'
     );
     menu.addItem(
+      'stringify...',
```
cycomachead: I think the word "stringify" is confusing. Namely many Snap<em>!</em> users aren't familiar with the term "string", but also, in my mind at least "stringify" implies are more generic use than Scratch Blocks syntax.

cycomachead: That later comment also means I have reservations about using the function `stringify` for something like this, as opposed to say `toScratchBlocks`, which I think is much more clear.

tjvr: Not the best name, no! But has to be better than "scratchblocks"...

blocks.js
```diff
@@ -9584,6 +9705,15 @@ SymbolMorph.prototype.setLabelColor = function (
     this.setColor(textColor);
 };
 
+// SymbolMorph to "scratchblocks"
+
+SymbolMorph.prototype.stringify = function () {
+    var symbol = {
```
cycomachead: This format seems confusing to me, especially since there's only one option. 

Why not just use a simple conditional?

blocks.js
cycomachead: > But has to be better than "scratchblocks"...

Wait, why? (or am I just missing something)

tjvr: The name "scratchblocks" is really unhelpful; to someone unfamiliar with the term, it doesn't really help explain what it is. Maybe "export text" or something?  The closest analogy I can think of is the export project summary feature from Scratch 1.x...

blocks.js
tjvr: It's mapping Snap! icon names to scratchblocks icon names.  We might add more icons in the future, I guess.

cycomachead: I get what it's doing, but only after reading it like a few times. That reason makes sense, though it seems better, to me, to  add complexity as needed, that's all.

blocks.js
cycomachead: I guess so, but "scratchblocks" is at least specific about what this is. If you google "scratchblocks" then your project at least comes up at the top so it's discoverable. 

Also, since this is only for the Scratch forum (at the moment) we might just say "export for scratch forum" (and update it as needed in the future)

BookOwl: How about "as forum code"?

blocks.js
```diff
@@ -8744,6 +8847,13 @@ TemplateSlotMorph.prototype.evaluate = function () {
     return this.contents();
 };
 
+// TemplateSlotMorph to "scratchblocks"
+
+TemplateSlotMorph.prototype.stringify = function () {
+    var category = this.parent.stringifyCategory() || ' :: grey';
+    return '(' + this.children[0].stringify() + category + ')';
```
dekrain: In Snap, template variables are in doubled rings, like here: `script variables ((a)::grey) ::grey`

tjvr: Already done! Please try the demo version

blocks.js
tjvr: `forum code...` is what I was thinking too!

blocks.js
tjvr: That seems like a lot of code just for copy/pasting scratchblocks!

If you don't mind I'm not going to do that here, and instead open a new issue for that improvement. :)

blocks.js
BookOwl: Your demo site either isn't updated to the newest code, or your code is still buggy, because I'm still getting incorrect results for rings

blocks.js
cycomachead: forum code sounds good to me!

blocks.js
cycomachead: It's mostly just layout / morphic code. 

And if we use it in 2 place, then we could create a `CopySelectionDialogMorph` and store it in widgets.js, so then it's like...no code. ;)

blocks.js
tjvr: > I'm still getting incorrect results for rings

What specific examples don't work?

tjvr:
Okay, I've renamed the methods & the menu item. :)

I opened #1648 for the dialog enhancement.

Other than that this should be ready to merge :)

cycomachead:
LGTM. (At least the code, I haven't had a chance to test it yet)

blocks.js
```diff
@@ -1852,6 +1852,35 @@ SyntaxElementMorph.prototype.isEmptySlot = function () {
...
+    }).join(' ') + this.toScratchblocksCategory();
+    if (nb) {
+        result += '\n' + nb.toScratchblocks()
+    }
+    return result;
+};
+
+SyntaxElementMorph.prototype.toScratchblocksCategory = function () {
+    // private. answers with scratchblocks category specifier
+    if (!this.category) return '';
```
cycomachead: Nit: Jens uses `{return ''; }` as the format of these things.

blocks.js
```diff
@@ -2518,6 +2547,14 @@ BlockMorph.prototype.userMenu = function () {
         'userDestroy'
     );
     menu.addItem(
+      'forum code...',
+        function() {
```
cycomachead: Nit: space between n and (

brianharvey: I vote for "Scratchblocks" over "forum code."  It really is quite Scratch-specific.  When we have our own forum with Snap!blocks we can change the item name.

tjvr: changed to `scratchblocks code...`. :)

cycomachead:
> I vote for "Scratchblocks" over "forum code." It really is quite Scratch-specific. When we have our own forum with Snap!blocks we can change the item name.

Speaking of which we should get on that. Remind me to bring this up next week!

blocks.js
BookOwl: Nevermind, it's working now.

jmoenig:
I've changed my mind about this, nice as it is. Basically, we will get another way to export scripts in our own format. For this we might make a forum plug-in that renders the exported scripts. But then we can also drag thoses shared scripts back *into* Snap, and use them as "real" ones. We need to ponder this some more....

tjvr:
Fair enough :-) Feel free to close!

brianharvey:
If the alleged our-own script-to-text format is > 1/2 year away, I would still argue for pulling this in the coming release.

jmoenig:
The format is mostly done, it's @bromagosa's PR #1746

PR #1663, opened 2017-02-16
fixes #1349

### LINKED ISSUE #1349: fill tool crash
Hi,
I already report this bug. When I try to fill in a big region with the costumes editor, the tool crashes. I tested it with Firefox and Google Chrome. Can somebody look at this problem.
Thanks

bromagosa:
It only happened in some browsers and in particular layouts due to the brilliant decision behind all JS numbers being floats. I guess some browsers round stuff up and others don't, thus 480 became 479.999998 in Firefox, where the mouse pointer position also had a fractional part, and since it couldn't detect the color of pixels in nonexistent float coordinates, the stack kept growing, forever looking for the next pixel.

All in all, I `Math.round`ed the position where the algorithm starts comparing pixel colors and that solved it.

### DIFF
--- a/paint.js
+++ b/paint.js
@@ -740,7 +740,7 @@ PaintCanvasMorph.prototype.floodfill = function (sourcepoint) {
         ctx = this.paper.getContext("2d"),
         img = ctx.getImageData(0, 0, width, height),
         data = img.data,
-        stack = [Math.round(sourcepoint.y) * width + sourcepoint.x],
+        stack = [Math.round(Math.round(sourcepoint.y) * width + sourcepoint.x)],
         currentpoint,
         read,
         sourcecolor,


jguille2:
Yes @bromagosa,
And unfortunately we have to live with these [subnormal](https://en.wikipedia.org/wiki/Denormal_number) numbers (sorry for this non-inclusive name) into our Snap projects.
float_numbers

jmoenig:
Yay! Thanks for tracking this down, Bernat. While others browsers never had such issues Firefox used to have conflicts between actual floats and logical integers all the time, which made me either round or truncate most dimensions in earlier versions of Snap. However, this resulted in often ugly animations. Therefore I was really glad when I found out that these conflicts were gradually getting resolved in Firefox and I was able to progress to better precision. Hmm... so, it seems there are still cases in which Firefox hiccups. In this case it seems fine to round (or should we use truncate/floor?), since this only affects the paint editor anyway.

This is fantastic, I'm so glad you found a solution for this longstanding, very annoying and hard to reliably reproduce issue!

bromagosa:
We need to round, cause truncating or flooring would be out by 1 position :)

paint.js
```diff
@@ -740,7 +740,7 @@ PaintCanvasMorph.prototype.floodfill = function (sourcepoint) {
         ctx = this.paper.getContext("2d"),
         img = ctx.getImageData(0, 0, width, height),
         data = img.data,
-        stack = [Math.round(sourcepoint.y) * width + sourcepoint.x],
+        stack = [Math.round(Math.round(sourcepoint.y) * width + sourcepoint.x)],
```
cycomachead: Do we need to round twice? I'm not sure why it's important to round `.y`?

Also, is it ever possible for this to include a pixel that's out of bounds if you were to select the lower right corner and things rounded up? Perhaps that's a super weird edge case, but I might write:
`parseInt(sourcepoint.y * width + sourcepoint.x)`. I don't know if that would make the normal case more likely to be 1px off of where someone clicked?

bromagosa: I first tried `Math.round`ing the whole expression and it didn't work. The numbers were off by so much that it paints outside the shape. You absolutely need to round `y`.

`parseInt` instead of `Math.round` at the outer expression also does work, but what it actually does is truncate the float at the integer part, which would result in a pixel being off in some edge cases.

jmoenig:
great, just to remember: We've just now also made sure that the primitive command block already fixes this, so it doesn't have to be adjusted anymore.

bromagosa:
Closing because I resubmitted it to the dev branch :)

PR #1665, opened 2017-02-19
Change gui sizes

erichake:
Hi Jens,

I've made some little gui changes for my young pupils : it's only a matter of taste, so I'm really not sure you will like it. I've submitted this pull request just in case you would like it !

Beside, I only spent few days inside snap! code… I tried to follow your way of coding, but I'm not sure I did it the right way... By the way your code is really brilliant ! I love this prototype way of doing things and I had a real "wow effect" when I realized that all things was made without DOM elements (well canvas only, I presume).

At first run (with empty localStorage) snap! category buttons look like this :

image

I've changed this to :

image

Also I incremented initial font sizes for menus, blocks and tabs. I've also changed palette and stage width. All of this could be seen here :


erichake:
Oups, I've made a mistake...

In gui.js, I wrote : 

> StageMorph.prototype.dimensions = new Point(360, 270);

But it should stay like it was written : 

> StageMorph.prototype.dimensions = new Point(480, 360);

brianharvey:
Sorry nobody's responded to this in a long time.  I'm way behind on email.

Even though I'm not Jens I'm sure he wouldn't accept this as a fixed change in size.  If you wanted to make the size of the palette tabs follow the Zoom Blocks value, that would be a bit less of a hard sell, because at least the user is explicitly asking to lose some scripting area space in favor of palette space.  But then there are other things that should change, too, most notably the Make a Block dialog.  Jens has not been encouraging people to work along these lines, but if you want this desperately, that's how to move this from nonstarter to unlikely.  :)

Meanwhile I'm closing this.

PR #1666, opened 2017-02-19
change import media dialog box

erichake:
I've changed import dialog box for costumes, backgrounds 	and sounds and I've created categories, like this :

image

Beside, this import dialog loads costumes and backgrounds as light thumbnails instead of large picture files. 

I've also found out that on my mac, a large amount of sound files were not loading correctly (wav codecs not recognised in some browsers, for example for "A-trombone.wav" on my firefox and my chrome). So I've changed .wav to .mp3, and it also reduced the Sound folder size by 3Mo.

 All of this could be seen here :


cycomachead:
I think this is really helpful! 

A few things:
* We should eliminate the `description: "",` in the JSON files since it's not used. (And in the case of costumes, it's a decent amount of bytes...)
* Why not organize "section" as a higher-level object?
  `{ 'section-name': [{}, {},...],  'section-name': [{}, {},...]}`
    Maybe you did consider that and decided against it, but this seems like it might help make it easier to reason about how files are organized.
* In the future, I think this could be split into a couple smaller pieces, like separating the thumbnails out because it will make it easier to review, but this is fine. :)

erichake:
Yes you're right : I've commited the whole thing at once but I shouldn't have to do it this way... I understand it's not easy for you to review : I'm really sorry about that !

- For description key you're right, I will remove it...
- For section key I thought of what you say, but I decided to do it another way because I wanted to give the ability (in the future) for one image or sound to be attached to more than one section. For example : 

`{ "name": "birthday", "fileName": "birthday.mp3", "section": "Music" }`

could become in future : 

`{ "name": "birthday", "fileName": "birthday.mp3", "section": ["Music" , "Party"] }`

jmoenig:
Thank you, @erichake , for this really nice contribution! You've really caught on the Morphic way of doing things very quickly and thoroughly, I'm impressed. I really like these changes, especially the categories and the thumbnails, this feels a lot better and is much nicer to use. I'll definitely take a closer look at this pull-request with the intention of integrating it into the main Snap trunk. But I'm going to have to ask you for a little patience, since I'm mostly travelling this week. In the meantime you've already convinced Bernat and Michael :-)

gui.js
```diff
@@ -3027,11 +3027,10 @@ IDE_Morph.prototype.importMedia = function (folderName) {
     // costumes, backgrounds or sounds
     var myself = this,
         msg = this.showMessage('Opening ' + folderName + '...');
-    this.getMediaList(
-        folderName,
-        function (items) {
-            msg.destroy();
-            myself.popupMediaImportDialog(folderName, items);
+    myself.getURL(
+        folderName + "/catalog.json",
```
cycomachead: This should be:

`this.getURL(this.resourceURL(folderName, 'catalog.json'), ...`

gui.js
```diff
@@ -3027,11 +3027,10 @@ IDE_Morph.prototype.importMedia = function (folderName) {
...
+        function (txt) {
+            myself.popupMediaImportDialog(folderName, JSON.parse(txt));
```
cycomachead: I'll need to this about this - we have a getMediaList function or something similar.

gui.js
```diff
@@ -3041,40 +3040,144 @@ IDE_Morph.prototype.popupMediaImportDialog = function (folderName, items) {
...
+    items.sort(itemsSortFilter);
+    sections.sort(sectionsSortFilter);
+    sections.push({ "name": "All" });
+
+    var listField = new ListMorph(sections,
+        function (element) {
+            return element.name;
+        },
+        null,
+        function () {
```
cycomachead: We don't need this.

gui.js
```diff
@@ -3041,40 +3040,144 @@ IDE_Morph.prototype.popupMediaImportDialog = function (folderName, items) {
...
+
+    var sectionsSortFilter = function (a, b) {
+        if (a.name === "other") return 1;
+        if (b.name === "other") return -1;
+        return itemsSortFilter(a, b);
+    };
+
+    var colect = {};
+    items.forEach(function (item) {
+        var section = (item.section === "") ? "other" : item.section;
```
cycomachead: I think we might to make the check more broad... Maybe `isNil(item.section) || item.section === ''` ?

Otherwise if someone leaves off a section key then the sections list would have a "null" section.

cycomachead: Actually, I think it'd be an `undefined` section...anyway I don't think we want that either. :)

erichake:
Thank's for this review, I've made the changes : 

- I think the getMedialList function is now useless in the code (it was looking for the COSTUME, BACKGROUND or SOUNDS text file inside the folders)

- new ListMorph() takes care about undefined params : you're right I didn't see it !

- ok for letting the possibility to "forget" the section key in the json files

erichake:
Oups sorry I said a wrong thing : I didn't see that getMedialList was used by getExamplesProjectList...

nathalierun:
@erichake Thank you for this great contribution 👍 

Would it be difficult to add a right button in the costumes area to import a costume from the media library ? idem for the sounds in the sound area ?
capture d ecran 2017-02-21 a 07 59 29

Because it was the first thing I tried to do when I looked for the import dialogue box for the costumes.

AlainBusser:
One other thing which would be great: That it be possible to draw the costumes with blocks, like draw a star with a sprite, then import the star as a costume; or draw a fraction with LaTeX and have it as a costume afterwards.

The biggest fun would be a "build a costume" block ;-) (imagine that a sprite could build its own costume when created as a clone)

jmoenig:
@AlainBusser we have that in Snap! It's a library called ```save and restore picture drawn by pen```

here are a bunch of sample projects that make use of this feature:

[Valentine](http://snap.berkeley.edu/snapsource/snap.html#present:Username=jens&ProjectName=Valentine)
[Ferris Wheel](http://snap.berkeley.edu/run#present:Username=jens&ProjectName=Ferris%20Wheel%202016)
[Towers of Hanoi](http://snap.berkeley.edu/snapsource/snap.html#present:Username=jens&ProjectName=Towers%20of%20Hanoi)

and the coolest of them all, the world's fastest fractals, by Dan Garcia:

[Fractals](http://snap.berkeley.edu/snapsource/snap.html#present:Username=jens&ProjectName=save%20and%20restore%20pen%20trails%20FRACTAL&editMode)

Enjoy!

cycomachead:
> I think the getMedialList function is now useless in the code (it was looking for the COSTUME, BACKGROUND or SOUNDS text file inside the folders)

I think this should be used, even if it's just calling JSON.parse. 

But also I think we should extract the sections stuff out of the actual morph layout code. I'll write more later after work when I can better look at some stuff. 

The mediaList could be adapted to look for folder/catalog.json which could also be used for libraries and example projects so that we don't have multiple formats.

> For section key I thought of what you say, but I decided to do it another way because I wanted to give the ability (in the future) for one image or sound to be attached to more than one section. For example :

And this makes sense. Both formats are easy so I don't mind much either way, I was more curious as to why. :)

AlainBusser:
Thanks for the examples; I have an other idea yet: [Gerbert abacus](http://calmeca.free.fr/calculmecanique_php/rubriques/Fichiers_Blaise_Pascal/Fichiers_historique/Fichiers_Abaques/Abaques.php?lang=eng) 

The sprites (Gerbert called them "apices") each have a numerical value (but depending on where they are inside the abacus) and each apice should have as a costume, the digit yielding its value.

I'll try it when I have some time

cycomachead (commented):
I left some code comments - most are minor. 

This basically is 2 parts, including #1594 
I think it would be good, if we do this to also update the code for libraries and examples, which would have no section data. It might make sense to do that first, but it doesn't matter me to. :)

index.php
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
```
erichake: Oups sorry, I didn't see it was in my commit !

gui.js
```diff
@@ -3041,40 +3040,140 @@ IDE_Morph.prototype.popupMediaImportDialog = function (folderName, items) {
...
+
+    var listField = new ListMorph(sections,
+        function (element) {
+            return element.name;
+        }
+    );
+
+    content.color = dialog.color;
+    listField.setWidth(listFieldWidth);
+    listField.contents.children[0].alpha = 0;
```
erichake: You're right : childrens are changing sometimes :-)
So I finally found the MenuMorph without dealing with children : 

listField.listContents.alpha = 0;

is far better...

erichake:
I've made the changes on last commit, but the setTimout is still in code of gui.js : 

capture d ecran 2017-02-22 a 17 53 20

And I really don't like setTimeouts :-)...

brianharvey:
I have a minor request:  Could you change the little "SVG" flag to "vector" or maybe just "vec"?  I would like our users not to have to know what an svg is.  (And no doubt someday someone will invent another vector file format.)  Otherwise it's great!

Adding "blessed by Jens" label because he did!

jmoenig:
@brianharvey , you can't just add "blessed by jens". I really love this PR and we're going to pull it, but "blessed by jens" means that I've reviews the code and it's something that helps me remember I actually did. So, please don't confuse my memory!

That said, @erichake, this is beautiful, and *now* I did review it. :-)

jmoenig:
Okay, sorry, I had to "unbless" this for now, because we found out, that a vast majority of the costumes are still broken, especially having the wrong size. This issue has been reported elsewhere in schools and needs to be addressed before we can pull this and work on it.

@brianharvey , what's the status on the broken costumes? These need to be fixed urgently, so we can work on both this and on the release.

jmoenig:
That said, @erichake your changes are beautiful, and we *will* incorporate them!

brianharvey:
Sorry, Jens, I thought I'd understood you to say that it _was_ blessed.

Remind me what's wrong with the costumes?  I just copied them from Scratch.

jmoenig:
@brianharvey , many costumes are way too big, about double the size they are in Scratch. Teachers (and Mary) have complained about that a lot over on Piazza, because they have to tell students to scale the whole sprite down to 50% size to get them to look right. Worse, some costumes are "normal", and others are not, so changing to another costume often requires a sprite-resize.

I'm pretty sure that this is an unintentional artifact that happens when you edit graphics on a retina display. Not in Snap but in some other image processing app. Anyway, this has been pretty urgent for a longish time, so we should work on it now, to fix it for the upcoming release. After that I really want most of @erichake's contributions. If we can get the costumes right in time, I'd love to include these contributions in the upcoming release as well.

brianharvey:
Hmm.  It looks to me as if the problem is with new Scratch-2.0 bitmap costumes; vector costumes and old Scratch costumes both seem okay.  I didn't edit any of the costumes, though, so I don't believe your theory.  _My_ theory is that Scratch added double-density costumes on purpose for the benefit of retina displays, and they have code that half-sizes them on input on non-retina displays.  Will investigate further...

cycomachead:
We really should work on this -- not sure how big the conflicts are now, but it should be fairly easily resolved.

cycomachead:
@jmoenig @brianharvey How do we feel about tidying this up and merging soon? 

The biggest "downside" is having thumbnails in the repo, but it's really not all that much data.

brianharvey:
Yes, please!

PR #1669, opened 2017-02-21
Draw an import button in Costumes, Backgrounds and Sounds tabs

erichake:
There is an import menu item, but I thought it would be nice to give young students another way to import medias from library...

That's why I added a button in Costumes/Backgrounds/ Sounds tabs : 

importbutton

It's possible to see it here : 

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -7458,7 +7458,8 @@ WardrobeMorph.prototype.updateList = function () {
         icon,
         template,
         txt,
-        paintbutton;
+        paintbutton,
+        importbutton;
 
     this.changed();
     oldFlag = Morph.prototype.trackChanges;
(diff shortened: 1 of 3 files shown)


cycomachead:
I do agree a there should be a function to import media here, however, I'm not sure it should be a glyph. Icons are nice because they don't need translating, but I think the text "Import [media]" is much better. 

However, this button looks really really nice, so maybe the provided hint isn't too bad. :) I don't have much opinion on this either way.

gui.js
```diff
@@ -7502,15 +7503,40 @@ WardrobeMorph.prototype.updateList = function () {
...
+    importbutton.color = IDE_Morph.prototype.groupColor;
+    importbutton.highlightColor = IDE_Morph.prototype.frameColor.darker(50);
+    importbutton.pressColor = importbutton.highlightColor;
+    importbutton.labelMinExtent = new Point(36, 18);
+    importbutton.labelShadowOffset = new Point(-1, -1);
+    importbutton.labelShadowColor = importbutton.highlightColor;
+    importbutton.labelColor = TurtleIconMorph.prototype.labelColor;
+    importbutton.contrast = this.buttonContrast;
+    importbutton.drawNew();
+    importbutton.hint = "Select a costume from the media library";
```
cycomachead: I think "Select" should be "Import"

We should also probably add a condition for the hint to say "background" if it's on the stage.

gui.js
```diff
@@ -7502,15 +7503,40 @@ WardrobeMorph.prototype.updateList = function () {
...
+    importbutton.setPosition(new Point(x, y));
+    importbutton.fixLayout();
+    importbutton.setCenter(icon.center());
+    importbutton.setLeft(paintbutton.right() + padding * 2);
+
```
cycomachead: One less blank line.

gui.js
```diff
@@ -7813,6 +7846,7 @@ JukeboxMorph.prototype = new ScrollFrameMorph();
 JukeboxMorph.prototype.constructor = JukeboxMorph;
 JukeboxMorph.uber = ScrollFrameMorph.prototype;
 
+
```
cycomachead: This isn't necessary. :)

gui.js
erichake: "Select a costume from the media library" is the standard message, I didn't wrote it : it was already in code (hint used for costumes and sounds menu items).

brianharvey:
This is nice.  If we're going to follow Scratch in this way, which I think is fine, we should also have an import-from-file button, right?

cycomachead:
> we should also have an import-from-file button, right?

Where would this go?

jmoenig:
This is really, really nice, @erichake ! We decided we're going to pull it but then probably only take the part that adds the import button, not the parts that change the general looks of the UI. That's why we added both tags "blessed" and "under review". Thanks for your great contributions.

Issue #1647, opened 2017-01-31
It needs to be easier to import to a variable


cycomachead:
Right now you have to do this through having a watcher on the stage. 

I propose the following:

* Add the import command to the _declaration_ blob of all variables (i.e. the palette or the script variables block)
* If the type of a variable is already a list, then we split the file by lines automatically.

brianharvey:
Is there a reason we don't just have a reporter for this?  Why does the import have to have a variable at all?  (As opposed, for example, to being directly the input to a SPLIT block.)

cycomachead:
> (As opposed, for example, to being directly the input to a SPLIT block.)

I suppose you're right this would work, but IMO, it's not great to shove tons and tons of text in an input slot - it's hard to read and takes up space in the scripting area. I guess, I wouldn't want to _encourage_ this, but it doesn't seem like a problem necessarily.

Also, and perhaps more importantly, AFAIK #20 Is still around and this would lead to wonky behavior. There's actually a bug linked to the split bock usage of pasting text into the input slot.

brianharvey:
Huh?
![import](https://cloud.githubusercontent.com/assets/3885074/22490394/f0c9871c-e7d1-11e6-9023-43cf0272a0ad.png)

cycomachead:
We can't do that because the web doesn't allow it. That also would break the portability of projects. The desktop app I have will work with a custom block but there's still no portability. 

I thought you meant like adding a context menu to every input. 

-- 
Michael Ball
From my iPhone
michaelball.co

brianharvey:
Hmm, so, we're not allowed to have a filename embedded in the program?  It has to be from a browser open dialog?  OK, in that case, the input to the IMPORT block should be a caption to be included in the open dialog, e.g., "choose a database file."

cycomachead:
Ah, well we can certainly build that! Well, minus the caption. You can't add a caption b/c the import dialog is attached to the window. 

![screen shot 2017-01-31 at 7 24 06 pm](https://cloud.githubusercontent.com/assets/1505907/22494252/db5d14c0-e7ea-11e6-8c53-0d7cd407d5d4.png)

However, in lots of cases, you don't want the import to be run during the execution of a program - like when building a hangman game, so I still think we should have a non-block method for importing stuff, but I could see a block being useful if you wanted to do something interactive.


BTW, try this project:
http://snap.berkeley.edu/snapsource/snap.html#present:Username=cycomachead&ProjectName=import%20block

Using:
[...]

Issue #218, opened 2013-10-24
Ability to Export a list


cycomachead:
It would be really handy if we could export a list to a text file.
I'm not sure of the best way of handling all cases but it would be helpful for students.

cycomachead:
@jmoenig @brianharvey 

I was working on some stuff for #983 a realized that we could basically export anything from a watcher as text, since every morph has a .toString() method.

bromagosa:
Closing this as per https://github.com/jmoenig/Snap/commit/dab03fac5ded5518d726ff716941c6ff377f0193 and https://github.com/jmoenig/Snap/commit/a681880965e8a724e6bb7a68bb8057f518c701e7.

PR #1670, opened 2017-02-22
Lists are exportable to JSON, CSV, plain text or XML from watchers

bromagosa:
Any love for this? I needed it myself and thought I'd contribute it to the trunk instead :)

image
image


jguille2:
Yes, I like it!!
If this goes to the trunk, I suggest a little more:
  - In the line of #1647 , I think import/export option of a var could be into any blob of it (not only into the watcher).
  - JSON and XML into import action will be great.
  - For me, JSON and XML are great. But for users, a CSV option will be the greater. I know JSON action is directly (and _serialize_ to xml was implemented too)... but we can consider this. I'm still using [this project]( https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/issues/1647#issuecomment-276627501) to import/export to CSV.

Thanks,
Joan

cycomachead:
I can see the use cases, though I think this misses the primary one that people want: a 1D list as newline-delimited file. :)

I also agree that a CSV is probably more common for 2D lists.

lists.js
```diff
@@ -398,6 +398,22 @@ List.prototype.becomeLinked = function () {
     }
 };
 
+List.prototype.asJSON = function () {
```
cycomachead: Why can't we define a `.toJSON` method and then just call `JSON.stringify()`?

Though, I don't know how that would work for list items that can't be easily serialized to JSON...

jguille2:
Hi @cycomachead 
But I think that CSV is also ok for 1D lists (new-line delimited but without commas because we have only one value (colum) per row)

I've updated [my project](http://snap.berkeley.edu/snapsource/snap.html#cloud:Username=jguille2&ProjectName=data_to_csv) in this way. Then we get:
  - A single text
csv_single
  - A simple list
csv_list
  - A "only one" list of a list (forcing 2D)
csv_onelistoflist
  - And a lists of a list (typical 2D)
csv_listsoflist

In this project "from csv" block is the reverse action.

cycomachead:
Oh that's true. I don't care how it's implemented though.

My concern is displaying all these acronyms to students and teachers. Perhaps "Plain text (CSV)" is a descriptive enough field to convey what people want. In my experience getting data out in a simple format is the most desired, and CSV will appear more complex than necessary to those who aren't familiar.

jguille2:
Yes, I undestand your point of view.

But this "to csv" is more powerful (we can choose quotes, separators...) and then, we can build our custom blocks for different users...

csv_toplainreporter

where

csv_toplainbuilding

cycomachead:
Having a default option doesn't prevent a custom block. Easily exporting a list has been a request for quite a few years. Using a custom block requires quite a few additional steps to get the data out of Snap! - it's something that often causes people trouble. 

Also, as an aside, I think importing will be a necessary followup to this. Regardless of format, it will be necessary to import data, otherwise users will expect this to be a way to share info and it won't work.

jguille2:
Yes, you're right.

I am merging two different things.

My example, like a lib,  are "only" custom blocks to play with them.

Returning to this PR (I apologize my dispersion) and to our extra requests to Bernat, in the sense to the implementation into the trunk, I suggest:

  - **Symmetrical import/expor**t options (json,xml... in both options).
  - Import/export **options added** in all instances of the variable (or at least, also added **into the palette**).
  - **Add a third option (plain text**) to the export/import actions. My suggestion was the same I've exposed in the [CSV example](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1670#issuecomment-281836503). To code this would be easier without quoting and separator options... but the behaviour could be the same.
  - And like the CSV example, also **symmetrical list/non_list vars behaviour**. Then we can decide the behaviour of JSON and XML of a simple string. The idea would be the same as csv case: the same options for all vars.

Thanks @cycomachead and @bromagosa !

lists.js
bromagosa: Because a List can't be converted to JSON, unlike an Array. Thus, if you have a List of Lists, and call stringify on the outer one, the inner one will be serialized as "A List [1,2,3,4]".

jguille2:
Note: Do not look (for now) at my example project.
I see it's not updated in the cloud!! I guess yesterday I lost the connection and I was working in local all the time. Excuse me :(

bromagosa:
The way it's implemented, adding support for another format is trivial. I needed these two myself, but I can add CSV too.

Adding plain text is a bit more complex because we need to make sure the list is uni-dimensional, and we need to decide for a separator that will make everyone happy (ha!).

lists.js
bromagosa: Oh now I know what you mean, and you're totally right! :sweat_smile:

lists.js
```diff
@@ -398,20 +398,18 @@ List.prototype.becomeLinked = function () {
...
+List.prototype.toJSON = function () {
+    return this.contents;
+};
+
+List.prototype.toCSV = function () {
+    // Expects the list to be properly mappable into a
+    // two-dimensional table.
+    return this.contents.map(function (eachRow) {
+        return eachRow.contents.map(function (eachCell) {
+            return JSON.stringify(eachCell);
```
cycomachead: Will this automatically add quotes around items with a `,` in them?

I love how simple this is. :)

lists.js
```diff
@@ -398,20 +398,18 @@ List.prototype.becomeLinked = function () {
...
-                        return itemToJSON(each);
-                    }) +
-                ']';
-        } else {
-            return JSON.stringify(item);
-        }
-    };
-    return itemToJSON(this);
+List.prototype.toJSON = function () {
+    return this.contents;
```
cycomachead: This won't work with a linked list, though, right?

lists.js
bromagosa: It will! :D 

Snap:
image

LibreOffice:
image

lists.js
bromagosa: Damn, you're right. Changing it to `asArray()` everywhere...

objects.js
```diff
@@ -8496,17 +8497,19 @@ WatcherMorph.prototype.userMenu = function () {
...
-                        myself.currentValue.toString(),
-                        'text/plain;charset=utf-8',
-                        myself.getter // variable name
-                    );
-                }
-            );
+            menu.addItem('export...', this.valueExporter('plain'));
+        } else if (this.currentValue instanceof List) {
+            subMenu = new MenuMorph(this.currentValue);
+            if (!this.currentValue.contents.some(
```
cycomachead: Why don't we make a function `isOneDimension` on `List`?

Also if a list is a single dimension, then it seems like CSV and Plain text should be the same output - I'd merely like the menu text to be more clear.

objects.js
```diff
@@ -8520,6 +8523,48 @@ WatcherMorph.prototype.userMenu = function () {
...
+        value = this.currentValue,
+        contents,
+        format = format || 'plain',
+        ide = myself.parentThatIsA(IDE_Morph);
+
+    return function () {
+        switch (format) {
+            case 'plain':
+                contents = value instanceof List ? 
+                    value.asArray().join(', ') :
```
cycomachead: Ah, yes, the delimiter debate. :)
 my vote here is for `\n` (since that's what BYOB 3 did) and it makes `split (data) by [line]` really easy to do. (Whereas with a `, ` delimiter then students will likely run into issues with `,` vs `, ` (which I mean `,\s`, since that's probably hard to tell the difference).

objects.js
bromagosa: We can do that, but this is the only place where we're checking if a list is one-dimensional. Smells like over-engineering to me.

As for CSV and plain text being the same, see the next comment.

objects.js
bromagosa: What you say here contradicts what you just said one comment above. If CSV on a one-dimensional list should be the same as plain text, then it should obviously be comma-separated. Otherwise, it needs two different names.

We will never manage to please everyone though, because separators are absolutely dependent on each case. I explicitly didn't add `plain text` at first to avoid discussions about separators. You prefer newlines, Joan may prefer semicolons, I'll vote for commas, and this will never be pulled.

This is getting way too long now. I contributed this because I had already written it for myself and it costed ~0 to issue a pull request, but I don't want to spend another morning tweaking it.

Feel free to change it though! :)

jguille2: Thanks @bromagosa,
I don't want to be so heavy... only a comment.
In my example, semicolons (or others) are the separators of the inner list items and newLine is the separator of the items of the main list. Then 1D lists have only newLine separators. You have seen the [example pictures](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1670#issuecomment-281836503).

Just only a comment  (discussing about this import/export feature)... not a work request :)

bromagosa: Yeah, but that's just a matter of taste and particular use cases.

CSV stands for **Comma** Separated Values, so it stands to reason that default separators should be commas.

One-dimensional lists can be thought of as single-row tables (thus comma-separated) or single-column tables (thus newline-separated). To me, it makes much more sense to think of them as rows, but again it's just a matter of your particular use case.

cycomachead: Maybe it's an interpretation of a CSV, but I always assumed a 1D CSV would be separated by newlines, because in a 2D CSV, the `,` separates the items of the inner dimension. A CSV file, just like a TSV is both a comma (or tab) _and_ newline separated file. 
In that case, the comma is the field separate and the newline is the record separator. 

If `[ [1, 2, 3], [a, b, c] ]` is turned into a CSV of:
```
1,2,3
a,b,c
```
Then I think it's reasonable to expect a newline to separate items. Each item in a 1D list is a separate record. When I think about how most 1D lists are used, I think this is more common and also maps fairly well to the idea of rows in a relational DB. 

In any event, I really see very list use for a text file that's one row. Newlines make the output data much easier to read, and work with.

bromagosa: But that absolutely depends on your use case!

If your list is a list of temperatures `[23, 22.4, 25, 30]`, then you're right, it's a single column. But if your list represents a person `['Bernat', 32, 'Barcelona']`, then you're wrong, it's a single row.

Which one is easier to deal with depends again on your use case. Anyone can search and replace commas, but not everyone knows how to `%s/\n/;/g`.

Anyway, I see I'm the only comma advocate, so let me change it. :angry: 

That is, until a fourth person joins the conversation and asks for semicolons.

cycomachead: > Which one is easier to deal with depends again on your use case. Anyone can search and replace commas, but not everyone knows how to %s/\n/;/g.

My primary reason for a newline is this: Both Snap! and Scratch (and BYOB 3) make working with list files separated by newlines easy. Many programing languages have a `readLine(s)` function as part of the standard library. 

Plus, it's much easier to read a long list of words separated by new lines rather than by a comma. (Well, I suppose that's open for debate - but in my experience with the stuff students use and that most people also pretty-print JSON for reading, I don't think it's that controversial.)

As far as some lists representing a single record - you're totally right that it's possible, but in my experience the output matters less in this case. If you really want a comma separated list, as a CSV then it seems easier to make a 1D list a 2D list (all you need is a wrapper `list` block) than it would be to turn the existing list into something separated by newlines -- which is currently a rather annoying process and requires knowledge of how to enter `\n` into Snap! which isn't very well known. 

Sorry for pushing this :( - Dealing with lists in a sane way has been a frequent comment for years, and all my experience suggests newlines are much easier for most students.

jguille2: I did not want to continue... but for me is clear, and not very depending of cases...

For example ['Bernat', 32, 'Barcelona'] is right in a single column (like [23, 22, 24] from mesurements). Because if it was the first entry of more entries (the first row) I'll make a list (entry) into other list (data). I'll make [['Bernat', 32, 'Barcelona']] because later, maybe I'll have [['Bernat', 32, 'Barcelona'], ['Joan', 41, 'Barcelona']].
Oh, I'm getting older :(

bromagosa: > `[['Bernat', 32, 'Barcelona'], ['Joan', 41, 'Barcelona']]`

My point exactly! This would **obviously** map to CSV as:

    "Bernat","32","Barcelona"
    "Joan","41","Barcelona"

Nobody would ever write it it as:

    "Bernat","Joan"
    "32","41"
    "Barcelona","Barcelona"

Whereas if we're using a list to store a single magnitude, like:

    names = ["Bernat", "Joan", "Michael"]

Then, of course, it makes sense to "columnize" them. (Or not! Who cares!)

But this discussion is already pointless since [I've updated my PR](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1670/commits/678d65bca9788341a63f3536053a3032d31ae30a) so that it splits by newlines.

By the way, Scratch 1.4 did export lists as newline-separated values, but *also* as comma-separated values in their HTTP API. What's most useful depends on what your lists represent and what you're going to do with the data after exporting it.

Can we please **\</discussion>** now? :)

liam4: Excuse me for barging in with opinions, but, have we discussed 3D lists? :)

---

Aside from which, I *think* I agree with @jguille2 ? I'd go for something like this:

```
NAME,AGE,PLACE
Bernat,32,Barcelona
Joan,41,Barcelona
```

..Simply because it's a list of entries. I can easily iterate through this with a "for each (person) in (all but first of data)".

The above dataset would be represented as this Snap<i>!</i> list, I believe?

```
[
  [NAME, AGE, PLACE],
  [Bernat, 32, Barcelona],
  [Joan, 41, Barcelona]
]
```

I guess you *could* say that a 1D format could be a list of data represented as something like this:

```
TEMPERATURE,27,25,26,22,24,26,25,30, (etc)
```

(First item optional, of course.) But here's the thing – that would *have* to be imported as this:

```
[
  [TEMPERATURE, 27, 25, 26, ...]
]
```

You see, there's a list around the real list containing our data.

So, why not just make it import as `[TEMPERATURE, 27, 25, 26]`, with the useful list on top? The simple answer can be gotten from our table of people – suppose it has just one entry:

```
Brian,67,America
```

If we were to import all 1D CSV data as a single list, it would look like this: `[Brian, 67, America]`; but as soon as we added another item, we'd get this: `[[Brian, 67, America], [Bernat, 32, Barcelona]]`.  And so we suddenly jumped from a 1D list of items to a 2D list of person-entries.

So, at that point, it would probably be easiest to simply break data along newlines, like so:

```
TEMPERATURE
24
28
27
26
25
(etc)
```

---

But I wonder, what should happen if each (or even just some) of the rows contain only one item?

Usually, especially in a CSV table, every row will follow the same format, such as this one:

```
NAME
Joe
Jake
John
Jens
```

But that's not *always* true – take this, for example:

```
NAME,FRUITS..
Joe,Apple,Banana
Jake,Dragonfruit,Pineapple,Passionfruit
John
Jens,Fractal,Orange
```

..John had no fruits. You can see the problem more clearly if you don't have the header row:

```
Joe,Apple,Banana
Jake,Dragonfruit,Pineapple,Passionfruit
John
Jens,Fractal,Orange
```

So the question is – should John be imported as `[John]`, or `John`?

Perhaps the answer is `[John]`, because some other people have fruits. But if *nobody* else has fruits, then you get the imported array of `[Joe, Jake, John, Jens]`, instead of `[[Joe], [Jake], [John], [Jens]]`. A header row could be handy, in this case, then, since it would count as a row that contains more than one item, hence forcing the other rows to be seen as lists containing only one item.

---

Half the problem is, of course, just maintaining the data structure. If I give you this:

```
LIST OF (CSV OF ([
  [1, 2, 3],
  [a, b, c],
  [x, y, z]
]))
```

..I expect to get `[[1, 2, 3], [a, b, c], [x, y, z]]`. If I give you this:

```
LIST OF (CSV OF ([
  [Joe, Apple, Banana],
  [Jake, Dragonfruit, Pineapple, Passionfruit],
  [John],
  [Jens, Fractal, Orange]
]))
```

..I expect to get `[[Joe, Apple, Banana], [Jake, Dragonfruit, Pineapple, Passionfruit], [John], [Jens, Fractal, Orange]]`.

In my humble opinion, JSON is *totally* the way to go, since it just makes everything make more sense – what if I give you `CSV OF [[Joe, [Apple, Banana]], [Jake, [Dragonfruit, ...]], ...]`? 3D lists are tough, but not with JSON, as you likely know! I do know that JSON is, though portable for programming languages, not so useful for human beings, though. Or regex'ers.

---

Double sorry for such a long post, especially when most of this has probably already been discussed; hopefully it at least makes my opinions and input accessible, all at once.

This isn't really an argument against anyone's ideas, just a collected stream of my input!

TL;DR:

* Comma (or semicolon 😛 ) column delimiters and newline row delimiters work well, since they'll provide you with a table of data that's simple to iterate, and easy to read the CSV of, as a human.
* Make sure data structure is maintained; CSV can do 2D arrays so it's reasonable to think giving the CSV block a 2D list or the LIST block the result of that will work.
* 1-or-greater (and 0-or-greater!) type entries can be messy, keep an eye out for those.
* N-dimensional lists are hard, unless you use JSON, but JSON's not really human-friendly.

bromagosa: Of course we've discussed 3D lists, or 4359D lists. That's what formats like JSON or XML are for, as you say, and that's what the **first instance of this PR** did. JSON and XML can represent *any kind of list* that we can have in Snap<i>!</i>, and that's why they're absolutely uncontroversial.

That's also why I didn't add CSV or plain text in the beginning, and you've all definitely proved my point that I should have stayed out of it, and let somebody else add support for CSV and plain text after (if) this PR went through, and let that somebody else take the bashing over commas, newlines and dimensions too.

brianharvey:
I don't understand the difference between CSV and plain text.

What happens if a delimiter character (for whichever format) occurs in a list item?  (Can we just put that item in quotation marks?)

The important test, I think, is not for one-dimensionality (which as people have said is just a degenerate case for CSV) but >2 dimensionality, for which CSV won't work at all!

Can we have less intimidating names for JSON and XML, like "Javascript data" and "Web page data" or whatever is right?  We could then put the official acronyms in parens for all of them "Javascript data (JSON),"  "plain text (CSV)," etc.

Is there a preferred form for import into Excel?  Or should we maybe export xslx (is that what I mean?)?

What do lists look like in XML and JSON?  XML is pretty purpose-agnostic, right?  So you can make up your own <LIST foo ...> notation?  But JSON is dictionary-based?  So do you use the array index as the key?

Most important question last:  This (at least the plain text one) is a fairly trivial exercise to write oneself in Snap<i>!</i> using COMBINE and MAP.  Everything we build in takes away a programming exercise.  Is this the right thing?  Honestly if I had to choose, I'd much rather have EXPORT (text) so that programs can initiate exporting, and leave the formatting to users.

Sorry I'm so behind on this thread.

jmoenig:
> I'd much rather have EXPORT (text) so that programs can initiate exporting, and leave the formatting to users.

+1

lists.js
```diff
@@ -398,6 +398,20 @@ List.prototype.becomeLinked = function () {
     }
 };
 
+List.prototype.toJSON = function () {
+    return this.asArray();
```
cycomachead: This should be `this.itemsArray();`, I think. So we don't convert linked lists to arrays.

cycomachead:
We can't have a block that exports a file. For the same reason we can't have a block that imports a file. In any event, an EXPORT block and an export UI command serve two entirely different use cases. A programatic means of doing something isn't as useful when you only want to share something with a partner. (Also, again, like importing, this is trivial to write in a desktop application. Well mostly trivial...we don't want kids overwriting other kids files...) This is orthogonal to this discussion, I think. Almost every item in Snap! (sprites, costumes, sounds, blocks, variables, etc) has a way of exporting them through Snap<em>!</em>'s UI. Lists are an anomaly here, and requests export them are some fairly low numbered issues... #218 


And using `combine` to export a newline delimited file is _not_ trivial. For one thing, it requires exposure to the tools blocks. While we moved up teaching the tools blocks in BJC really early on with lists, I do not think we should expect students to need to understand them. Secondly, and more confusingly, it's very difficult to enter a newline in Snap<em>!</em>. You must either be aware of the built in `unicode () as letter` block and understand how to lookup the unicode value for a newline, (and depending on where you land some sites only give you hex codes), or be aware of the multiline library, and how you'd use that. We get way way way more questions about exporting a list in Snap<em>!</em> than we ever did in BYOB. 

As far as names -- we should make them friendly, but we should _not_ obfuscate their true meanings. For one thing, all of this is valid "web data"; there's nothing that demands an API use XML or JSON or well, anything. I'm OK with using the format as "(JSON)", but it needs to be something you can search for. The worst thing we do, is create something interesting, but then kids and TAs have no way of searching for the information they need. 

What I propose is this:
* All items be able to export JSON and XML formats
* 1D lists have an option for "Plain Text"
* 2D lists have an option for "CSV" or possibly "Plain Text (CSV)"

There are so so so so so many student projects which operate on data stored as lists, and many students do clever things like precomputing data and exporting it, or writing text files to import them. The stuff in this PR gets _really_ good when we can get importing to mirror exporting


To answer other questions. (Which you may have figured out, but for the record).
> What do lists look like in XML and JSON? XML is pretty purpose-agnostic, right? So you can make up your own <LIST foo ...> notation? But JSON is dictionary-based? So do you use the array index as the key?

* XML lists look a subset of a Snap<em>!</em> serialized project. They're just calling out directly to that code. XML is just the structure of the markup, and without a schema, I feel like it's usually pretty useless. 
* JSON isn't really a dictionary. It _can_ be a dictionary, but it can be an array too. This is just a set of 2D arrays, or that's my understanding of how it should be.

brianharvey:
Yeah, okay, you're right about newlines.  I forgot how annoying that is.  I'm almost tempted to add a NEWLINE reporter to some library or other!

OTOH if we made this a library, rather than a UI extension, people could at least read the code to see how it's done.  (We should strive to make that a cultural expectation whenever you use a library!)

About JSON, I'm confused by "2D arrays."  Are you making assumptions about the shape of the list?  Lists don't necessarily have a consistent dimension.

bromagosa:
> All items be able to export JSON and XML formats
> 1D lists have an option for "Plain Text"
> 2D lists have an option for "CSV" or possibly "Plain Text (CSV)"

But that's precisely what this code does... :(

> OTOH if we made this a library, rather than a UI extension, people could at least read the code to see how it's done. (We should strive to make that a cultural expectation whenever you use a library!)

To me it's clear that some things belong exclusively to the UI. You wouldn't have a block that opens the paint editor, or a block that selects the circle tool in the paint editor. In the same way, you wouldn't have a block to export lists, because you never want to trigger a system file save dialog from inside a program.

> About JSON, I'm confused by "2D arrays." Are you making assumptions about the shape of the list? Lists don't necessarily have a consistent dimension.

JSON can represent arrays of any dimensions.

Anyway, I think I'd better close this, it's clearly too controversial. Who knew!

cycomachead:
No, please don't close this! :( Now that we've opened the flood gates, I'm willing to deal with the arguing for this. Dealing with list data is pretty much the one major regression Snap! has from BYOB. 

I wasn't making any comments about whether your code was right or wrong when answering Brian - I was simply explaining my ideal design. 

And yeah, I should have said `ND` and not just 2D, my bad.

bromagosa:
Hey, no worries! The code is still there, I originally wrote it for Snap4Arduino and I've just integrated it there :)

I'm reopening it, please let me know when you reach consensus :+1:

DyslexicAwe:
jguille2 wrote [here](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1670#issuecomment-281922287):

> Note: Do not look (for now) at my example project. I see it's not updated in the cloud!! I guess yesterday I lost the connection and I was working in local all the time. 

How much work (hours) have you lost, and would you be willing to write it again? It would be really useful to have the blocks your created to be able to import from and export to CSV.

jguille2:
Hi @DyslexicAwe ,

My note was temporary (I saw it was not updated) but [that used example](http://snap.berkeley.edu/snapsource/snap.html#cloud:Username=jguille2&ProjectName=data_to_csv) is still shared.
 
But now, I'm using [this modification](http://snap.berkeley.edu/snapsource/snap.html#present:Username=jguille2&ProjectName=csv). In this, I deleted 'quote' concept because it was not consistent (for export/import methods). Anyway, if your data have like ';', ','.... you can use a custom separator ('$$$' for example).

That's all,

Joan

jmoenig:
one issue we just now talked about is what happens if the list(s) contain(s) non-encodable stuff such as sprites, costumes, sounds etc. I feel we need to think about this a little more. But it's an important feature that many people want, so we *do* want to think about it! Thanks, @bromagosa

brianharvey:
Man, I'm so envious about all these @jmoenig-@bromagosa conversations lately!

Sprites and things _are_ encodable, of course; it's just that the encoding might end up including most of the project via the environments of procedures.  And, yeah, since serialization is temporarily destructive it might require stopping the clock.

But I think probably the best thing would be to do a really good job of defining print forms for these non-text-like things.  For sprites, for example, it might be something like

#[sprite Alonzo [locals (score 23) (serial_number 6) ...] [blocks ...]]

(I know I'm handwaving here, but the general idea is to give the information that's likely to be interesting and/or useful in determining which sprite this is.  Most sprites don't have local variables or blocks at all, so they'll just be #[sprite Alonzo].)

For blocks, including anonymous ones, something like

#[block <is %small substring of %big ?> expr... #closure (var val) (var val) ...]

where the <is...> could be delimited by parens for reporters or brackets for commands, or #anonymous for a grey ring.

But all this is probably a long-term project, and in the short term just have #sprite and #ring as placeholders.

cycomachead:
> But all this is probably a long-term project, and in the short term just have #sprite and #ring as placeholders.

We pretty much have this for almost every object, since they implement a `.toString()` method. 

I agree with what you say - a simple version works for now, and we can make iterative improvements later. The goal of this isn't to ever use the serializable Snap! Objects, so simply including an id (like the sprite name, which I believe we already do) should work well enough.

brianharvey:
My idea wasn't that we would want to re-import the printform.  I'm imagining that the user has a list of blocks, and just wants to know what's in the list.  Even for sprites, the names may not be enough if you make a bunch of clones.

jmoenig:
Users want to export data so they can import it to another project, as a way to share data between projects. Short of being able to create an ad-hoc server (which we *really* want) export/import works. But we should make it easier to encode / decode text. In dev mode we already have a bunch of encoding options for text. Maybe we could simply add a library for the most common cases of lists that only contain alphanumerical stuff.

I'm just really, really reluctant to encourage any notion that "data" is only about strings and numbers.

brianharvey:
> I'm just really, really reluctant to encourage any notion that "data" is only about strings and numbers.

Yeah, +1 to that!

cycomachead:
Yeah, I guess my point was more about how JSON/CSV exports were intended to be used (based on Bernat's original post). For anything that could be (re)-imported or shared, then we already have the capabilities to encode anything as XML and import that individually. (Well, it might not be exposed in the UI, but we certainly could allow dropping a list.xml file into the IDE or something, if desired.)



> Maybe we could simply add a library for the most common cases of lists that only contain alphanumerical stuff.

I think that makes sense, too. Though I see this as a separate issue. Getting data out of Snap<em>!</em> is still a bit tricky IMO. And a list → JSON block would be fine, but then you have to be able to set that to a variable a export it via the watcher which isn't transparent at all. See the next issue I'm going to write.

bromagosa:
As per https://github.com/jmoenig/Snap/commit/dab03fac5ded5518d726ff716941c6ff377f0193, this is now obsolete :)

cycomachead:
Mostly obsolete, but I think this patch has a couple useful things. :)

But I guess we can ignore XML...

-- 
Michael Ball
From my iPhone
michaelball.co

PR #1674, opened 2017-03-01
Fixing issue translating projectName only in the first language change

jguille2:
Hi again,
## Issue -Current behaviour
  - Starting up Snap! does a global localization (right of course!), also setting default project name (_localize('untitle')_)
  - When I change the language, IDE is translated, but not current project (I think this is fine). For example, sprites names don't change, even if I have the default sprite name (Sprite, Objekt, Objecte...)
  - But there is a curious behaviour with Project Name. In the first change it is translated (untitle, Unbenannt, Sense títol...) and after this (this first change), it is not translated anymore.

## Origin and Proposal
  - The problem is that the startup  is not defining _projectName_. After this, when I change language, _projectName_ is defined.
  - I think the right behaviour is, like sprite names, not translate projectName after startup.
  - My proposal is **to set projectName value in the first writting (in the startup)**.

Thanks,
Joan

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -918,8 +918,9 @@ IDE_Morph.prototype.createControlBar = function () {
             return;
         }
 
+		if (!myself.projectName) myself.projectName = localize('untitled');
         this.label = new StringMorph(
-            (myself.projectName || localize('untitled')) + suffix,
+            myself.projectName + suffix,
             14,
             'sans-serif',
             true,


cycomachead:
Yep, it's a bug and this fix makes sense to me! 👍

jmoenig:
We might already have a fix for this when we pull #1778. So we're going test this and keep this under review. Thank, @jguille2 !

jmoenig:
Oh, we're just noticing that Michael's PR does *not* address this issue. We'll keep this under review. Thanks!

stirlingstout:
I've just noticed this behaviour when playing with the current dev. I'm not sure I can convince myself about the 'right' solution but I favour the principle of least surprise. If I change the language from English to French then I'm not surprised when **Untitled** changes to **Sans titre** (I didn't put it in, Snap! did, so it should/could change it). However, if I then change back to English I am surprised when **Sans titre** doesn't change back to **Untitled**.

Strangely, and somewhat inconsistently, I'm not surprised when **Sprite** and **Stage** don't change in the same way.

Changing from one language to another and back is not something that a lot of people are going to do very often, so the proposed solution would get my vote.

jguille2:
Hi @stirlingstout and @jmoenig,

Yes, this is the behavior that I reported... I've returned to study this (I had forgotten all the details... it's been ten moths!) and I will change my PR... but I need your decisions. I try to explain all the details:

### The origin
  - It's clear. In [the startup](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/bf3eedab751b7f4181ee05e56b89d96878fc2197/gui.js#L228), our project has not a projectName.
  - We see a _localize('untitle')_ but it's only [a label](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/bf3eedab751b7f4181ee05e56b89d96878fc2197/gui.js#L971). ProjectName is still empty.
  - When we change language, to get a complete environment translation, we load again the project using the _project.toXML_ and then, the project [is named](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/bf3eedab751b7f4181ee05e56b89d96878fc2197/store.js#L1626) with the localized 'Untitled'.
  - Now, we have a real _projectName_ and the _untitled_ (or _Unbenannt_, _Sense títol_...) is fixed (now it is not a label for a real untilted project).

### Options
  1. The easier option is **to name the project in the beginning** (with the localized _untitled_). This issue will be fixed... but also it have a drawback. To begin with a unnamed project is interesting because Snap! asks us for a name when when we want to export or save the project.

  2. Other option is **to change _toXML_ function**. We can write _name="@"_ only when project name is defined. It is also easy to fix by this way... but I will have to check if this cause any other problem (having an no named xml).

  3. **To change the language changer** (the [reflectLanguage function](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/bf3eedab751b7f4181ee05e56b89d96878fc2197/gui.js#L4831)). I think is the best option. I have to study it thoroughly, but I think it is not difficult. I can take care for more details to get a simple Language changer (without any other change (not like the [change of tab](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/bf3eedab751b7f4181ee05e56b89d96878fc2197/gui.js#L4847) ...)

Jens, tell me something and I'll remake the PR following your instructions.

Joan

PR #1734, opened 2017-04-23
Add basic logging code from iSnap

thomaswp:
This pull request includes a basic version of the logging code from [iSnap](go.ncsu.edu/isnap). It includes a basic logger class that outputs to the console, logging calls throughout the codebase, and a few additions to the Snap! project data-structure.

To enable logging, simply copy the `logging/config.js.example` file to `logging/config.js`. Then you should see logging output in your console. 

This pull request is mainly intended as an offer to make the logging code official, and a request for feedback on how to make it suitable for the main Snap branch. I'm mainly expecting comments, but if it seems suitable as-is, feel free to merge it in and make your own changes.

Primary features include:
* An "abstract" `logger.js` class and a basic `console-logger.js` implementation that outputs to the console. The logger handles saving snapshots of the code with each log if the code has changed.
* Logging calls throughout the codebase.
* A `logging/config.js.example` file. When copied to `logging/config.js`, this file says what type of logger to use and can also be used to configure the snap cloud URL. If this file is missing or empty, the logger is not created, but no errors will occur.
* Blocks now have unique IDs that are saved/loaded with a project.
* The project XML now contains an <editing> section that includes open editors and their unsaved custom block code. When a project with <editing> block is loaded, these block editor dialogs will be opened as well.

Logs come as JSON objects with a few primary fields:
* `message`: A label for an event which occurred (e.g. "Block.snapped")
* `code`: An XML representation of the project after the event occurred (if changed), with the stage image removed.
* `time`: The time (as a timestamp) when the even occurred.
* `projectID`: A GUID for the project currently being edited
* `userInfo`: An object with info about the current user. It currently contains a `browserID` GUID which is stored as a cookie in the browser and a  `sessionID` GUID for the current browser session. In other settings, this can be augmented with a `userID` if users log in to use Snap!.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -2138,6 +2138,12 @@ SyntaxElementMorph.prototype.endLayout = function () {
 BlockMorph.prototype = new SyntaxElementMorph();
 BlockMorph.prototype.constructor = BlockMorph;
 BlockMorph.uber = SyntaxElementMorph.prototype;
+// ID for the next BlockMorph to be created
+BlockMorph.nextId = 0;
+// Flag for whether copied BlockMorphs should also copy block IDs
+// This should default to false, but is temporarily set to true in special
+// circumstances, such as block relabeling
+BlockMorph.copyIDs = false;
 
 // BlockMorph preferences settings:
 
@@ -2165,6 +2171,7 @@ function BlockMorph() {
 }
 
 BlockMorph.prototype.init = function (silently) {
+    this.id = BlockMorph.nextId++;
     this.selector = null; // name of method to be triggered
     this.blockSpec = ''; // formal description of label and arguments
     this.comment = null; // optional "sticky" comment morph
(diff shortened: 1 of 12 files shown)


blocks.js
```diff
@@ -3334,29 +3398,22 @@ BlockMorph.prototype.eraseHoles = function (context) {
 // BlockMorph highlighting
 
 BlockMorph.prototype.addHighlight = function (oldHighlight) {
```
thomaswp: Not technically related to logging, but this was a nice refactoring I made when building some of my own highlight code, so I threw it in there.

logging/config.js.example
```diff
@@ -0,0 +1,10 @@
+
```
thomaswp: The config file itself isn't versioned so you can have different configs for different deploys.

logging/console-logger.js
```diff
@@ -0,0 +1,16 @@
+// Log to the console
```
thomaswp: I know you like giant files, but for my logging code I've split it up into class-based files. I'd prefer to keep it that way, since that's how my codebase has it, but if it's a problem they could be squashed.

logging/main.js
```diff
@@ -0,0 +1,35 @@
+var Trace;
```
thomaswp: The name "Trace" is a bit of an artifact of my original naming conventions. I recognize now it logs errors and other things, but I haven't had the heart to change the name.

thomaswp (commented):
I added a few explanations to clarify changes.

liam4:
(Fixed link to iSnap: http://go.ncsu.edu/isnap)

PR #1746, opened 2017-05-12
Allow downloading block stacks as XML

bromagosa:


liam4:
Next step – copy to and paste from the clipboard.. :)

bromagosa:
That's much harder because we don't have access to the clipboard, but we could probably have an internal `localStorage`-powered blocks clipboard :)

cycomachead:
We do have access to the clipboard. :)

Well, if users Do Cmd-C / Cmd-V, and we could work on a function that calls it, I believe.

blocks.js
```diff
@@ -2593,6 +2593,22 @@ BlockMorph.prototype.userMenu = function () {
         },
         'open a new window\nwith a picture of this script'
     );
+    menu.addLine();
+    menu.addItem(
+        'download stack',
```
cycomachead: Do we use the term "stack" anywhere? Why not a "script"? (or am I misunderstanding something?)

jmoenig:
Nah, Michael, the infuriating thing about web programming is that there are all kinds of features available to users that cannot be invoked programmatically. Clipboard is just one of them. :-(

cycomachead:
There has been slowly increasing support for clipboard APIs for a while but fairly recently I think things pretty decent. 

https://w3c.github.io/clipboard-apis/

I use multiple websites where this is possible, and AFAIK they aren't using any of the flash based solutions since I've disabled flash. 

-- 
Michael Ball
From my iPhone
michaelball.co

blocks.js
bromagosa: Right, "script" would indeed be better :)

bromagosa: I changed it to "script" in ee126f1 :)

jmoenig: Yay, thanks, Bernat!

jmoenig:
We agreed to pull this one but at first "hide" it so you only see it when you hold the shift key when invoking the context menu. Then we'll play with it and decide when to "unhide" it.
@bromagosa: Can you adjust this PR, and maybe also look into the "receiver" issue in store.js again, because I've changed the structure of scripts to no longer require a receiver.

brianharvey:
[offtopic] Of course we have access to the clipboard, at least for text -- you implemented that so people could paste in their passwords! [/offtopic]

jmoenig:
no, we don't have access to the clipboard. At least not in any programmatical / API way. It's infuriating.

gui.js
```diff
@@ -4086,6 +4089,60 @@ IDE_Morph.prototype.openMediaString = function (str) {
...
+
+    if (Process.prototype.isCatchingErrors) {
+        try {
+            xml = this.serializer.parse(str, this.currentSprite);
+            script = this.serializer.loadScript(xml, this.currentSprite);
+        } catch (err) {
+            this.showMessage('Load failed: ' + err);
+        }
+    } else {
+        xml = this.serializer.loadScript(str, this.currentSprite);
```
dekrain: Should it be `parse`?

PR #1761, opened 2017-06-13
Fix ios ibooks

aidanlane:
Add a margin to the top for iBooks on iPad and iPhone, where a black
bar can sit across the top, blocking access to the widget's controls.


liam4:
Ah, yes, the good old "GitHub automatically selects the upstream repository (e.g. Snap--Build-Your-Own-Blocks or scratch-flash) rather than the repository the code changes are in (e.g. snapapps, scrap-mod). I've had this happen to me before. Fun! :tada:

aidanlane:
Oops, yes, apologies for that!

Issue #1762, opened 2017-06-13
reportURL block does not work when accessing Snap! over https


thomaswp:
If someone wants to use Snap! over `https` (https://snap.berkeley.edu/run), for example to include it in an iframe on a secure webpage, or because they generally like to practice safe internet usage, the `reportURL` block does not work. 

![image](https://user-images.githubusercontent.com/1750176/27086596-a6fe9b38-5020-11e7-9355-cc0c969485b0.png)

This is the case for obvious reasons: the block always uses `http` and modern browsers will not allow insecure AJAX requests on a secure page.

I've submitted a simple fix that simply uses `document.location.protocol` instead of a hardcoded `http:`. A more robust solution would probably give the `reportURL` block a dropdown of protocol choices.

Obviously if the page you want to fetch doesn't support https, you can't get the data when using Snap over http no matter what you do, so there's only so much we can do.

cycomachead:
Are there many pages that _don't_ support https? I kind of think we should just change the fault for every request, but I don't know how likely this is to break anything.

liam4:
> Are there many pages that don't support https?

Typically locally hosted servers use HTTP, not HTTPS – for example you might make a simple Python program that writes files to disk when it gets HTTP requests; it would be inconvenient to have to program in HTTPS support for a simple experiment server like that.

jguille2:
Hi everybody!
I understand this issue (_http_ requests are not allowed under an "_https_ environment") but I think we need a more generic solution (also allowing _https_ requests under a typical "_http_ snap").

Then Yes, we need a dropdown of protocol options.
![blockshttp](https://user-images.githubusercontent.com/2926315/27150819-fa067afc-5148-11e7-99ae-a7269623abad.png)
![httpoptions](https://user-images.githubusercontent.com/2926315/27150827-005e8778-5149-11e7-8217-90e3afe188c1.png)
 It is an easy implementation... but it need to be blessed, because this question was opened (and not closed) in the past (for example [here](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/issues/782))

In the meantime, you can build your own custom block to do this job. If you want, I share [this project](http://snap.berkeley.edu/snapsource/snap.html#present:Username=jguille2&ProjectName=httpRequest) with a functional http/https reporter.
![blockhttps](https://user-images.githubusercontent.com/2926315/27150961-61b6b252-5149-11e7-8d5a-d0822681109b.png)

Joan

[...]

PR #1763, opened 2017-06-13
Allow reportURL to work over https; fixes 1762

thomaswp:
Simple fix for #1762

### DIFF
--- a/objects.js
+++ b/objects.js
@@ -850,7 +850,7 @@ SpriteMorph.prototype.initBlocks = function () {
         reportURL: {
             type: 'reporter',
             category: 'sensing',
-            spec: 'http:// %s',
+            spec: (document.location.protocol || 'http:') + '// %s',
             defaults: ['snap.berkeley.edu']
         },
         reportIsFastTracking: {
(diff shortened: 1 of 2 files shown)


objects.js
```diff
@@ -850,7 +850,7 @@ SpriteMorph.prototype.initBlocks = function () {
         reportURL: {
             type: 'reporter',
             category: 'sensing',
-            spec: 'http:// %s',
+            spec: (document.location.protocol || 'http:') + '// %s',
```
cycomachead: If we pull this, I'd suggest just leading this as plain http.

threads.js
```diff
@@ -2138,7 +2138,8 @@ Process.prototype.reportURL = function (url) {
     var response;
     if (!this.httpRequest) {
         this.httpRequest = new XMLHttpRequest();
-        this.httpRequest.open("GET", 'http://' + url, true);
+        var protocol = document.location.protocol || 'http:';
```
cycomachead: We would probably want to check if this is `file:` and convert that to http. I do think people make http requests from regular file pages. It's not technically the most supported mode of using Snap<em>!</em> but it does happen.

cycomachead:
Closing because of 37dbcad27f03fe9edd8c43875305c3d2c29b59f9

PR #1771, opened 2017-06-23
Use <input> for internal c&p text handling.

cycomachead:
The current version uses <textarea> which means that everything has
autocaps applied, which can make it really annoying to try to login.

There's a minor side effect: copying and pasting outside of Snap!
(on mac OS) doesn't keep any newlines. Newlines are still saved and
rendered OK. This is only a minor annoyance when copying JS code outside
of Snap!

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -5207,12 +5207,16 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
 
 CursorMorph.prototype.initializeClipboardHandler = function () {
     // Add hidden text box for copying and pasting
+    // Use an <input> not a <textarea> to prevent autocaps from taking affect.
     var myself = this,
         wrrld = this.target.world();
 
-    this.clipboardHandler = document.createElement('textarea');
+    this.clipboardHandler = document.createElement('input');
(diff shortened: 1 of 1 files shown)


liam4:
> This is only a minor annoyance when copying JS code outside of Snap!

Ah.. I'd say that's worth investigating, before pulling! Copying JS code and pasting it elsewhere is certainly something I'd like to be able to do, and it's something that I'll be sure to go "why doesn't this just work?" in the future about.

cycomachead:
Well it works. New lines aren’t kept. It’s valid JS so pasting into an editor that formats code is fine. 

Pasting code in with new lines is also fine. 

The problem is that no matter how hard I try, the browser doesn’t respect auto caps rules when using textarea. 

For me the benefit is that no auto-capitalizing every field is a better experience, including writing JS functions. 

This seems to be a mostly Safari and iOS problem but that is a reasonably sized group. 

-- 
Michael Ball
From my iPhone
michaelball.co

cycomachead:
The recent tweaks for better internationalization have helped here so I don't think this is necessary anymore.

PR #1779, opened 2017-06-28
Allow script pic with result for all blocks. 

### LINKED ISSUE #1777: No script pics for error reports?
We want to be able to teach how to use error reports, but currently, we can't get a script pic of one being reported. Can we have that feature?

cycomachead:
Fix #1777

If you have an error w/ a command block, you can now get a script pic:

However, it won't show the whole block stack. And if no result is returned, you don't get any picture back. This seems like less than optimal behavior, but since this is a shift-click option, I think it's OK.

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -2686,17 +2686,15 @@ BlockMorph.prototype.userMenu = function () {
     );
     if (shiftClicked) {
         top = this.topBlock();
-        if (top instanceof ReporterBlockMorph) {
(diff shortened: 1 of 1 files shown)


cycomachead:
Note - I can't yet make it show the entire script w/o tweaking how we display errors in general. This will require a bit of thought. :/

cycomachead:
Closing b/c these was solved at some point. :)

PR #1781, opened 2017-06-28
Show the block when there is an inside error

cycomachead:
If an error occurs within a block, this will show the block inside the error message. 

errors

It's only the block that generated the error - no stack trace. Yet! (though that should probably be a different function.)


brianharvey:
I love it!

Your example reminds me: Could we please change "list.at is not a function" to "ITEM of non-list"?

jmoenig:
yes, this is really nice, Michael. Thank you!

cycomachead:
> Your example reminds me: Could we please change "list.at is not a function" to "ITEM of non-list"?

Yeah, there's a couple existing PRs for that.

#1629

cycomachead:
Got the proper image sizing working, so this should be good. :)

cycomachead:
@jmoenig the conflicts here should be easy to fix. I can do that very soon -- Is it possible to get this in the 5.0 release?

cycomachead:
@jmoenig I resolved the conflicts; I still think this would be a great addition to the 5.0 release. 😄

bromagosa:
I think so too! :pray:

brianharvey:
We kind of declared a feature freeze last week.  (We did a month ago, too, but we really meant it this time, and have hardly added anything since.)  For what it's worth, I think it'd be great, but I can see how if I were Jens I'd be nervous about some potential edge case.

On the other hand there are about a thousand missing or obsolete help screens, so I'll be busy for a while...

cycomachead:
We've added a bunch of stuff since then, it'd be nice to get this in.

(Also there's always potential edge cases, but given this only shows 1 block, I think it's safe.)

bromagosa:
I agree again that this would be very useful.

brianharvey:
Jens?

cycomachead:
@jmoenig Can we talk a look at this again? Thanks! 🙂 

---

I made a slight tweak to this since last time, since using `block.fullImage()` on a command block isn't very desirable as it will show the rest of the script. In this way, the goal is to only show the "first" block, but with inputs that cased the error. In the 99% case the inputs _are_ useful, because they help you identify exactly which instead of the block caused the error. (Think about how often you might use a block like item of or length or such.)

Also, this now *always* shows the block that errors, not only when the the error happens inside a block. The reason for this is that in a long script, seeing the example block will help identify the  error, and in some cases, the error can occur "beneath" the bottom of the screen when the script is long. This definitely happens. Even when the error is the second block in a script, the block image is more clear than the the little bubble arrow. 

I'm not sure if the error image is useful when there's only 1 block or it's the _first_ in the stack, but I don't think it's harmful. If others disagree, that's an easy case to fix.

Here's some more examples:

Screen Shot 2019-08-03 at 11 39 55 PM

Screen Shot 2019-08-03 at 10 58 41 PM

Screen Shot 2019-08-03 at 6 19 35 PM

brianharvey:
By the way, I've never really understood the "Inside: Error:" text.  Does it mean "inside this custom block is the one causing the error"?  I think that would be clearer like this:

Error: Too many inputs
in block:  (picture of the block causing the error)
inside toplevel block:  (picture...)

cycomachead:
It means it's inside some custom block. 

I agree it's not great... Your suggestion is better, though somewhat more complex. But essentially, amounts to the fact that we really do just need stack traces when errors happen deep inside custom blocks. I think that's a separate function. 

"Top Level" is a little hard to define because is the top block the immediate parent in the stack trace or the top of the calling script.

brianharvey:
"Top level" means a script in the scripting area, not a custom block body.

I think a minimal change that would clarify those messages would be to replace "Inside:" with "Inside a custom block:\n".

bromagosa:
I love this and I think it'd be very useful. I definitely have lost quite a lot of time hunting for "inside errors".

cycomachead:
Added the enhanced text to error message. 

Also adding localization.

cycomachead:
Closing in favor of #2857

PR #1810, opened 2017-07-20
Sound envelope addition/click reduction

micahcowan:
I'm not sure it 100% eliminates clicking at note start/stop - listening very carefully, I believe I can still sometimes hear it... might be timing limitations in the browser implementation of the Web Audio API. I've also encountered some inconsistencies in how "clicky" very fast notes can sound (eighth notes, sixteenth notes, etc).

Still, seems to be a vast improvement on the current noisy clicking sounds found at the starts and, especially, stops, of played notes.

This is my first foray into the Web Audio API, and in particular I don't know the relative portability of various facilities in the Web Audio API. It might be well to add some safety checks, or wrap things in try-blocks, and fall back to hard-setting the gain value as before.

### DIFF
--- a/objects.js
+++ b/objects.js
@@ -7553,7 +7553,6 @@ Sound.prototype.toDataURL = function () {
 function Note(pitch) {
     this.pitch = pitch === 0 ? 0 : pitch || 69;
     this.setupContext();
-    this.oscillator = null;
 }
 
 // Note shared properties
@@ -7581,58 +7580,52 @@ Note.prototype.setupContext = function () {
         throw new Error('Web Audio API is not supported\nin this browser');
     }
     Note.prototype.audioContext = new AudioContext();
-    Note.prototype.gainNode = Note.prototype.audioContext.createGain();
     Note.prototype.volumeLevel = 0.25;
-    Note.prototype.attackTime = 0.075;
-    Note.prototype.releaseTime = 0.075;
+    Note.prototype.releaseTime = 0.1;
 };
 
 // Note playing
 
 Note.prototype.play = function (secs) {
-    var attackStart, attackEnd, releaseStart, releaseEnd;
-    this.oscillator = this.audioContext.createOscillator();
-    if (!this.oscillator.start) {
-        this.oscillator.start = this.oscillator.noteOn;
+    var attackStart, releaseStart, releaseEnd;
+    var oscillator = this.audioContext.createOscillator();
+    var gainNode = this.audioContext.createGain();
+
+    if (!oscillator.start) {
+        oscillator.start = oscillator.noteOn;
     }
-    if (!this.oscillator.stop) {
-        this.oscillator.stop = this.oscillator.noteOff;
-    }
-    this.oscillator.type = 'sine';
-    this.oscillator.frequency.value =
+    oscillator.type = 'sine';
+    oscillator.frequency.value =
         Math.pow(2, (this.pitch - 69) / 12) * 440;
-    this.oscillator.connect(this.gainNode);
-    this.gainNode.connect(this.audioContext.destination);
-    this.gainNode.gain.value = 0;
+    oscillator.connect(gainNode);
+    gainNode.connect(this.audioContext.destination);
+    gainNode.gain.value = 0;
 
-    // attack:
+    // attack (instant, no ramp):
     attackStart = this.audioContext.currentTime;
-    if (this.attackTime < (secs / 2)) {
-        attackEnd = attackStart + this.attackTime;
-    } else {
-        attackEnd = attackStart + (secs / 2);
-    }
-    this.gainNode.gain.cancelScheduledValues(attackStart);
-    this.gainNode.gain.setValueAtTime(0, attackStart);
-    this.gainNode.gain.linearRampToValueAtTime(this.volumeLevel, attackEnd);
+    gainNode.gain.setValueAtTime(this.volumeLevel, attackStart);
+
     // release:
-    releaseEnd = attackStart + secs;
-    if (this.releaseTime < (secs / 2)) {
-        releaseStart = releaseEnd - this.releaseTime;
-    } else {
-        releaseStart = releaseEnd - (secs / 2);
-    }
-    this.gainNode.gain.setValueAtTime(this.volumeLevel, releaseStart);
-    this.gainNode.gain.linearRampToValueAtTime(0, releaseEnd);
+    releaseStart = attackStart + secs;
+    releaseEnd = releaseStart + this.releaseTime;
+    gainNode.gain.setValueAtTime(this.volumeLevel, releaseStart);
+    gainNode.gain.linearRampToValueAtTime(0, releaseEnd);
 
-    this.oscillator.start(0);
-};
+    oscillator.start(0);
 
-Note.prototype.stop = function () {
-    if (this.oscillator) {
-        this.oscillator.stop(0);
-        this.oscillator = null;
-    }
+    // resource deallocation closure
+    var _this = this;
+    var dealloc = function () {
+        if (gainNode.gain.value == 0) {
+            gainNode.disconnect();
+            oscillator.disconnect();
+        }
+        else {
+            // Release somehow hasn't finished. Check again in a second.
+            setTimeout(dealloc, 1000);
+        }
+    };
+    setTimeout(dealloc, (secs + (_this.releaseTime * 2)) * 1000);
 };
 
 // CellMorph //////////////////////////////////////////////////////////
--- a/objects.js
+++ b/objects.js
@@ -7582,7 +7582,9 @@ Note.prototype.setupContext = function () {
     }
     Note.prototype.audioContext = new AudioContext();
     Note.prototype.gainNode = Note.prototype.audioContext.createGain();
-    Note.prototype.gainNode.gain.value = 0.25; // reduce volume by 1/4
+    Note.prototype.volumeLevel = 0.25;
+    Note.prototype.attackTime = 0.075;
+    Note.prototype.releaseTime = 0.075;
 };
 
 // Note playing
@@ -7600,7 +7602,10 @@ Note.prototype.play = function () {
         Math.pow(2, (this.pitch - 69) / 12) * 440;
     this.oscillator.connect(this.gainNode);
     this.gainNode.connect(this.audioContext.destination);
+    this.gainNode.gain.value = 0;
+    this.gainNode.gain.linearRampToValueAtTime(this.volumeLevel, this.audioContext.currentTime + this.attackTime);
     this.oscillator.start(0);
+    this.released = false;
 };
 
 Note.prototype.stop = function () {
@@ -7609,6 +7614,13 @@ Note.prototype.stop = function () {
         this.oscillator = null;
     }
 };
+
+Note.prototype.release = function () {
+    if (!this.released) {
+        this.released = true;
+        this.gainNode.gain.linearRampToValueAtTime(0, this.audioContext.currentTime + this.releaseTime);
+    }
+}
 
 // CellMorph //////////////////////////////////////////////////////////
 
--- a/objects.js
+++ b/objects.js
@@ -7626,7 +7626,6 @@ Note.prototype.play = function (secs) {
     this.gainNode.gain.linearRampToValueAtTime(0, releaseEnd);
 
     this.oscillator.start(0);
-    this.released = false;
 };
 
 Note.prototype.stop = function () {
--- a/objects.js
+++ b/objects.js
@@ -7590,6 +7590,7 @@ Note.prototype.play = function (secs) {
     var attackStart, releaseStart, releaseEnd;
     var oscillator = this.audioContext.createOscillator();
     var gainNode = this.audioContext.createGain();
+    this.gainNode = gainNode;
 
     if (!oscillator.start) {
         oscillator.start = oscillator.noteOn;
@@ -7614,7 +7615,7 @@ Note.prototype.play = function (secs) {
     oscillator.start(0);
 
     // resource deallocation closure
-    var _this = this;
+    var releaseTime = this.releaseTime;
     var dealloc = function () {
         if (gainNode.gain.value == 0) {
             gainNode.disconnect();
@@ -7625,8 +7626,16 @@ Note.prototype.play = function (secs) {
             setTimeout(dealloc, 1000);
         }
     };
-    setTimeout(dealloc, (secs + (_this.releaseTime * 2)) * 1000);
+    setTimeout(dealloc, (secs + (releaseTime * 2)) * 1000);
 };
+
+Note.prototype.stop = function () {
+    var currentTime = this.gainNode.context.currentTime;
+    var releaseEnd = currentTime + this.releaseTime;
+    this.gainNode.gain.cancelScheduledValues(currentTime);
+    this.gainNode.gain.setValueAtTime(this.volumeLevel, currentTime);
+    this.gainNode.gain.linearRampToValueAtTime(0, releaseEnd);
+}
 
 // CellMorph //////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


liam4:
I made a comparison video: https://streamable.com/177d3

micahcowan:
Thanks very much for that, @liam4!

You may notice that the notes sound somewhat less contiguous now. This could possibly be mitigated by reusing an oscillator for contiguous notes, and shortening the envelope. I _suspect_ that this would cause the note to shift frequency without jumping to a different place in the waveform (the current method of creating new oscillators, jumps to the start of the waveform at each new note, which exacerbates the clicking noise).

I suspect that might not be guaranteed behavior, if it's even true, and it probably wouldn't reduce the clicking as much as the current method. Note, too, that the breaks between notes exemplified in @liam4's stream, are somewhat less noticeable in longer-lasting notes (these ones are being played at 1/5 of a beat). And, of course, it's a lot more work to try to make that happen.

The biggest difference in the character of the sound, is due especially to the note attacks also being ramped now, instead of jumping in; the release probably has less of a difference in how the note "feels". My primary focus was on eliminating the clicks to as great a degree as possible, and something around 3/4 of 1/10 of a second seemed to do the trick (which is what it's currently set to: linear ramp up for 0.075 seconds, and ramp down at 0.075 seconds of the end - if the notes are shorter than 0.15 seconds (as the stream example notes probably are?), it just ramps until the note's halfway done, and then ramps back down from that point.

However, if the character of the sound has been changed too much, perhaps a compromise can be found, with shorter envelopes that eliminate much or most of the clicking, but not all of it. I suspect we can afford to lose more time from the attack than the release, given that notes seem to start with the waveform zeroed (due to new oscillator nodes), and we're using a gentle sine waveform; but I still heard the clicks sometimes at much less than 0.075 seconds ramping up.

jmoenig:
So, this is really nice for slow melodies, @micahcowan , but when I test this with some of my more complex musical projects that have a lot of polyphony, harmony and quick short notes in them, it becomes way less precise and the clicking sounds come back with a vengeance. Seems we'll have to live with one shortcoming over another :-/

micahcowan:
Yeah, I confess I'm less than satisfied with my proposed solution, myself. In particular, the note attacks are audibly long, and change the character of the sound. Since most of the clicks don't come from the attacks (though higher frequency probably produces more clicks at starts), it should probably be removed.

You may find that quick notes will have better luck if you remove the attacks, but keep the ramped releases - the sound will feel more "precise", but it should still eliminate much of the clicking. Perhaps you could try a version with the attack portion removed (just setting the gain immediately to 0.25), and see whether that improves your experience.

However, if we really want to emulate the expected way a synthesizer or keyboard works, the release shouldn't actually be coming as part of the note itself (and thereby limited by the note length), but should begin upon note release. The behavior when another note is played after the previous has ended, would depend on whether we want to pretend each thread is a monophonic synthesizer (in which case, the note should jump immediately to the new frequency, on the same oscillator so as to maintain its phase, and not release), or a polyphonic synthesizer (in which case, the first note should be sounding its release simultaneously with the attack of the second note). In either case, this means sound synthesis would effectively need its own separate thread apart from execution, so as to maintain the note release while program execution is continuing.

This of course is a fundamental change to how the music playing works, but then there's not very much to it right now, so perhaps not all that drastic.

Also, @jmoenig, can you link to some of these more complex musical projects? I was primarily using the simple examples @maryfries provided, plus a simple repeated-same-quick-note test.

micahcowan:
It also occurs to me that we could allow players to adjust note envelopes (including attacks and releases), and maybe while we're at it the waveform too, so that they can make their own judgments regarding these trade-offs (this would still really only work if we did releases after the note ends, and not as part of the note itself). We could allow them to change these settings globally, similar to "set tempo", or if we supported a thread-local "play notes [notes] with instrument [instrument-description]", we could conceivably support a rather more organ-like range of sounds.

Of course, perhaps this represents serious feature-creep. Would you support a feature like this @jmoenig (if you didn't have to implement it yourself, of course)?

Alternatively, we could provide low-level access to the Web Audio API and let people roll their own. I suppose this could be done via a library rather than built into Snap!. Though it would surely require turbo/warp-level timing control to function well.

micahcowan:
Worth noting, I only really get clicks on note _starts_ (as opposed to ends) during polyphony. I suspect it's because, with a gain of 0.25, four simultaneous notes can saturate the amplitude, and cause mild distortion. So the attack probably isn't needed at all (at least for the current sine waveforms).

Playing around more with my implementation, I was able to reproduce what you're talking about wrt less precision. I wasn't getting it earlier with the longer attack/release times, but I modified the code to base the attack and release times on a certain number of wave cycles, instead of a static fraction of a second (to minimize impact on the sound itself), and the shorter these attack/releases were, the  more I noticed weird delays: notes lasting too long and then abbreviating the next note, things like that. Timing issues. It improved with turbo mode, but still not fantastic.

My guess is that using these ramped values risks making the Web Audio browser facilities work too hard, and in combination with all the other things that _Snap!_ does every tick, it's sometimes just too taxing. I'm not sure why the size of the ramp would matter so much, though: seems like the calculations should be the same regardless. Perhaps there's an overhead in initializing timed value changes, and if the timing is short enough, the timed event ends before the audio system is ready for it? Setting attack/release times of a single cycle ((1 / frequency) seconds) produces extremely severe timing inaccuracies in the notes.

I'm guessing your more complex polyphony examples had greater polyphony than the ~4 that @maryfries's example used, and that this too taxed the system sufficiently to produce the same effects I'm seeing.

In light of these results, I suppose this project must be abandoned for now, until such time that most peoples' browsers handle Web Audio more efficiently, or _Snap!_ finds ways to drastically reduce its overhead every thread tick.

micahcowan:
Hm. If we could find a way to reuse the oscillator between note changes, keeping the phase synchronized with legato note start/stops, we could at least reduce the occurrence of some of the clicks. We'd probably have to avoid even calling `Note.stop` and `.start`, though, and I imagine that could take some fancy maneuvering, or (again) require a separate thread for note control.

It might also be conceivable we could use envelopes without these gain value ramps, by using oscillators to manage them instead. Not sure I could figure that out, though.

brianharvey:
Have you experimented with the idea of having the fade of a note overlap the next note?  That sounds more realistic with how pianos, at least, actually work.  In the above conversation I see reports of many experiments (for which thank you!) but not that one.  If it works, that overlap should make it okay to do the simple thing of one note per oscillator.

micahcowan:
@brianharvey, I did actually suggest that, but it was hidden in one of several unnecessarily verbose suggestions, so I'm not surprised you didn't see it ^_^

In that comment, I claimed that it would require a separate thread for managing music, but I think that incorrectly relies on assumptions tied to how music currently works in _Snap!_

The release can be scheduled, as can the note stop. All that really needs to happen to support this, is allowing new notes to create new oscillators while old ones are still playing, and manage disconnection of oscillators upon notes ending (which can be scheduled with the oscillator's `.onEnded()` method). So yeah, we should do this.

It does seem like the linear ramping feature used for the release incurs some performance penalty... at least it does when used for brief attacks (as noted above). Hopefully if we just do it on release, it won't be enough to mess with timing. Definitely worth a try!

micahcowan:
Okay, I'm really happy with this commit. This has the note releases hanging over the starts of new notes. I've removed all note attacks, it doesn't seem to affect the clicking much (except when polyphony is high enough to risk a little distortion). It sounds way smoother than current **master**, and the timing appears to be stable, even with significant polyphony and fast notes (provided, of course, that turbo or warp is enabled - current **master** requires this as well)

I used [this project](http://snap.berkeley.edu/snapsource/snap.html#present:Username=micahcowan&ProjectName=Micah%20music%20test) to test with. It does some snazzy chords from the original problem report, runs some polyphonic scales at increasing speeds, and the first section of the Bach Menuet from the examples on the _Snap!_ front page.

brianharvey:
Thanks, that will help with a lot of complaints I think.

micahcowan:
I noticed while inspecting some resulting audio files, that the audio bandwidth consumed by consecutive notes is increased, since of course the volume of the fadeout overlaps the next note, meaning that when the new note starts, the combined waveforms approach the distortion limits more than they did before.

I expect this means that successive chords can handle less polyphony than it could before, and we might want to reduce the default gain from 0.25 to 0.20 or 0.15 (which might be a good idea anyway, since even the original version at 0.25 could begin distorting at more than four simultaneous notes)

micahcowan:
Here's some audio demonstration of polyphonic scales (sorry, no snazzy video like @towerofnix's - I can record audio, or desktop video, but apparently not together at the same time).

* [Original _Snap!_](http://micah.cowan.name/tmp/snap-orig.mp3)
* [Improved version](http://micah.cowan.name/tmp/snap-improved.mp3)
* [FULL demo played through _Snap!_ with the improvements](http://micah.cowan.name/tmp/snap-full-demo.mp3)&#x2a;

These scales are part of the larger [audio demo project](http://snap.berkeley.edu/snapsource/snap.html#present:Username=micahcowan&ProjectName=Micah%20music%20test) I linked above.

&#x2a; The full demo recording did get the default volume adjusted down to 0.15, and you can still hear mild distortion in it. I'm not sure why - when I'm _not_ recording using Audio Hijack, I don't hear any distortion whatsoever at the default volume of 0.25.

micahcowan:
BTW, I've got no problems doing a `git rebase -i` to squash all these "rewrite" commits into a single commit in a new pull request for the purposes of repo cleanliness, if you like - there's just no point until we reach a point where everyone likes the solution. :)

towerofnix:
[It's beautiful.](https://streamable.com/v7a6m) (Unfortunately, my video isn't so beautiful; for whatever reason it recorded at an incredibly low volume! - So turn your speakers up.)

Another thing I found nice - if you stop a project in the middle of a note, the note ends very smoothly, rather than with the clicking sound.

micahcowan:
Yeah, I had a harder time hearing the clicks in that demo than when I play it on my computer, even with the volume turned all the way up. But they're definitely still there! Thanks for doing that @towerofnix ^_^

> Another thing I found nice - if you stop a project in the middle of a note, the note ends very smoothly, rather than with the clicking sound.

Yeah. That's basically because nothing actually tells notes to stop, ever, anymore. This could be a "misfeature", actually, if someone issues a ridiculously long note or chord, it _will_ play to completion, unless you actually close that tab or something. Previously it would be aborted instantly I think.

If that winds up being a legitimate problem, there are ways around it, but it wasn't clear to me that it was worth it. We can't tie the oscillator and gain nodes to the `Note` type anymore, because those `Notes` are sure to go away before the oscillator should stop sounding, leading us back to the original problem. We could have a top-level object (outside all threads) containing references to all currently playing oscsillator and gain nodes (including those that are "done", but still sounding their note release), strictly for the purpose for telling all of them to do a hard disconnect, but it'd have to be random-access (since order of note creation doesn't imply order of note deletion, and notes that are completely done playing must remove themselves from the top-level object), so I'm thinking it'd be something like a dictionary-style object with random, and guaranteed-unique, keys. The deallocator closure I added for the nodes would know its own key, and remove itself from the top-level object...

As I say, not sure it's worth the bother. Guess we'll know if either Jens insists on it, or a user complains that they played a 500-beat note and couldn't get it to stop... 😆

brianharvey:
Umm.  That will for sure happen; you must not know any eight-year-olds.  I'd be _much_ happier if clicking the stop sign stopped all sounds.  I think I vaguely remember from Data Structures class that what you want for the set of oscillators is a heap (the CS2 kind, not the opposite of the stack).  But I could easily be wrong.  The idea would be that each sound would compute when it's supposed to stop, and would insert itself in the heap in the right place in O(log n) time, and when its time came, it would be on top of the heap, and the redisplay process could pop it off.  So sounds wouldn't have to find themselves when they finish.

Anyway, there's a "stop all sounds" block, so we need that capability regardless of my desire for the stop sign to do it.

While I'm making trouble, I would really love a mute button (speaker with a slash through it, which seems not to be in Unicode for some reason) up on the tool bar.

cycomachead:
I agree we need Stop to stop long playing notes -- otherwise lab may become a nightmare. (Not sure whether audio files are separate or not, I would a assume so, but just wanted to point that out.)

Re. Exposing detailed JS properties. 
I'm personally all for this. I don't know they should be build into Snap<em>!</em> by default, but I certainly know people (especially musically inclined folk) who would appreciate a library that lets them tweak things to no end.

towerofnix:
> so I'm thinking it'd be something like a dictionary-style object with random, and guaranteed-unique, keys.

Sorry to stray off-topic, but have you ever heard of [Symbols](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Symbol)? They're practically exactly what you propose. (Though they're a relatively new and not-so-well-supported JS feature; they probably wouldn't fit well into Snap<i>!</I>'s own codebase.)

micahcowan:
> Sorry to stray off-topic, but have you ever heard of Symbols? They're practically exactly what you propose. (Though they're a relatively new and not-so-well-supported JS feature; they probably wouldn't fit well into Snap!'s own codebase.)

Not off-topic at all, as you say, they're perfect for this sort of situation. But also as you say, they're not sufficiently portable.

However, we could probably do just as well by having a global index that's incremented every time it's used. No need for randomness and uniqueness-checking, there.

>I think I vaguely remember from Data Structures class that what you want for the set of oscillators is a heap (the CS2 kind, not the opposite of the stack). But I could easily be wrong. The idea would be that each sound would compute when it's supposed to stop, and would insert itself in the heap in the right place in O(log n) time, and when its time came, it would be on top of the heap, and the redisplay process could pop it off. So sounds wouldn't have to find themselves when they finish.

Yeah, that could be a good idea. Of course, removing oneself from a dictionary is likely even faster than O(log N) time (I'd expect something near O(1)), but having a single place be responsible for all removal might provide advantages over each note spawning its own cleanup, as I currently have it. We might want to run it less frequently than every frame, though, as otherwise we're liable to check for removals much more frequently than we perform them.

<hr/>

However, I just realized that it can be much simpler than either of these solutions - there's a straightforward way to access all of the inputs of an AudioContext's destination node, so we can just iterate over all of those and detach them all, for hard stops.

micahcowan:
I've recently realized that there's a difference in behavior for pauses, as well. In current release, if you hit the pause button, or change tabs, the currently-playing note will hang on indefinitely until you resume, which can be fairly annoying. In my version, the note is already pre-programmed to die at the appropriate time, so no hangers-on, regardless of whether you've clicked pause or changed tabs (I prefer the behavior of my version).

We could consider having pause suspend the note altogether, and then reestablish it again on resume, but this requires some effort, particularly if we want to avoid clicks at start and stop again; and I don't believe it's likely to be worth it. And, as I say, the current behavior is already an improvement (IMO) over **master**.

brianharvey:
I don't understand the "change tabs" part.

micahcowan:
> I don't understand the "change tabs" part.

Browser tabs. If I click on a different browser tab than the one _Snap!_ resides in, the behavior is the same as if I press the "pause" button, at least as far as music is concerned (currently, both in my pull request and in **master**) .

cycomachead:
In that case, I agree with you this is an improvement. (Though we should probably create a new issue tfor the desired behavior.)

brianharvey:
Ah, sorry.  I was thinking of the scripts/costumes/sounds tabs, and couldn't make sense of it that way!

micahcowan:
Okay, so I think it was _way_ easier than I imagined it was.

So, I was wrong about the `AudioDestinationNode` being able to list all of its inputs. Apparently it can give you how many there are, but not access to them.

However, all currently-sounded notes fall into one of only two categories: notes that are "currently playing", and notes that have "stopped", but are currently releasing.

Notes that have "stopped" may safely be ignored when the program has ended: they will only be sounding for another tenth-of-a-second or what have you, and have effectively already been ended.

Notes that are still actively playing, are accessible to the context as `this.activeNote`. So, all I have to do is put back a method in the `Note` class, allowing a thread context to cancel the note, and use it. So I've done that. The only difference is that there's no longer an explicit call to `Note.stop()` when the note has reached its end - it's just there to end the note prematurely, before it's scheduled to end.

PR #1812, opened 2017-07-25
Undo and collaborative editing support

brollb:
Collaborative editing options will only show if the server supports it (ie, has a websocket connection that can be used for the collaborative editing).

For an example of a server that can be used for the collaborative editing features, check out https://github.com/NetsBlox/collaboration


brollb:
I am going to be moving these changes to a different branch in our fork and changing the master branch to include networking capabilities and other additional capabilities that we have been experimenting with in Snap!. 

I will be closing this PR and opening a new one from this new branch!

PR #1851, opened 2017-09-05
Webcam dialog

bromagosa:


brianharvey:
Wow, I didn't realize this was on the list!  Pretty cool.  Do we really want a new sprite rather than a new costume?  (Or are there two camera buttons, one over the sprite corral and one at the top of the Costumes tab?)

bromagosa:
Yep, there are two of them, just where they used to be in Scratch 1.4 and BYOB :)

gui.js
```diff
@@ -1480,6 +1481,41 @@ IDE_Morph.prototype.createCorralBar = function () {
...
+    cambutton.contrast = this.buttonContrast;
+    cambutton.drawNew();
+    cambutton.hint = "take a camera snapshot and\nimport it as a new sprite";
+    cambutton.fixLayout();
+    cambutton.setCenter(this.corralBar.center());
+    cambutton.setLeft(
+        this.corralBar.left() + padding + newbutton.width() + padding + paintbutton.width() + padding
+    );
+
+    if (location.protocol === 'http:') {
```
cycomachead: Does this work OK with file:// URLs?

Also, not a blocker but just curious -- do you know if the same restriction applies to Electron / nw.js apps?

gui.js
```diff
@@ -1480,6 +1481,41 @@ IDE_Morph.prototype.createCorralBar = function () {
...
+        cambutton.hint = 'Due to browser security policies, you need to\n' +
```
cycomachead: Should we maybe name this 'camera requires https' or something and stuff it in a translation file? (Would make it easier to translate, and reduces a bit of duplication below).

objects.js
```diff
@@ -9524,4 +9525,132 @@ StagePrompterMorph.prototype.mouseClickLeft = function () {
...
+    var myself = this,
+        stage = this.ide.stage;
+
+    this.videoElement = document.createElement('video');
+    this.videoElement.hidden = true;
+    document.body.appendChild(this.videoElement);
+
+    if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
+        navigator.mediaDevices.getUserMedia({ video: true })
+            .then(function (stream) {
```
cycomachead: You've probably checked, but `navigator.mediaDevices.getUserMedia` is always defined in browsers that support promises?

This effectively blocks all IE's, and requires Chrome and safari w/in the last 3 years. That's probably not too big a deal -- but if `navigator.mediaDevices.getUserMedia` is defined and promises are not then, this seems like a bad situation.

gui.js
jmoenig: this is an excellent, idea, @cycomachead !
@bromagosa , I can take care of this after I pull it, so don't worry about it. I'd like to pull this when we reach consensus about "file:" protocol  and Michael's other remarks. What do you think?

objects.js
```diff
@@ -9524,4 +9525,132 @@ StagePrompterMorph.prototype.mouseClickLeft = function () {
 
 StagePrompterMorph.prototype.accept = function () {
     this.isDone = true;
+}
+
+// CamSnapshotDialogMorph /////////////////////////////////////////////////////////
```
jmoenig: @bromagosa any particular reason why the CamSnapshotDialogMorph should be in ```objects.js``` rather than in ```gui.js```? Not that it means much. If it's okay with you I'd probably move it over myself... (?) Tnx!

gui.js
bromagosa: The restriction does not apply for file:// URLs, but accessing the camera from these [won't work anyway](https://stackoverflow.com/questions/17682978/getting-not-allowed-to-load-local-resource-error-while-trying-to-attach-a-medi), it will give out this error:

`Not allowed to load local resource: blob:null/e90fa29d-f409-442a-94ba-98fa9999eda1`

This works great in nw.js apps, I'm using this very same code in Snap4Arduino :)

gui.js
bromagosa: Cool! Thank you, Jens! :)

But we do not have an English translation file, do we?

jmoenig: Haha, in fact, we do have an English translation file for this exact purpose :-)

objects.js
bromagosa: Thing is, `mediadevices.getUserMedia` returns a Promise, [as defined by the standard](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia). So, nope, IE users are going to be out of luck :(

Should we gray out the button for this case too?

gui.js
bromagosa: Oh wow, I didn't know we did! I can't find it anywhere, what's it called?

objects.js
cycomachead: IMO, yes. Hopefully, `mediadevices.getUserMedia` is undefined in those case, and no exception would be raised, so it wouldn't be a huge deal, but a grey button is less confusing.

bromagosa: For what it's worth, IE doesn't define `mediaDevices.getUserMedia` either.

Why is anybody using that piece of garbage?

Should we gray out the button for IE users? We would need to show a different hint though.

gui.js
jmoenig: It's inside ```locale.js```. But nevermind, I'll take care of it!

objects.js
bromagosa: Hmm, no reason, really. I guess because it was about costumes, but it makes sense to have it in gui.js :)

gui.js
cycomachead: Good to know about the desktop stuff. That post seems to suggest HTTP is OK with a localhost server? :O That doesn't make sense to me, but I wonder if we can support that. 

Sigh, I should just test my auto-updater app some more. Then we wouldn't have to worry about offline stuff.

bromagosa: Exactly, there's no issue accessing it locally with a webserver.

cycomachead: In that case, this is wrong to disable the button, because you're not going to be using https, right?

bromagosa: Ouch, you're right! I always run my server with https, but you're totally right.

Now there's a problem. How do we detect if the user is running http locally?

cycomachead: I was going to write that last line, and the realized there's not a good way. :/

We could just check if the location is `localhost` or `127.0.0.1` or any of the few others, but that seems annoying. 

Perhaps what we should do is just try to create a video stream, (inside a try/catch) and store whether or not it's successful? That would solve the IE disabling issue too. 

`IDE_Morph.prototype.videoCaptureAvaialbe = function () {...}`

I don't know if there's a performance hit to doing that or if we can cache the video element to be used if you do open the dialog?

Actually, this might be a better check anyway because you might be on a device without a webcam.

objects.js
cycomachead: Ah true. A different hint would be helpful. We could just add a line or two to the same hint and just say "This doesn't work in IE, and you need a web cam". It semi-publicly shames internet explorer which seems OK. 

Hopefully people are just using Edge now.

PR #1863, opened 2017-09-16
TypeScript

as-com:
**Note:** This pull request is _not_ meant to be merged right now; it is merely here to start a discussion on the scalability of Snap<i>!</i> development.

So I was looking at the source code of Snap, and I noticed that there are many areas where it could be improved. Here's what this pull request has accomplished so far:

- Machine-translated the ES5-style source code into clean(er) ES6 syntax (with classes, const/let, arrow functions, etc.)
- Broke down the massive JavaScript files into smaller files, with one class per file (much like Java)
- Turned all of the JavaScript files into [TypeScript](https://www.typescriptlang.org)
- Introduced the TypeScript compiler and Webpack as a build tool
- Made a more suitable .gitignore file

Before you run away screaming, please note the advantages of this slightly more complicated setup:

- **Code is easier to navigate** - smaller files, less scrolling
- **Editors crash less** - some editors really don't like navigating/parsing/refactoring files that have >12,000 lines of code each
- **Development is less error-prone** - TypeScript's type annotations (not added in this pull request yet) can catch many of the errors commonly made in JavaScript programs
- **Better autocomplete** - Many editors support TypeScript (or have plugins for it) and can display accurate autocompletion options when writing code
- **Dependencies are managed automatically** - ES6 imports allow the build tool to automatically put the scripts in the right order
- **Faster to load** - Webpack compiles everything into a single compressed JavaScript bundle, which is faster to load than loading each script individually
- **Potential performance improvements** - I noticed that a lot of Snap<i>!</i>'s inheritance code is written in a rather strange manner (such as setting the prototype to a new parent object), which is not what modern browsers expect. The code TypeScript emits from the class constructs is typically code that JavaScript engines can optimize better. Also, JS engines don't like [call-site polymorphism](http://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html), and TypeScript can help identify these bottlenecks.

Of course, there are some disadvantages of refactoring the code like this:

- People need to have Node.js installed to build Snap<i>!</i>
- TypeScript's type syntax also needs to be learned ([but it's still pretty easy](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html))
- Build step required

I created this pull request to start a discussion; it is not meant to be merged (yet). The code was semi-automatically refactored into TypeScript-ish, and does not currently compile (a lot of imports are missing right now) nor does it function properly (someone more familiar with the Snap<i>!</i> codebase might do a better job in refactoring the code into more idiomatic TypeScript).


carloslfu:
Is such a nice idea, is a lot of work that worth doing. I have been using Snap 3 Years ago. Now I use Typescript / Webpack for personal and business projects and I can help if needed.

as-com:
In case you are wondering what TypeScript could look like when applied to Snap<i>!</i>, see this commit, where I add types to some of the classes and functions: https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1863/commits/5bb7bc39a7949b4b0ea6c99cf59f983a075c02dc

brianharvey:
Okay, having read all that, _now_ can I run away screaming?  This (by which I mean TypeScript in general as well as this use in particular) strikes me as the very paradigm of a solution in search of a problem.  People who want to program in Java already have Java, and it even runs in browsers.  But no, they're not happy until every other language is turned into Java also.

Apart from the general distastefulness of B&D languages, a more specific problem with using one to implement Snap<i>!</i> would be making it harder for would-be users of the JSFunction block, who'd have to learn the extra incantations.

But of course it doesn't matter what I think; only one person gets a vote on this implementation idea.

as-com:
@brianharvey We probably shouldn't bundle the TypeScript compiler with Snap (it's huge). Since TypeScript compiles to JavaScript, people can continue writing the JSFunction block in plain JavaScript without any changes.

I wouldn't consider TypeScript to be a "B&D language," since the (optional) type annotations are mainly to prevent errors (such as misspelling a property in a class or passing a number to a function that only accepts a Point argument). TypeScript *is* JavaScript, so if you need to do something unsafe, it will let you do so.

brianharvey:
> mainly to prevent errors

Sure, that's what they all say!  :P

djsrv:
I have to agree with @as-com. Even though the build step is an extra annoyance, TypeScript and Webpack are incredibly helpful for writing clean, organized code. The (lack of) code organization is one of the worst parts of Snap! in my opinion, and I'd argue that the current style is much more off-putting to would-be tinkerers than the proposed solution.

I'm in favor of using type annotations (to prevent errors of course :P), but like @as-com said, they're optional, and TypeScript still allows you to make your code a heap of garbage if you really want to. It doesn't compromise flexibility.

brianharvey:
> The (lack of) code organization is one of the worst parts of Snap! in my opinion

Ooh, now you've done it!  :O

But, of course, being a mature adult, Jens won't yell at _you_; you're a kid.  He'll yell at me instead; I'm the designated whipping boy for all children in our gang.

JK.

jmoenig:
Well, now, I'm not yelling at anyone, but it's clear that I will never agree to use anything that resembles a "build" system. You know, the whole idea is about "live" programming!!

So, y'all like a zillion small files better than a few long ones? Fine, that's a matter of taste, but - honestly - *all* my text / code editors can handle all my source files no problem, so there's hardly a *technical* necessity to change anything.

*Efficiency* is a different matter, especially if it's about some internal optimization decisions within a particular VM. I usually try to stay clear of those, because, you know, the point of high level programming languages is to abstract those away...

Anyway, I hope you're having fun navigating the Snap source code system!

as-com:
Regarding the build system, Webpack has a "watch" mode, where the bundle is automatically recompiled when you save a file (which is much faster than you'd think). So effectively, it's still "live" programming, since by the time you've Alt-Tabbed to your browser and hit the refresh button, it's already done recompiling without you needing to do anything. I think there's even an option to configure Webpack to refresh your browser(s) for you!

In my experience with exploring Snap's code, WebStorm really doesn't like these giant files - it gets really slow when using functions like Find Usages or any of the refactoring options.

BookOwl:
@brianharvey, do any browsers still come with Java? I thought that Java applets were long dead.

as-com:
@BookOwl Java applets *are* dead. Most modern browsers like Chrome and Firefox with the notable exception of Safari (although it is still disabled by default) have already dropped support for Java applets.

brianharvey:
Interesting.  I still get update-Java popups from Oracle, maybe because Firefox 32 got rid of all my saved passwords, so I'm stuck at 31.  Now you're all going to yell at me...

as-com:
@brianharvey https://addons.mozilla.org/en-US/firefox/addon/password-exporter/

Now you can (and please do!) upgrade to the latest version, which is, by the way, version 55. :P

brianharvey:
Oh now you've done it.  I downloaded 55 and discovered that I don't even have to use that extension; somewhere between 39 and 55 (I got to 38, I'm pretty sure, before I gave up on trying new versions) the passwords came back.  So I'm back to security-rule-conforming, yay.  On the other hand, a dozen other extensions broke, so my look and feel is all wrong, and I'm going to have to spend the rest of the evening chasing down the issues and finding replacements.

But, back to the original question, I found this in Preferences>Applications:
firefox001
so it can't be entirely gone.

Jonathan50:
JNLPs are for starting Java programs on your computer from your browser, not for running Java applets in your browser.

cycomachead:
Sigh, there's a bunch of trade-offs here, but I don't think the first post represents them correctly.
I'm putting this out mostly as a information, and illustrating my pain points in developing Snap! stuff.  

First thing:
This only affects developing Snap<em>!</em>, not using it -- which (IMO) are two different meanings of the word 'live' -- though as-com is right, that webpack has modes that solve this problem. (In fact, there are modes which automatically reload the page when the source files change, but this wouldn't work for Snap<em>!</em>.)

Two be clear, this pull request does 2 main things, which are related by necessarily dependent:
* Sets up a build system
* Migrates to using typescript

Once setup, it's easy enough to use webpack in development with a server that auto-watches files. That said, the process is sometimes a pain to configure and it certainly makes the start up cost for new developers higher. (You must install node, in this case yarn too, and then bunches of dependencies.)

The small files are the one area that I will say make development _easier_. It's not about editors, but about knowing where to find things, and not thinking about whether something is in a particular file or not. 

As far as claims:

> Efficiency is a different matter, especially if it's about some internal optimization decisions within a particular VM. I usually try to stay clear of those, because, you know, the point of high level programming languages is to abstract those away...

This is a point in favor of a build system, I think. In this case the build system is also a Javascript-to-Javascript complier that makes some of those optimizations for you. 

As far as other claims:
* Loading performance isn't something this will really solve. We're downloading megabytes of JS that need to be parsed and executed. TS, or plain ES6 won't change that. In fact, if we had the capability of sing a server that supports HTTP/2, we might see better performance by using multiple files because you get much better responses across multiple requests. That said, file loading time is _not_ the big problem with Snap<em>!</em>. (Considering that browsers should be caching files, it should mostly be a 1 time problem anyway.)
    * A step of minification, removing comments, etc would also improve things, but again, it's mostly marginal.
   * In a world where we have a modern web server, multiple smaller files are also great, because browsers would only have to grab the pieces that changed. Currently, we just tell people to blow their caches away, which is dumb.

* In terms of performance from compliers and what-not, we can have much of this discussion by focusing on native ES6/ES201x versions of Javascript, without bringing TypeScript into this. A simple build system (for maximal compatibility) would get most of the benefits discussed without many of things people also don't like.
Modern browsers keep increasing support for new JS features, so it won't be long before we could write native ES2015 modules, and not use a build system in development. Safari supports them, and Chrome does in v61.

As far as contributions go:
Pretty much everyone I know learning JS is being exposed to the new syntax and paradigms. The  `class` keyword and structure is easier for people to grasp than prototypes -- though this is mostly syntactic sugar, but one that is more explicit and consistent across different JS projects. (i.e. Snap<em>!</em>'s use of `init()` and `uber`, compared to `constructor` and `super`)

brianharvey:
I think some of y'all need a refresher on the First Law of Engineering:

**If it's not broken, don't fix it.**

(Alternate wording: **If it works, leave it alone.**  Both commonly found in the wild.)

Exempted from this refresher are the parts of Michael's message having to do with needing to clear browser cache by hand for new releases.  That _is_ broken, and should be fixed.  But it should be fixed in the minimally disruptive way, imho, not by dragging in other questionably desirable things such as having millions of tiny source files.

[rant] Among the people who need a refresher on the First Law are the Mozilla Foundation, who apparently are planning to break all the Firefox extensions in the release after next, for no particular discernable reason other than "we know what you want better than you know what you want."  I already had to find the "ESR" version in order to be allowed to add the Page Colors button, and even then only by putzing around in about:config.  Luckily there's PaleMoon, which is apparently a Firefox offshoot by some guy who _doesn't_ think he knows what I want better than I know what I want.  Unluckily the Mac version is said to be flaky.  I think I'm just going to crawl back into my hole and leave the world to you whippersnappers. [/rant]

as-com:
I forgot to mention these in the main post...

Also note that another feature of Webpack is that it can automatically add hashes to its file outputs, so as long as you keep snap.html uncached (and the rest aggressively cached), users will never have problems with cache issues - the hashes will change and the browser will only download the updated parts.

Also, Webpack can automatically handle messy things like code splitting and loading assets on demand, so if there is a feature that requires a lot of code, you can have Webpack split it off into a separate file and load it when you need it. That's where the most load time performance improvements come from.

bromagosa:
Seriously, if we wanted to use a language that compiled into Javascript, it should be one that offers something extra in terms of live programming and powerful programming constructs, like [Clojure](https://github.com/clojure/clojurescript) or [Smalltalk](http://www.amber-lang.net/), but I don't think that's a very good idea either. Notice how these two implementations of programming languages are also written in JS. I think it makes all the sense in the world to use "the assembler of the web" to write programming languages.

As a side note, what kind of bloated IDEs are you using that can't handle relatively medium-sized files? My vim can open the whole Snap! codebase in less than 50ms without even blinking.

brianharvey:
> My vim can open the whole Snap! codebase in less than 50ms without even blinking.

What he said.  Except he misspelled "Emacs."

cycomachead:
> If it's not broken, don't fix it.

To be clear, I'm not necessarily advocating for a build system, but there are parts of the development process that I (and others) do feel are broken. There are many times where a set of smaller, more precisely named files and some folders would help find where things go.

as-com:
> Except he misspelled "Emacs."

Let's not have this argument here. Take it to `/dev/null` please.

> what kind of bloated IDEs are you using

I'm using WebStorm (from JetBrains). It's actually able to open and edit the files fine, but it gets sluggish when it tries to perform code analysis, refactoring, or symbol searching on a file like `morphic.js`.

brianharvey:
> There are many times where a set of smaller, more precisely named files and some folders would help find where things go.

For what it's worth, lots of small files would hurt my working style a little bit, but folders would hurt a lot, by making grepping harder, especially if I gave in to the temptation to cd into one of those folders.

BookOwl:
> For what it's worth, lots of small files would hurt my working style a little bit, but folders would hurt a lot, by making grepping harder, especially if I gave in to the temptation to cd into one of those folders.

That when you switch to an editor that has whole project grepping, or you could install [this emacs plugin](https://github.com/bbatsov/projectile) that lets you grep over a whole project even if you are in a subfolder.

BookOwl:
For what it's worth, I would find it much easier to navigate Snap's source code if it was split up into smaller files.

as-com:
Apparently a study shows that TypeScript (and similar) reduces bugs by 15%: http://ttendency.cs.ucl.ac.uk/projects/type_study/documents/type_study.pdf

I've noticed this effect of TypeScript first hand - it really helps the editor make accurate autocompletion suggestions and saves me time from tearing my hair out over a property name misspelling.

brianharvey:
Ah, okay, evidence is good.

If this were a group project like Firefox or something, I'd find the evidence compelling.  But in our case it suffers from the same flaw as all of education research:  Knowing what works best statistically for a large group of {teachers, programmers} tells me nothing about how {I, Jens} will work best.  So this interesting general discussion both won't and _shouldn't_ influence how Snap<i>!</i> is built.

cycomachead:
Also, linters would get us partially there, and I'm skeptical of studies about typescript that have MS employees as an author.

BookOwl:
If Snap! is going to switch to a different implementation language, it should obviously be a [language with monads](http://www.purescript.org/).

brianharvey:
> a language with monads.

😀 Great idea, you should rewrite it; I don't think any of the rest of us is competent to do it!

cycomachead:
In order to clean things up, I'm going to close this. 

As interested as I would personally be in using TS, there's no way it's happening in Snap<em>!</em> right now. :) But hey, you can comment on closed issues, so feel free to keep talking!

PR #1866, opened 2017-09-22
Webcam changes

bromagosa:
* Fixed for retina displays
* Fixed for any cam resolution and aspect ratio
* Cam support only disabled after failing to initialize video stream


jmoenig:
Yay, thanks, Bernat! This one I'll also pull on Monday.

PR #1867, opened 2017-09-22
Download script (as a hidden menu item)

bromagosa:


store.js
```diff
@@ -1116,7 +1116,7 @@ SnapSerializer.prototype.loadBlock = function (model, isReporter, object) {
     } else if (model.tag === 'custom-block') {
         isGlobal = model.attributes.scope ? false : true;
         receiver = isGlobal ? this.project.stage : object;
-        if (isGlobal) {
+        if (isGlobal && receiver) {
```
jmoenig: I never understood why we need to check for the receiver here. You sure it doesn't break anything else?

jmoenig: Actually, I'd like to know which cases this condition captures. Is this only about single scripts?

bromagosa: This was about importing single scripts, yep. I kind of forgot why I needed to add this guard, it was a long time ago (should've added a comment), but I know if I didn't there would be an error when importing a single script xml containing a custom block instance.

As far as I can tell, having had this feature in Beetle Blocks for a very long time, this doesn't break anything :)

jmoenig: Right, it doesn't seem to break anything. OTOH for the new inheritance mechanism I got rid of needing receivers mostly anyway, so in the new Snap! version this guard might not be needed anymore. Since it's hidden and experiment anyway I thought I'd just remove it, so we'll find out what happens ;-)

PR #1868, opened 2017-09-22
fixes #1860 (var refactor issue)

### LINKED ISSUE #1860: Refactor-renaming a parameter that has a type or default value not working
The rename input should hide the extra characters and only display the parameter _name_.

bromagosa:


blocks.js
```diff
@@ -3216,7 +3216,7 @@ BlockMorph.prototype.refactorThisVar = function (justTheTemplate) {
     // taking care of its lexical scope
 
     var receiver = this.scriptTarget(),
-        oldName = this.blockSpec,
+        oldName = this.instantiationSpec || this.blockSpec,
```
jmoenig: wait! that was the whole fix?! Yay, thanks!

jmoenig:
thanks, Bernat! I'll pull it on Monday.

PR #1871, opened 2017-09-26
Avoid flickering when resizing page

tjvr:
_@brianharvey This is the bug we were talking about._

Brian and I were experiencing some rather upsetting flicker when resizing the window in Safari. This seems like a useful thing to fix, particularly for people who have problems with flickering!

Here's how resizing works:

1. The resize event triggers `fillPage()`, which resizes the `worldCanvas`. 
2. Some time later, the scheduler will redraw the screen (I believe this is handled in `updateBroken()`?)

The issue is that setting the width/height of a `<canvas>` implicitly clears it. Since we don't redraw until the next frame, the screen remains clear until then: causing a flicker of white page.

### DIFF
--- a/morphic.js
+++ b/morphic.js
@@ -11663,6 +11663,10 @@ WorldMorph.prototype.initEventListeners = function () {
         function () {
             if (myself.useFillPage) {
                 myself.fillPage();
+
+                // fillPage will resize worldCanvas, which clears it.
+                // so we redraw immediately, to prevent flickering
+                myself.updateBroken();
             }
         },
         false


jmoenig:
Thanks, Tim!
Interesting how this is such an issue in Safari, but not in Chrome. I like your fix and will pull it later!

jmoenig:
Hmmm.... that doesn't seem to do anything for me. I'll revert....

tjvr:
Weird, it definitely helped for me.

@brianharvey How about you?

jmoenig:
are you using Safari 11? It's a nightmare...

tjvr:
> are you using Safari 11

Sorry, I forgot to clarify! I'm still on Sierra / Safari 10.

tjvr:
> Interesting how this is such an issue in Safari, but not in Chrome

Safari tends to be slightly more picky than Chrome about canvases; I suspect Safari's just sticking more closely to the spec. For example, trying to work with a (0,0)-sized canvas in Safari will cause errors :-)

brianharvey:
I'm on Safari 9!

PR #1887, opened 2017-10-14
Copy project URL to clipboard on share

namoscato:
This at least works in Chrome, but I was not able to test this end-to-end locally due to whitelisted server origins. I'm also up for suggestions on how to better target the `<input>` element.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -6295,11 +6295,19 @@ ProjectDialogMorph.prototype.shareProject = function () {
             'Share Project',
             function () {
                 myself.ide.showMessage('sharing\nproject...');
+
+                var hash = 'present:Username=' +
+                    encodeURIComponent(SnapCloud.username.toLowerCase()) +
+                    '&ProjectName=' +
+                    encodeURIComponent(proj.ProjectName);
+
                 SnapCloud.reconnect(
                     function () {
                         SnapCloud.callService(
                             'publishProject',
                             function () {
+                                var isCopied = false;
+
                                 SnapCloud.disconnect();
                                 proj.Public = 'true';
                                 myself.unshareButton.show();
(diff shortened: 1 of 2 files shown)


jmoenig:
very interesting, thank you! Let me have a closer look at this...

brianharvey:
If I'm understanding what this is trying to do, I'm against it; I don't like software to make decisions about my clipboard so I'm surprised when something doesn't work.  It's not very hard to click on the URL bar and then click copy.

namoscato:
@brianharvey, fair point.

I use Snap! to teach an Introduction to Computer Science course though, and surprisingly, it was difficult for some of the students to copy the URL from the browser's address bar. Project submission is dependent on sharing at the moment, and I was looking to optimize this flow. This seemed liked the quickest path, but do you have any thoughts on an alternative UX?

brianharvey:
Groan.  I suppose it could be a configurable option... if Jens will allow another one in the settings menu.

cycomachead:
We do have an alternate UX. Namely a #1084 -- though we could extend that to always show the URL dialog when you click "Share" even before you click "open".

jguille2:
Hi @cycomachead ,

I only point out that this PR is from 2015... and of course it would need different adaptations (especially the changes in the cloud functions).

Anyway, you know this feature was added in [Snap4Arduino](http://snap4arduino.rocks/run/#), so you can find [there](https://github.com/bromagosa/Snap4Arduino/blob/master/src/s4a/gui.js#L418-L460) an updated proposal.

You can test it in [Snap4Arduino online](http://snap4arduino.rocks/run/#).
[Note you also will find an "strange" _Open from URL..._ option... but required for our desktop version (without url bar))

Joan

PR #1891, opened 2017-10-20
Fix reportURL for URLs that have // in the middle

bromagosa:
Reported by @brianharvey.

When using proxies we often embed "https://" right in the middle of the URL. The reportURL block was mistakenly thinking that the protocol had already been added to the URL, thus trying to access a relative path.


jmoenig:
Looks legit, thanks, Brian and Bernat!

Issue #1890, opened 2017-10-19
SVG costumes don't export


maryfries:
I want to flip an SVG costume because most face right and I need one that faces left. I know we can't edit them in snap yet, but I can't export it and edit it either.

If I right-click one and choose export, a file downloads, but it won't open:

![image](https://user-images.githubusercontent.com/6549787/31785756-3924deb0-b4d4-11e7-9752-9dabc69905c8.png)
![image](https://user-images.githubusercontent.com/6549787/31785759-3b255d20-b4d4-11e7-881a-feee71f27e41.png)
![image](https://user-images.githubusercontent.com/6549787/31785769-432b3a30-b4d4-11e7-8e90-196c184d81ac.png)

brianharvey:
Confirmed with 4.0 and 4.1 in Chrome and Firefox with the "Abby" costumes from our costume collection.

jmoenig:
Is this *really* confirmed, Brian? does exporting "work" correctly in other browser? Did you try editing the original "Abby" costumes from the Media folder? I can import, export and re-import the "Abby" costumes fine in Chrome. Snap does *not* alter the SVG code, it just reads and dumps it, so it's rather unlikely that this is a Snap! issue.

jguille2:
I also can confirm this bug. You can import, export and re-import successfully. The problem is when you use directly an svg costume from Snap! costumes library (_Costumes..._ option menu).Then you can't export them with success.

So the problem is not into the import method. The bug is in the Snap! image coding.

If you import a svg, you can see Snap! stores " data:image/svg+xml;base64,PH...", but if you use a Snap! svg image, you'll see "data:image/svg+xml;utf8,%3Cs..."

I've been testing, and a solution could be change the [media import process into gui.js](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/blob/master/gui.js#L3350-3351):

Then
```javascript
img.src = 'data:image/svg+xml;utf8,' +
    encodeURIComponent(txt);
```
coult change to
```javascript
img.src = 'data:image/svg+xml;base64,' +
    window.btoa(txt);
```
I think that's all. I can send a PR to facilitate the change... but I wait for your opinions.

Joan

brianharvey:
Thanks, Joan, for coming to my rescue!  Here Jens was thinking I'm an idiot... :)

jmoenig:
[...]

PR #1893, opened 2017-10-20
Fixing #1890 - problems with svg costumes export

jguille2:
Changin loading process for svg media to allow its later export


jmoenig:
Thanks, Joan, for this excellent analysis and fix!

PR #1895, opened 2017-10-20
Libraries

brianharvey:
Didn't have time to write foreach in JS.  The other HOFs are candidates
for promotion to primitiveness.


jmoenig:
Say again, why are you rewriting stuff in JS that we can write in Snap itself? I love showing off MAP made in Snap *all the time* in just about *every talk* I give.

jmoenig:
I just don't like the current variadic implementation, so I always show the very simple implementation of MAP. And I use it on lots of data, lists of 100+k lengths

brianharvey:
Well, why did you write it in JS for your media project?  Because it's fast!  And that's equally important when you're trying to do big data.  That's the main answer; what follows is attempted answers to your followup questions.  :)

We don't look inside of MAP until late in the curriculum.  And then we don't _look_ inside it; we have kids write it.

Eventually I'll get around to inventing hybrid blocks.  Meanwhile I'm pushing in the direction of just including the hofs as primitives.

cycomachead:
Also, this is not an issue with this PR, but in general: Why do we need two copies of tools.xml? Why can the libraries directory just reference ../tools.xml?

Two copies of the same data should be avoided unless absolutely necessary.

jmoenig:
Thanks! Is this ready to be pulled now?

jmoenig:
Aw, come on...! Should I wait?

jmoenig:
Ready?

PR #1903, opened 2017-10-26
Update LICENSE file so it gets properly displayed

damiencaselli:
GitHub treats LICENSE files in a particular fashion. Whenever this file
is present in a repository, an icon is displayed at the top of the
repository page. This commit enables this behavior.

LICENSE text itself was updated to mirror the one from gnu.org website,
which had some differences such as https everywhere.


jmoenig:
Thanks, Damien! I'll pull this probably next week (after the dust of our big release has settled), and put it online.

PR #1907, opened 2017-10-29
Update Portuguese translation

MMSequeira:
Several new translations, a few corrections, some translations of non-official strings made official.


jmoenig:
Thanks, Manuel! I'll pull this sometime next week and integrate it.

PR #1914, opened 2017-11-02
Improve Unicode Blocks for Some Emoji

cycomachead:
This makes the 🎃 work properly in the `unicode of` block.
Snap<em>!</em> now correctly reports '127875'.

~~However, `unicode (127875) as letter` still doesn't work quite right.~~
Fixed..was the missing return. D'oh!

Also I noticed edge cases in the `unicode of` block. Mostly around the values `false` and `0`. Minor, but I fixed that up.

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -2688,7 +2688,7 @@ Process.prototype.reportStringSize = function (data) {
 };
 
 Process.prototype.reportUnicode = function (string) {
-    var str = (string || '\u0000').toString()[0];
+    var str = isNil(string) ? '\u0000' : string.toString();
 
     if (str.codePointAt) { // support for Unicode in newer browsers.
         return str.codePointAt(0);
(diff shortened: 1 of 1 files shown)


threads.js
```diff
@@ -2688,12 +2688,20 @@ Process.prototype.reportStringSize = function (data) {
...
+        return str.codePointAt(0);
+    }
+    return str.charCodeAt(0);
 };
 
 Process.prototype.reportUnicodeAsLetter = function (num) {
     var code = +(num || 0);
+
+    if (String.fromCodePoint) { // support for Unicode in newer browsers.
+        String.fromCodePoint(code);
```
dekrain: Maybe `return String.fromCodePoint(code);` ?

cycomachead: D'oh, thanks!

jmoenig:
Thanks, Michael! I'll pull this next

PR #1950, opened 2017-11-30
Fixing #1932, csv spliter, according to RFC 4180 spec

### LINKED ISSUE #1932: CSV Format Weirdness?
Using "split [ ] by [CSV]" seems to give results that don't follow CSV specification. For example, single-quotes work the same as double-quotes when in fact they shouldn't do anything.
Are you guys using a different spec from _RFC 4180_?

jguille2:
This fix #1932, bringing CSV spliting to [RFC 4180](https://tools.ietf.org/html/rfc4180) specifications and therefore, getting a full compatibility with spreadsheets exports.

Theses specifications are:
  - We have only one record, with fields separated by commas.
  - The last field in the record must not be followed by a comma. An end _comma_ means an extra empty field.
  - Spaces are considered part of a field and should not be ignored (they don't need text delimiters).
  - The only text delimiters are _double-quotes_. Single quote is plain text.
  - Fields with _double-quotes_ or _commas_ (as content) must be enclosed in double-quotes (as delimiters).
  - _Double-quotes_ content must be escaped by preceding it with another _double-quote_ [""].  There are not other escapers (nor [/"]).

It seems to be coherent with spreadsheet export operation (full tested with LibreOffice and a single test with Excel).

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -2745,15 +2745,10 @@ Process.prototype.parseCSV = function (string) {
     // by lines.
     // Following RFC 4180 specifications
 
-    var re_valid = /^(?:"(?:(?:"{2})*[^"]*)*"|([^",]*))(?:,(?:"(?:(?:"{2})*[^"]*)*"|([^",]*)))*$/;
-    if (!re_valid.test(string)) {
-        return new List();
-    }
-
     var re_value = /(?:^|,)(?:"((?:(?:"{2})*[^"]*)*)"|([^",]*))(?=(?:,|$))/g,
         a = [];
 
-    string.replace(
+    var remaining = string.replace(
         re_value,
         function(m0, m1, m2) {
             if (m1 !== undefined) {
@@ -2766,6 +2761,10 @@ Process.prototype.parseCSV = function (string) {
             return '';
         }
     );
+    if (remaining !== '') {
+        // if remaining contains something, string has a non valid format
+        return new List();
+    }
     return new List(a);
 };
 
--- a/threads.js
+++ b/threads.js
@@ -2743,35 +2743,29 @@ Process.prototype.parseCSV = function (string) {
     // parse a single row of CSV data into a one-dimensional list
     // this assumes that the whole csv data has already been split
     // by lines.
-    // taken from:
-    // https://stackoverflow.com/questions/8493195/how-can-i-parse-a-csv-string-with-javascript-which-contains-comma-in-data
+    // Following RFC 4180 specifications
 
-    var re_valid = /^\s*(?:'[^'\\]*(?:\\[\S\s][^'\\]*)*'|"[^"\\]*(?:\\[\S\s][^"\\]*)*"|[^,'"\s\\]*(?:\s+[^,'"\s\\]+)*)\s*(?:,\s*(?:'[^'\\]*(?:\\[\S\s][^'\\]*)*'|"[^"\\]*(?:\\[\S\s][^"\\]*)*"|[^,'"\s\\]*(?:\s+[^,'"\s\\]+)*)\s*)*$/,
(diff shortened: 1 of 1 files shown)


brianharvey:
Thanks!  So, you think we can/should ignore the issue of newlines in quoted fields?

jguille2:
csvsplit

brianharvey:
Yes, but since split-by-csv handles a single record/line, the normal style of use is 
csv-split
so by the time your code sees it, it's no longer possible to combine the two lines.  This is why I think we have to change the definition of split-by-csv so that it splits the entire file into a list of lists.

jguille2:
Yes, it is clear... and it was the discussion between a general implementation or the fix for current spliter.

Current [split by csv] is already running with _newlines_ ... but the problem is you break before the csv _spliting by line_ (of course, only if it have _newlines_ into fields).

I'll send another code (maybe into a custom block... for a lib) to do the full csv import (we can do a little more... exporting...). But I think we can fix current behavior and go on...

jmoenig:
Thanks, Joan! I'm not sure, should I pull this or does this need more discussion?

brianharvey:
This is one of those the-good-is-the-enemy-of-the-best situations.  Joan's solution is way better than what we have now, and will probably handle 95% of the cases, so we'll never get around to the other 5%.  I guess it's fine to pull, as long as we get that library one done also.

jmoenig:
Thanks, Brian! I'm marking this as "blessed" then, and will pull it in the next pulling-party. I agree about the library, let's (get someone to) work on that :-)

jguille2:
Hi @jmoenig, @brianharvey 

Yes, I think we can close this issue (about the _csv spliter_ and its spreadsheet compatibility).

I will return to work to get a _csv library_ but, if you have a little time you could look at [this Snap4Arduino comment](https://github.com/bromagosa/Snap4Arduino/pull/215#issuecomment-354107948) with the idea of implementing a full feature of variable import/export.

The main idea is use the XML format for a full import/export implementation and add a "plain text"  option (_.txt_, _.lst_ or _.csv_ depending on the variable content) to get a partial import/export feature. This last "partial plain text" method is not doing a complete store of data (complex lists of lists, other first class data...) but it is very useful to be able to work with the data (for example with a spreadsheet) and also presents a minimum coherence for an import/export feature.

If you think it's interesting for Snap!... Anyway, any comment will be welcome.

Joan

jmoenig:
I had to revert this, because it didn't work, caused an infinite loop and Snap! to crash when I tried to parse an actual CSV.

To paraphrase Knuth: Please don't just prove that your code is correct, but make sure it actually works :-)

jmoenig:
@jguille2 can we somehow reopen this PR, or could you perhaps issue another one with this very code, so we can debug it together? thanks!

jguille2:
OOps!!

I'm sorry! I try to check my proposasl before a PR... but here I'm not sure what's happening... Its [an issue](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/issues/1932) from November and a lot of discussion and changes in the meantime.

Anyway, I'm sorry... and I'll check this problem.

jguille2:
Excuse me, @jmoenig 

Maybe I'm missing something... because I've re-checked the code and I've seen it running as I expected.
csvslitted

Could you point me that example that crashes Snap?

jmoenig:
Ah, maybe *I* made a mistake, hmm... I swear it didn't work when I tried it with a big CSV. The whole page started to become unresponsive at I had to force-kill Chrome after 30 mins. But you're right, it could be something else that's happened since November...

jguille2:
Hi @jmoenig 

I can not reproduce the problem. I've tested again with Firefox and Chromium (not the last versions), with CSVs with different strings (double quotes, simple quotes, commas...) and I haven't had any problem.

Also, I've tested with large files... and the result has been fine... although this has no relation with "csv plitter"... because [split by csv] is only a "csv line splitter" and if we have tons of records, [split by line] is doing the job. (I don't want to restart here the discussion about csv import feature... only fixing the issue reported about single quotes and the standard protocol and spreadsheet operation).

Continue...

PR #1955, opened 2017-12-06
Update lang-ru

temaps:


jmoenig:
Ah, interesting! Your PR would allow help files to be translated, also, right? By pulling this we could also let everybody else translate help screens to their language, because they would each be in separate directories. Let me look at this more closely...

temaps:
Yes! Just put the pictures transfer to the folder with the language ID. For the Russian folder "ru" for German, the folder "de". If the picture is not translated, it will automatically load the English picture of a folder "help"
`tmpimg.onload = function(){pic.src = ide.resourceURL('help'+'/'+SnapTranslator.language, spec + '.png');};
+        tmpimg.onerror = function(){pic.src = ide.resourceURL('help', spec + '.png');};`

jmoenig:
I've tried to integrate your translation additions in 13d4ac87fd1575266b2530dcc16685c2534955fe, keeping your proposed little changes to the translation system open for the moment...

temaps:
Thank You! :-)

PR #1956, opened 2017-12-08
fixes #1913

### LINKED ISSUE #1913: Cursor Type Beam Stuck Glitch
The flashing text indicator does not disappear when a click inside a type area and drag from the template is done.
image
Only occurs when click is made from inside the template window.
image image

bromagosa:
I just added a call to `this.world.stopEditing();` before copying a template morph when grabbing it.


jmoenig:
Yay, thanks, Bernat! I'll pull this shortly...

PR #1966, opened 2017-12-18
Undo and collaborative editing support

brollb:
Collaborative editing options will only show if the server supports it (ie, has a websocket connection that can be used for the collaborative editing).

For an example of a server that can be used for the collaborative editing features, check out https://github.com/NetsBlox/collaboration


brollb:
Just for the record, this has the same content as #1812 and was only re-created to allow me to make the PR from a feature branch rather than the master branch in the fork of Snap!

PR #1968, opened 2017-12-20
Suggestion to allow Snap menus customization for Snap! extensions

jguille2:
Snap menu functions are building and painting their menus at the same time.

This PR is only spliting building and painting (adding a _return_ of the menu before painting) to allow customization of these menus into Snap! extensions.

For example, Snap4Arduino has extra-options. This PR will allow to remove an ugly hack.

Thanks,
Joan

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -562,7 +562,8 @@ IDE_Morph.prototype.createLogo = function () {
     };
 
     this.logo.mouseClickLeft = function () {
-        myself.snapMenu();
+        var menu = myself.snapMenu();
+        menu.popup(myself.world(), myself.logo.bottomLeft());
     };
 
     this.logo.color = new Color();
@@ -813,7 +814,10 @@ IDE_Morph.prototype.createControlBar = function () {
     // projectButton
     button = new PushButtonMorph(
         this,
-        'projectMenu',
+        function () {
+            var menu = myself.projectMenu()
+            menu.popup(myself.world(), myself.controlBar.projectButton.bottomLeft());
+        },
         new SymbolMorph('file', 14)
         //'\u270E'
     );
@@ -837,7 +841,10 @@ IDE_Morph.prototype.createControlBar = function () {
     // settingsButton
     button = new PushButtonMorph(
         this,
-        'settingsMenu',
+        function () {
+            var menu = myself.settingsMenu()
+            menu.popup(myself.world(), myself.controlBar.settingsButton.bottomLeft());
+        },
         new SymbolMorph('gears', 14)
         //'\u2699'
     );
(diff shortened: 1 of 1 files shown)


bromagosa:
This is in line with what @cycomachead, @brollb and I agreed would be easy, harmless changes that would make it easier for extension devs to modify menus.

A good convention would be that functions creating menus should always return those menus, and callers are the ones who should pop them up.

From my side, this PR looks good and would help us get rid of some spaghetti in -at least- Snap4Arduino and Beetle Blocks :)

jmoenig:
I'd like to understand more about the spaghetti code that this PR addresses. If an extension wants to add items to a menu, wouldn't you just override the function that build the menu? Tnx!

bromagosa:
You would, but then when Snap! gets updated and you've overridden 10 menus, you have to scoop through each item in each menu to see what changed and apply those changes to your overridden functions as well.

If menus return themselves, you can just decorate them, add items to them, and return them again. This way, even if Snap! is updated and menus change, your decorators will always work.

cycomachead:
Is there interest in pulling this? It's a small tweak and still good?

(There's a couple formatting nitpicks, but otherwise I think this is fine.)

bromagosa:
Since this was never pulled we found another strategy that kinda works. It's a kludge, but it lets you modify menus without having to override functions.

The idea is:

```
originalMenuFunction = aMenuTriggeringFunction;
aMenuTriggeringFunction = function () {
  originalMenuFunction();
  var menu = world.activeMenu;
  // add or remove items or modify the menu in any other way
  menu.popup(world, menu.position());
}
```

Is it ugly? You bet. Does it work? Totally. Is it better than having to go through every single menu function every time there's a new release? Most definitely.

jmoenig:
methinks this isn't ugly at all, and it works perfectly. Yay!!

bromagosa:
Well, it's ugly from the point of view that the menu has already popped up when you modify it. 

I'd be cleaner to modify it before it's had the chance to show up at all, but this is okay, I guess.

Modifying it is obviously so fast that no one will ever know, but every time I bring up a menu in Snap4Arduino *I know* the menu has popped up and re-popped up in there! ;)

jmoenig:
yeah, but that's even before the display gets a chance to update itself, so nobody will ever actually see the change. If it's faster than the speed of light you're fine :-)

brianharvey:
I want users to be able to edit the menu bar (adding a new menu) or a menu (adding or removing an item).  Just sayin'.

cycomachead:
It really depends on the situation — it’s definitely uglier than proper support and on slower devices you’re more likely to notice something. 

There’s always competing interests but this is not as easy for library / extension authors. (Not terrible mind you, but not as clear.) 

-- 
Michael Ball
From my iPhone
michaelball.co

PR #1969, opened 2017-12-24
#1944 Always set fragment identifier on setURL()

### LINKED ISSUE #1944: 'Save as' to a saved project should update the URL
If I'm in shared project-1 and save it to project-2 then the name of the project changes and subsequent saves update project-2. But if I refresh the tab I'm back to project-1.

namoscato:
Resolves #1944.

Ensures the fragment identifier is cleared out when performing "Save as" on a shared project and "Project URLs" is disabled (the default).


jmoenig:
This makes sense to me. Thanks!

PR #1970, opened 2017-12-24
Add build process for JavaScript uglification and cache busting

namoscato:
I'm not sure what the build / release process looks like for this project at the moment, but I figured I would take a pass at this regardless.

This introduces [gulp](https://gulpjs.com/) to address two items:

1. **JavaScript uglification** via concatenation and [UglifyJS](http://lisperator.net/uglifyjs/) brings 16 requests / 1.8 MB down to 1 request / 606 KB
2. **Cache busting** via a "version" query string mitigates stale browser code after a release

To build, you will need [Node.js](https://nodejs.org). Once that is installed:

```
$ npm i -g gulp-cli
$ npm i
$ gulp all
```


cycomachead:
Right now, there's no interest in maintaining a build system for Snap<em>!</em>. (I personally support the idea of having one, but Jens and others don't want the complexity.) So, I'm going to close this.

Also, since technology has adapted a bit, the idea of concatenation of assets isn't as useful anymore. Concatenation helps by reusing an active server connection. However, in Snap<em>!</em>'s case, since the total number of dependencies from the server is relatively low, and we now are seeing greater user of HTTP/2 servers, there's more benefits to serving (a minified) the bundle separately.

PR #1977, opened 2018-01-03
Use raw image src if costume image not loaded

brollb:
This is a pretty minor async bug. If a costume is serialized before the image data is loaded, it will basically clear itself (because the current image data is nothing).

Not a very pressing issue but I thought I would make a PR in case you were interested!

### DIFF
--- a/store.js
+++ b/store.js
@@ -1342,6 +1342,7 @@ SnapSerializer.prototype.loadValue = function (model) {
                 v = new SVG_Costume(null, name, center);
                 image.onload = function () {
                     v.contents = image;
+                    v.imageSrc = null;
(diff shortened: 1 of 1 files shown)


jmoenig:
Oooh, interesting, Brian! Does this address asynchronous loading of costume ressources?

brollb:
Thanks! 

This should address it when the costume is loaded from xml (as the image data is loaded async). It looks like that is the only spot that costumes are loaded async (that I know of anyway) - are there any others that I am missing?

In other words, it covers the only async loading of costume image/svg data that I know of - if there are any other spots that I missed, I would be happy to address them as well!

PR #1983, opened 2018-01-16
crayons library

brianharvey:
SET PEN TO CRAYON and CHANGE CRAYON BY blocks, with helpers


jguille2:
Uf! Colors again 😄

I would have to remember many things... but, at the moment, only a comment...

- A lot of good work with the colors selection and the mapping to search the "nearest crayon".

- The only problem I see is [change crayon by] is too slow. If these crayons blocks went to the core, the crayon number could be stored as a sprite property (yes if the pen color is changed with other blocks, crayon number must be reset). Then, I could paint changing the crayon quickly.

brianharvey:
Right, it's too slow because it basically has to recreate the color chart every time.  On the list of things to do is to invent a way for libraries to create global variables (preferably in a separate namespace to be sure not to conflict with user-defined ones).  Then we can remember which crayon we're using.

I think we decided _not_ to put crayons in Snap<i>!</i> itself, partly so people can create their own (CHANGE CRAYON (n) TO (rgb)) or otherwise mess around with the library.  (And I'm already having second thoughts about a few of the colors. :-/ )

jguille2:
Hi @brianharvey ,

I wanted to see speed changes and comment something... but, first, two important things before merging:

- Last commit has erased custom block definitions (from other libraries) that are used inside. Then this lib is broken! Plenty of _obsolete_ blocks when it is called alone.

- The lib is not added to LIBRARIES catalog.

And an extra comment to think (maybe later) about "secondary" libraries (libraries using other basic libraries). We'll get the same block definition in different libraries (the same name... and the same definition?). This can be a problem with their maintenance. Of course, this is not a problem with our personal custom blocks (a flexible way to create our own libraries) but maybe "official" libraries can have another robust behavior.
A draft idea: To build a single repo of block-definitions and the libraries catalog only lists of these blocks to pick them in groups.
Excuse me. I don't want to complicate this PR... Just a thought to share with you.

Joan

brianharvey:
Oh god thank you!  No wonder Jens doesn't trust me...  
@jmoenig Please pretty please pull Ilay's transitive-closure change to Export Blocks.

jmoenig:
@brianharvey please do *not* introduce branches to the Snap repo, but instead *clone* the repo to issue pull requests. Tnx

jmoenig:
I just wanted to integrate this, but there are errors popping up about block definitions missing. In particular the ```(...) empty?``` predicate seems to be missing as it is used by the version of ```map```  you're using here.

Tip: When you create a library, make a little project with all the demo/test cases in the scripting area, and also import all the other libraries you need. Once your test cases all pass unload all unused blocks and then export *all* blocks as the library. That way you always get *exactly* the right blocks.

brianharvey:
Aargh, okay.  I still want Ilay's transitive-closure feature!

jmoenig:
with a transitive hull we'll end up getting getting whole sprites, costumes, sounds, even entire projects etc. in libraries. Just fix your stuff, it's great, I want to pull it already :-)

brianharvey:
On 2/1/18 1:27 PM, Jens Mönig wrote:
> with a transitive hull we'll end up getting getting whole sprites, costumes, sounds, even entire projects etc. in libraries. Just fix your stuff, it's great, I want to pull it already :-)
>

No, silly, just transitive with respect to procedures. :-)

I'm busy this afternoon but will fix the library this evening.

jguille2:
One suggestion,

Brian, could you **rebase** _crayons_ branch before merging this PR?

If we don't rebase PR with multiple commits (especially if they have _merges from master_), git history becomes a mess...

Thanks

jmoenig:
Ugh, yeah, that would be nice. But hey, I'm just gonna merge this in now because I want to work on getting that release out. Thanks, everybody, for your support!

PR #1984, opened 2018-01-16
updating Turkish Language

yildizhuseyin:
we would like to add Turkish language and add missing words.


jmoenig:
Wow, this really brings the Turkish translation up to speed, which is exciting. Thank you! I'll pull this for the next release.

PR #1986, opened 2018-01-16
Minor enhancement for keeping the appMode while opening a project

hamidzr:
`openProject` call doesn't take into account whether currently the UI is in appMode (fullscreen stage) or not and thus it displays a few elements despite being in appMode.
ss

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -1041,6 +1041,7 @@ IDE_Morph.prototype.createPalette = function (forSearching) {
 
     this.palette.setWidth(this.logo.width());
     this.add(this.palette);
+    if (this.isAppMode) this.palette.hide();
     return this.palette;
 };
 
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks, Hamid! Can you explain how this issue happens? I could swear that originally Snap! was supposed to switch to edit mode whenever a new project is opened. A quick check seems to indicate that this still is the case, but - of course - when you've written something yourself it's often hard to find an error but you somehow subconscientiously avoid it ;-)

hamidzr:
My bad - I didn't do the best job of explaining why I was looking into this. You are correct that opening a project from app mode will exit app mode automatically and will not have this issue. 

However, we have been recently looking into running Snap! on small mobile devices (primarily cell phones) in a 'run-only' sort of mode (since the screen sizes can make any real editing virtually impossible). Therefore we are considering locking the user into the appMode, but we still need to be able to open projects without leaving this mode. In order to achieve this, Snap! will be driven by some sort of wrapper code which is designed specifically for this family of devices.

We are submitting this PR because we thought that other Snap derivatives might also be interested in a similar approach so we figured we would share it just in case. So basically, we are presenting it as a new (minor) suggested feature rather than a bug fix

PR #1991, opened 2018-01-19
Fixing lang-zh_CN issue reported in #1980

jguille2:
Thanks @hljhttp,
I push your reported fix to unbreak  简体中文 translation. And then, we can close https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/pull/1980 with that merged code.


jmoenig:
Ah, I see. Thanks, @jguille2 !

PR #1881, opened 2017-10-04
New update of the Spanish language file

cienciascontic:
I've added new translations to the lang-es.js. See attached here the file in zip format. Sorry I'm not so experienced in GitHub!!
[lang-es.js.zip](https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/files/1356041/lang-es.js.zip)

PR #1992, opened 2018-01-19
Spanish translation update

jguille2:
Add Cristián Rizzi Spanish translation update (thanks @cienciascontic !)
- Adding new strings and fixes (and date and translation credits)
- Fixing format, identation and deleteing TAB controls chars
- Deleting string translations not used in Snap!

And we can close #1881


jmoenig:
Thank you so much!!

PR #1995, opened 2018-01-19
Convert unsupported categories to 'custom'. Fixes #1994

### LINKED ISSUE #1994: Gracefully handle unsupported blocks (when importing from Snap extension)
Currently, Snap does a pretty good job of doing a "best effort" import when importing xml generated from an extension (which has blocks that are not a part of vanilla Snap). However, it still sometimes throws an error when importing projects using blocks that are not part of vanilla Snap.

I believe this happens when the importing xml has blocks belonging to an unsupported category which are connected to one another (resulting in a failure when trying to make them "zebra colored").

brollb:
I added checking if the block category is supported in the current environment; unsupported categories will be converted to "custom" on import.

### DIFF
--- a/store.js
+++ b/store.js
@@ -805,7 +805,8 @@ SnapSerializer.prototype.loadCustomBlocks = function (
             child.attributes.s || '',
             object
         );
-        definition.category = child.attributes.category || 'other';
+        definition.category = SpriteMorph.prototype.categories.indexOf(child.attributes.category) !== -1 ?
+            child.attributes.category : 'other';
         definition.type = child.attributes.type || 'command';
         definition.isGlobal = (isGlobal === true);
         if (definition.isGlobal) {


DavidGasku:
Seems that my problem could be addressed by this fix:
Sometimes I create custom blocks in a custom category, and a JS block adds that category. 
In its actual state, the custom blocks can't be accessed. They should be in custom category by default.
What is worse is that a search freezes Snap for some seconds if the search starts with the same letter of the custom block.

Check this project:
https://snap.berkeley.edu/snapsource/snap.html#present:Username=davidgasku&ProjectName=customCategory

There is a **customBlock** in a **customCat**. Before using the run block, you can't access the block, and if you try to find "customBlock" it will freeze. once you add the customCat, everything is fine.

Anyway, I think the best solution would be to create a nice way to create custom categories, and store them in the project. It would be great to setup lessons.

Issue #1988, opened 2018-01-17
Mouse wheel events


DavidGasku:
Would there be the possibility to add mouse wheel events (up, down & click).

PR #1997, opened 2018-01-22
Add scroll events to receiveInteraction hat morph

bromagosa:
Addresses #1988.

### DIFF
--- a/objects.js
+++ b/objects.js
@@ -4827,6 +4827,10 @@ SpriteMorph.prototype.mouseDownLeft = function () {
     return this.receiveUserInteraction('pressed');
 };
 
+SpriteMorph.prototype.mouseScroll = function (y) {
+    return this.receiveUserInteraction('scrolled-' + (y > 0 ? 'up' : 'down'));
+};
+
 SpriteMorph.prototype.receiveUserInteraction = function (interaction) {
     var stage = this.parentThatIsA(StageMorph),
         procs = [],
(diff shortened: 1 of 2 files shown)


bromagosa:
It was shorter to implement than to ask if there was interest for it :stuck_out_tongue_winking_eye:

objects.js
```diff
@@ -4827,6 +4827,10 @@ SpriteMorph.prototype.mouseDownLeft = function () {
     return this.receiveUserInteraction('pressed');
 };
 
+SpriteMorph.prototype.mouseScroll = function (y) {
+    return this.receiveUserInteraction('scrolled-' + (y > 0 ? 'up' : 'down'));
```
jmoenig: Hmm, this feels weird on a Mac nowadays, where they've mirrored the direction. But it's nice to play with this!

jmoenig: ... and - of course - we're missing the value (y) that's been scrolled. But it's actually rather useful even without that. Thanks!

jmoenig: Hmmm... "up" and "down" is more of a problem than I expected it to be, especially since you can change it in your settings. This is going to be hard to explain to users. I wish we could get the actual data of the scroll wheel or touch pad...

brianharvey: Call it "topward" and "bottomward."  :-)

Apple is just wrong about this, and we should take a stand for truth, justice, and the German way.

bromagosa: What about something like scroll and reverse scroll?

PR #2013, opened 2018-02-09
fixed dangling for "penjant"

joamuran:
An small translation bug...


jmoenig:
Ah, this crossed my own integration efforts, so I've already added this in 20c3fee33a16f8eeccafafbcd779ba59ba2cdae7
Thanks!

jmoenig:
Can you please check whether the result now makes sense to you? Thanks!

PR #2015, opened 2018-02-12
New user account verification

bromagosa:
This PR addresses user accounts verification. When users sign up, they are flagged as unverified and an email is sent to them with a link to verify their account.

Unverified accounts will still be able to log in for three days, but after that period their account will be suspended and they will need to request a new verification code.

When a user logs into an unverified account, a message pops up showing the number of days left before their account is suspended.


jmoenig:
Yay!! Will pull this one in a bit. Thank you much, Bernat!

PR #2018, opened 2018-02-13
unescaping all parameters

bromagosa:
I'm astonished none of the tests we ran failed. I was not escaping any parameters!


jmoenig:
hmmm.... maybe nobody used any funny characters in either their usernames or project titles. Or, nobody really tested it :-/

bromagosa:
True. I just found out because of a project that ended in `?`

jmoenig:
(online now)

PR #2019, opened 2018-02-14
Another slip

bromagosa:
I'm sorry about this, I seem to break something every time I fix something.

`encodeURIComponent(null)` returns `"null"`, so the previous `username || '%username'` idiom wasn't working now, as `encodeURIComponent(username) || '%username'` would result in `"null" || '%username'`, breaking almost all queries that require credential checking :(

This fixes it and hopefully doesn't introduce new bugs again.


jmoenig:
happens to me everytime ;-)
online now.

PR #2020, opened 2018-02-14
Fix the MOTD feature.

cycomachead:
We should backport this, but the changes are pretty small.
When no file is there, the check is ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 *.swp
 .tern-port
+motd.txt
--- a/gui.js
+++ b/gui.js
@@ -310,21 +310,6 @@ IDE_Morph.prototype.openIn = function (world) {
 
     this.reactToWorldResize(world.bounds);
 
-    function getURL(url) {
-        try {
-            var request = new XMLHttpRequest();
-            request.open('GET', url, false);
-            request.send();
-            if (request.status === 200) {
-                return request.responseText;
-            }
-            throw new Error('unable to retrieve ' + url);
-        } catch (err) {
-            myself.showMessage('unable to retrieve project');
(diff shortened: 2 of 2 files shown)


cycomachead:
@jmoenig @brianharvey @bromagosa 

IMO, it'd be good if we can start using this with a custom MOTD on each of the Snap! instances.

jmoenig:
Thanks, I'll pull this later. It's good the the mechanism gets updated. However, I've personally grown very skeptical about this feature and in general don't think it's a good idea. So, let's have a discussion about this before we start showing such messages.

brianharvey:
On 2/14/18 4:36 AM, Jens Mönig wrote:
> Thanks, I'll pull this later. It's good the the mechanism gets updated. However, I've personally grown very skeptical about this feature and in general don't think it's a good idea. So, let's have a discussion about this before we start showing such messages.
> 

No, you're missing the point, which is that we need it /right now/!

jmoenig:
We can't have it *right now*, because we're not deploying to the current installation, only to dev.

cycomachead:
Always shown messages are annoying. However, they serve a very specific purpose and are necessary at times. 

Cutting access off to a service is more annoying for users. 

And there’s literally nothing stopping us from updating the existing version! It’s a copy and paste or we could make a minor new release off 4.1.1. 

We need to do this. 

-- 
Michael Ball
From my iPhone
michaelball.co

brianharvey:
On 2/14/18 9:43 AM, Jens Mönig wrote:
> We can't have it *right now*, because we're not deploying to the current installation, only to dev.
> 

We could, just this change, if you didn't have this strange aversion to the 
whole idea.

PR #2023, opened 2018-02-15
Revert "Fix the MOTD feature."

jmoenig:
Reverts jmoenig/Snap--Build-Your-Own-Blocks#2020
doesn't work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .DS_Store
 *.swp
 .tern-port
-motd.txt
--- a/gui.js
+++ b/gui.js
@@ -310,6 +310,21 @@ IDE_Morph.prototype.openIn = function (world) {
 
     this.reactToWorldResize(world.bounds);
 
+    function getURL(url) {
+        try {
+            var request = new XMLHttpRequest();
+            request.open('GET', url, false);
+            request.send();
+            if (request.status === 200) {
+                return request.responseText;
+            }
+            throw new Error('unable to retrieve ' + url);
+        } catch (err) {
+            myself.showMessage('unable to retrieve project');
+            return '';
+        }
+    }
+
     function applyFlags(dict) {
         if (dict.embedMode) {
             myself.setEmbedMode();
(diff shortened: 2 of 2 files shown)


cycomachead:
What doesn't work?

jmoenig:
Ah, sorry for not responding. I was pulling this shortly before the release, and right when I tried it I got all kinds of errors in my JS console. In retrospect it occurs to me these errors might have been caused by the files it was looking for not being present in the "dev" distro or on my local server. So, yeah, this has probably nothing to do with your code, and all with the mechanism.

You know, I do regret this feature. I think just popping up an ```alert``` as we ended up doing last weekend might actually have been better, because it sure got everybody's attention. Maybe it's as easy as just that. The ```alert``` had the additional benefit of being right inside the ```snap.html``` file, so we didn't have to exchange more files in the mirrors.

One problem with the MOTD feature was aggressive browser caching. Since we're now addressing this by manually changing the version date in the ```snap.html``` file anyway, we might as well *only* touch that file when we need to alert users about impending doom.

Yes, I'm fine with taking the MOTD feature out altogether.

brianharvey:
In that case, can we please leave the message code in snap.html, commented out, so we can revive it easily in a crisis?

jmoenig:
inserting an ```alert('panic!')``` statement is easy enough, we don't need a template for that.

brianharvey:
Easy for _you_.  Having a comment in the right place would make it easy for me also.

cycomachead:
Thanks for mentioning that. It’s confusing when something doesn’t work even though I definitely tested that spect. 

Literally anywhere in JS that’s evaluates will suffice. 

But while alert is nice and effective, we should be careful. It’s one of those features that browser vendors are slowly restricting the use. In most browsers there’s now a checkbox that allows users to ignore alerts from a domain. I don’t know if it’s per-session or per-domain, but we need to be careful in the future. 

-- 
Michael Ball
From my iPhone
michaelball.co

cycomachead:
> One problem with the MOTD feature was aggressive browser caching. Since we're now addressing this by manually changing the version date in the snap.html file anyway, we might as well only touch that file when we need to alert users about impending doom.

Oh yeah, I missed this earlier. In the future if er need such a thing we can do `getURL('motd.txt?' + (new Date()).toString())` and that would fix any caching issues.

PR #2028, opened 2018-02-20
Checking if user is verified on session init

bromagosa:
This makes sure we're checking if a user is verified when the session starts (so, when you open Snap!).

We're also adding some info on how to get your verification code.

Please let me know if the wording is okay.

### DIFF
--- a/cloud.js
+++ b/cloud.js
@@ -208,6 +208,7 @@ Cloud.prototype.checkCredentials = function (onSuccess, onError, response) {
         function (user) {
             if (user.username) {
                 myself.username = user.username;
+                myself.verified = user.verified;
             }
             if (onSuccess) {
             	onSuccess.call(
--- a/gui.js
+++ b/gui.js
@@ -276,6 +276,22 @@ IDE_Morph.prototype.openIn = function (world) {
         function (username) {
             if (username) {
                 myself.source = 'cloud';
+                if (!SnapCloud.verified) {
+                        new DialogBoxMorph().inform(
+                            'Unverified account',
+                            'Your account is still unverified.\n' +
+                            'Please use the verification link that\n' +
+                            'was sent to your email address when you\n' +
+                            'signed up.\n\n' +
+                            'If you cannot find that email, please\n' +
+                            'check your spam folder. If you still\n' +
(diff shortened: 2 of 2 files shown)


brianharvey:
Looks fine to me.  Thanks.

jmoenig:
Is this an emergency patch? Do you want me to pull it now, make a minor release and put it online right away, or should we collect what else is coming up for a maintenance release next week or so? (Either way is fine with me)

jmoenig:
WAIT! Now it's sending me a message that my account is still unverified and asks me to click on a link from an email I never got. SOMETHING is VERY wrong here, I think I understand all the hubbub that's going on. Can we talk about this real quick in person?

jmoenig:
Oh, now I'm confused. I need to log out and log in again, and now I'm no longer getting the message. Hmm.... so everything seems to be right, except when users chose to stay logged in.

bromagosa:
No, don't worry, that happens because the SnapCloud object didn't have the
verified field when you accessed Snap!. Once you log in again, it will get
it from the server.

I couldn't think of a clean way to prevent this from happening to users who
were already logged in. We can revert this until we find a solution if you
think it's important.

2018-02-20 16:48 GMT+01:00 Jens Mönig <notifications@github.com>:

brianharvey:
I just tried connecting; I wasn't logged in before.  When I clicked login, 
I got the dialog.  As soon as I typed my login, before I had a chance to 
type my password, I got the account-unverified message covering the login 
dialog.  This is very disconcerting!  At least it should wait until I enter 
my password.

And, I'm a user who stays always logged in, so I want that to work.  :-)

bromagosa:
@brianharvey is this still happening to you? I just tried it at the Citilab with several accounts and nobody had this issue.

brianharvey:
Worked this time.  But I'm assuming it's just once per user, no?

brianharvey:
... but "stay logged in" isn't working.  I have to re-log in every time.

bromagosa:
See my email, please :)

2018-02-20 19:38 GMT+01:00 Brian Harvey <notifications@github.com>:

PR #2030, opened 2018-02-22
Changed reset password and resend verification dialog texts

bromagosa:
This was happening:

image

When trying to re-verify an already verified user, we are just informing the user and not sending any emails, so I removed the part that Snap! was adding to the message returned by the backend.

It now looks like this:

image

If we want to add more text, we should do that in the backend side.

I also removed the extra dot from the reset password dialog.


cycomachead:
LGTM, we should merge this. :)

jmoenig:
Thanks! I'll merge this when I get back from the conference I'm going to.

PR #2036, opened 2018-02-27
Add a CSP to upgrade http requests to https

cycomachead:
This will make it easier for existing projects that add images or JS libraries and, AFAIK, request http projects to load things when snap is served over https. 

**TODO**
When/If we serve snap.html over a plan http connection, we need to see if this also upgrades requests to local servers that aren't going to have SSL.


jmoenig:
Thanks! I'll merge this when I get back from the conference I'm going to.

PR #2040, opened 2018-03-02
Update the domain to cloud.snap.berkeley.edu

cycomachead:
- updates querystrings
- Also updates the domains picker for the cloud.
- the https one was broken before. (4431 is a port that works in development)


bromagosa:
Yay! This should fix all issues with third party cookies!

PR #2051, opened 2018-03-06
Update lists.js

michaelolifant:


jmoenig:
Huh, how didn't anybody notice this before?
Thank you!

Issue #2054, opened 2018-03-11
Sound Recorder: When clicking on 'Record' again,  it does not stop


boyzhirui:
When recording a sound, by clicking the 'record' button again, it won't stop. In Scratch, both 'Stop' and 'Record' button work when a user wants to stop recording.
![Sound Recorder Issue](https://user-images.githubusercontent.com/6066749/37255710-9787d8c8-2526-11e8-95cf-95c95bcca004.png)

PR #2055, opened 2018-03-11
Fix the sound recorder issue #2054

boyzhirui:
During the process of recording a voice, when the user clicks the 'record' button again, it won't stop. This pull request aims to fix this issue.


jmoenig:
Wow, good catch, thank you!

PR #2056, opened 2018-03-12
Lang files automatic generation

superruzafa:
When translating, usually one has to manually scan other lang-XX.js files (or the source code) to discover what strings were missed. This also lead to translate potentially non-active strings, propagating a bunch of crap amongst language files.

Also I observed that the same string could be translated up to 3 times:

- "input names"
- "input names..."
- "input names:"

Is for that I rewrote the i18n + l10n in Snap!

### Parameterized localizable strings

Literal strings are now translated with functions before display it (what was done far now was to translate the variable containing it in the target morph).
In addition to translate the string, these functions _mark_ it both as localizable and extractable. 

Current functions are (short naming is for readability):
- `_()` Translates and marks the string as extractable
- `_nop()` Does not translate but marks the string as extractable (string will be used somewhere else)
- `_expr()` Translates but doesn't mark the string as extractable (needed to translate expressions containing translatable strings, the former `localize()`)

The whole string contains now all information: no more translatable partial strings concatenated with hardcoded strings. Dynamic text is provided through placeholders. For more than one replacement you can provide a dictionary instead of a string.

```js
// Previous scenario:

SnapTranslator.dict.de = {
    'Are you sure you want to delete': 'Wirklich l\u00f6schen?'
}

localize('Are you sure you want to delete') + ' ' + proj.projectname + '?',

// becomes

Wirklich löschen? MyProject?
```

```js
// New scenario

SnapTranslator.dict.de = {
    'Are you sure you want to delete {{ projectName }}?': 'Wirklich l\u00f6schen {{ projectName }}?'
}

_('Are you sure you want to delete {{ projectName }}?', proj.projectname)

// becomes

Wirklich löschen MyProject?
```

- Strings ending with ellipsis (...) or colon (:) are very common. Those whose translation cannot be found are searched again by normalizing them, that means, removing the trailing symbols (... or :). You can write in source code "input names...", "input names:" or "input names" and the translator will only have to provide a single translation.

### Automatic generation of lang-XX.js files

I added a (hidden, experimental, shift needed) menu entry under "Languages..." named "Generate lang-XX.js file...". It triggers the source code parsing process, looking for functions named `_()` or `_nop` and extracting the first literal string argument (along with its location in the file).

After normalizing (you know, by removing trailing ellipsis (...) and colons (:)) and grouping by string they are merged with the current translated strings.

Finally a lang-XX.js file is generated with such as information. The strings used by Snap! for which there is no available translation are explicitly marked as `undefined`. The translated strings that are not used by Snap! are placed under the `deprecated` key.

```js
    // lang-de.js excerpt
    {
    'Replace Project': // gui.js:6438 gui.js:6456
        undefined,
    'Are you sure you want to delete\n"{{ projectName }}"?': // gui.js:6498 gui.js:6522
        'Wirklich l\u00F6schen?\n"{{ projectName }}"?',
    'Delete Project': // gui.js:6500 gui.js:6524
        'Projekt l\u00F6schen',
    }
````

### Other improvements

- Language files structure has been changed. A dictionary has 2 entries: `metadata` (for information about the translation) and `strings` containing the translations themselves.
- Translations clean up
    - removed trailing symbols: (... … : .)
    - removed spaces around \n
    - fix typos
    - trim strings
    - shrink multiples spaces into one
    - short messages, label, hints... does not end with full-stop anymore
- Metadata's translators is now an array containing translator information in the form `Name [<email>] [(website)]`
- The generation process uses the [acorn library](https://github.com/acornjs/acorn) for build the source code AST.
- The generation process is fully asynchronous through `Promise`s! (Tested OK in last versions of Firefox, Chrome, Opera and Vivaldi in Mac)
- Added `containsKey(Object, string)` and `isArray(array)` global functions.

### TODO, ideas...

- Get rid of the "\n" in the middle of the strings and provide something like `wordwrap(string, 40)` in the source code to format long lines.
- Rething Snap! UI messages.
- Translator notes are lost :(
- Add generation options (whether encoding strings or not, include source locations...)
- Add a `_dev()` function to tag development mode only strings (translators has a lot of stuff to do and they prefer to translate user strings first)
- Provide a translation percentage progress

**TL;DR:** Automatically generate an always up-to-date lang-XX.js file while giving the translator more freedom to write more accurated translations.

### CODE REVIEW

**Comment by hasso** [comment]
`removed trailing symbols: (... … : .)`

While I generally in favour to pull request, I strongly object against this. This is one of few things providing context to the translations and often require different translations in languages with different logic from English. Obvious case is that there is no difference in string "edit" whether it's a verb or noun, but many languages have to make difference. And I'm not even talking about Fenno-Ugric languages (Finnish, Hungarian and Estonian) where verbs and nouns could be in many forms depending where they appear in UI ([reference](https://9gag.com/gag/a2qAMje/the-word-dog-and-its-forms-in-finnish-and-hungarian) - it's only slightly exaggerated).

Ideally there should be something like [msgctxt in gettext](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html) to give translators the context even if a string in English doesn't.

PS. I'm Estonian translator.

**Comment by superruzafa** [comment]
Interesting.

Sadly,
- the Snap! strings were written with english in mind (along with its the gramatical rules)
- the translation system uses a key/value dictionary and once you translate a word, its meaning will apply to all those places that use it (the msgctxt context here is "global" 😄 )

I removed all ellipsis and colons from translations because I found that in the vast majority of the cases the same string (including trailing symbols or not) was translated in the same way. That was a waste of time for translators, I believed...

However, I didn't removed ellipsis nor colons from the source code: If Snap! finds a string like "edit..." it first would look up for a translation whose key is "edit...". If none found it would ask for a translation with key equal to "edit" (it'd remove the trailing ...) and eventually would append "..." at the end.

Taking adventage of this and as a workaround, I could track that information to notify the translator that those strings are also used with "..." and/or ":" in case he/she would to provide a more accurated, additional translation. If no provided, the canonical string (the one without trailing symbols) would be used.

```js
    "edit": // byob.js:1157 gui.js:7699 objects.js:3255
        "Muuda",
    // "edit...": // uncomment this to override the "edit" translation // morphic.js:8730 morphic.js:9477
    //     undefined,
    // "edit:": // uncomment this to override the "edit" translation // morphic.js:8730
    //     undefined,
```

Anyway, as I mentioned above, the Snap! UI messages need to be rethinked: think about all combinations of "edit" in lower and uppercase, with and without "...", ":", "?"

Bonus: Here are some weird cases I found:

- Sometimes the ellipsis mark was left out of the translation like in `localize('find blocks') + '...'`, even when as you said this would change completely the meaning of the purpose of that string.

- For example, I found that the string "unused block(s) were removed", without any context, was often translated to something like _"deleted all non-used blocks"_. That message is prepended in code with the number of blocks that has been actually removed and the final translated string, while makes sense in english, is shown to the user as "13 _deleted all non-used blocks_"

- Extremely uncontextualized strings like "do it" make that you have to find the place (or places!) that string is used to imagine what the developer would have in mind when he/she wrote it (by my experience, "do it" applies to TextMorph's `eval()` action).

jguille2:
Gracias Alfonso @superruzafa,

I think this is very interesting. Maybe it won't be a quick implementation... but we can take the opportunity to reopen this outstanding issue. Some comments:

- Sometimes, I also miss _translate contexts_ for a more detailed work... but I think it is not necessary and we can survive with current dict system.

- You said "UI messages need to be rethinked" and yes! For me, the key is that these _translatable strings catalog_ is not only a reference for translators... It must be the reference for creating/editing UI messages. I can not create a new string without looking at this reference catalog (if it already exists, if there is something similar...)
  If this reference is done and applied, these problems with trailing symbols and `\n` will be gone... because some format issues must be out of the translation system. In this sense, I like your suggestion of  something like `wordwrap(string, 40)` to avoid all these `\n`...
 
- And I add the suggestion to implement an English translation. It can help to improve English messages without changing code strings (and also, affecting all translation files). And later, it would cause a separation between "code messages" and English messages... and this can be useful in the sense you said above (differentiation in messages required for other languages...)
  As usual, strings not translated show their 'code string'. So, this English translation would not be an obligation ... but a new possibility.

superruzafa:
Merci @jguille2 

The good news is that even an english translation from... english 🤔 can be done.
In master:locale.js there is an example of english rewordings:

```js
// rewordings in English avoiding having to adjust all other translations
'any':
    'random'
```

The problem is still the same: one string -> only one translation regardless the context. But it's not the only one.

In the Catalan translation, for example, there are two entries for "duplicate":
```js
// blocks:
'duplicate':
    'duplica\'m',
...
// Developer mode menus
'duplicate':
    'duplica',
```
The latter remains. The meaning is similar and understandable by the user but this not needs to be a rule for the rest of languages.

I also discovered that a (generic, contextless) string is used only with a given context in mind and so the translation could be done more accurately, like in:
```
'none': // byob.js:3875 objects.js:5421 objects.js:5427
    'cap bloc', // read as "no block at all", even when "none" didn't make any reference to blocks
```
While this demonstrate the translator's concern to providing an accurate translation, the problems would arise when in the future someone will make use of `localize('none')` in a complete different context.

Some people made literal translations, even for the _longer lookup strings_:
```
"costumes tab help": // it should be resolved as 'import a picture from another web page...'
    'ayuda de la pestaña de disfraces': // means "help for the tab of the costumes"
```

The problems I tried to address with this PR was to _gettextize_ Snap! as much as possible (a slightly different approach to the one you discussed at #1551):

- Autodiscover all translatable strings (solved by marking strings with `_()` and avoiding dynamic translations)
- Get rid of partial translations leading to weird localized strings (solved by used {{ parameterized }} strings instead of hardcode concatenation)
- Get rid of lookup strings (as Android does AFIK)
- Give the translator a better context (same point that above + give him/her the file:line in the source code where the translation is requested + gather by order of appearance, not alphabetically)
- Automatic generation of lang-XX.js files (automatic -> good)
- Avoid similar translations (edit, edit..., edit ) **this has been possibly a mistake**, but see my previous comment about override similar translations)

By fixing or reverting the last point, I strongly believe that this PR is a good thing because it enhances the translator experience at minimal cost for the developer.

The strings in the code should be more verbose/better contextualized. I agree with you that an additional english translation would address this problem.

PR #2057, opened 2018-03-13
remix project method

bromagosa:
A minor addition to cloud.js that calls the new `remix` API endpoint.

This won't affect Snap! at all, and isn't urgent.


jmoenig:
Thanks, Bernat. I need to make a quick emergency release anyway (because pulling a recent bug fix had the side effect of dramatically slowing down list operations, whew), so I might as well pull it now.

PR #2072, opened 2018-03-25
Update byob.js

michaelolifant:


cycomachead:
Can you describe more about what problem this fixes? Right now it looks as if this won't word because it removes detecting quotes.

PR #2084, opened 2018-04-20
Programmatic declaration of permanent variables

brianharvey:
Modifies the SCRIPT VARIABLES block so that it can be changed into
GLOBAL VARIABLES or SPRITE-LOCAL VARIABLES.

The purpose is for libraries that want to initialize variables for
later use by the user of the library.  Paul asked for this.

### DIFF
--- a/blocks.js
+++ b/blocks.js
@@ -1072,6 +1072,18 @@ SyntaxElementMorph.prototype.labelPart = function (spec) {
                 true // read-only
             );
             break;
+        case '%vartype':
+            part = new InputSlotMorph(
+                null, // text
+                false, // numeric?
+                {
+                    'script' : ['script'],
+                    'sprite-local' : ['sprite-local'],
+                    'global' : ['global']
+                },
+                true // read-only
+            );
+            break;
         case '%spr':
             part = new InputSlotMorph(
                 null,
--- a/objects.js
+++ b/objects.js
@@ -1123,7 +1123,8 @@ SpriteMorph.prototype.initBlocks = function () {
         doDeclareVariables: {
             type: 'command',
             category: 'other',
-            spec: 'script variables %scriptVars'
+            spec: '%vartype variables %scriptVars',
+	    defaults: ['script']
         },
 
         // inheritance - experimental
--- a/threads.js
+++ b/threads.js
@@ -1445,11 +1445,21 @@ Process.prototype.evaluateCustomBlock = function () {
 
 // Process variables primitives
 
-Process.prototype.doDeclareVariables = function (varNames) {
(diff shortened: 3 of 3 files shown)


jmoenig:
Hmm... this code is technically correct but runs into a few issues. Existing projects need to be migrated to the new ```script vars``` block version, otherwise these blocks will be labelled as ```obsolete```. Same with translations, all of them need to be updated. These are technicalities and can be done, but I'm still not convinced that this won't scare users. We go to great lengths showing variable watchers for newly created variables and checking for name collisions for existing ones in other non-local scopes (although I disagreed about this back then). Then there's the question of how to get rid of programmatically created variables again, shouldn't there also be a ```delete variable``` option?

All in all to me this appears to be more of a convenience-hack for curriculum developers than something our "normal" users will want (although I do love and appreciate hacks, haha...). How about we make a little "meta-programming" library with two blocks that let you programmatically create and delete global and sprite-local variables? We could even put these into your "Eisenbergification" library...

brianharvey:
I wanted to do that in the first place, but there's no way to create variadic upvars (and upvars aren't exactly the right thing anyway) in a custom block.

We could make them separate blocks, to avoid messing up SCRIPT VARIABLES, and hide them behind a check box in the gear menu?

towerofnix:
>but there's no way to create variadic upvars in a custom block

This doesn't sound that technically difficult to do - just have an "upvars..." parameter block you can drag out, which is a list where each item corresponds to the value of one of the upvars. Is there any reason this hasn't been done yet, besides "it's on the todo list"?

brianharvey:
There's an issue about the default naming of the variables.  For this block there's no issue about the _values_ of the variables; we need access to their names.  I just had a conversation with Jens in which he suggests just text-string inputs rather than variables, since the actual variable orange ovals will appear in the Variables palette.

For other uses of variadic upvars, I think the name given by the user in the block definition has to be treated as a prefix, so the actual default variable names as seen outside the block will be FOO1, FOO2, etc.  And, as you say, inside the block FOO will be a list, not of values, but of first class variables, a new data type that would have to be draggable over the pulldown menu in SET, CHANGE, etc.  So it's not trivial to invent, and not clear there's actually a need for it.

gdavid04:
Why not allow typing in a variable name or using a reporter to determine which variable to get, set, show, hide?

blocks.js
```javascript
case '%var':
    part = new InputSlotMorph(
        null,
        false,
        'getVarNamesDict',
        false
    );
    break;
```

I'll insert the getter implementation later...

jguille2:
Hi!
I think we can close this PR because you pushed the _make-variables.xml_ lib .

That's really useful Thanks!

Joan

PR #2092, opened 2018-05-09
updated German translation

jadga-h:


jmoenig:
Thanks, Jadga, I'll pull this together with the next batch of changes!

PR #2094, opened 2018-05-20
Small updates to the Portuguese (pt) translation.

MMSequeira:


jmoenig:
Thank you Manuel! I'll pull this together with the next batch of changes!

PR #2118, opened 2018-06-09
Add download for source and libraries

gdavid04:
Added view source on GitHub and download source and libraries from GitHub buttons.

### DIFF
--- a/gui.js
+++ b/gui.js
@@ -2506,6 +2506,24 @@ IDE_Morph.prototype.snapMenu = function () {
             );
         }
     );
+    menu.addItem(
+        'View source on GitHub',
+        function () {
+            window.open(
+                'https://github.com/jmoenig/Snap--Build-Your-Own-Blocks',
+                'SnapSourceViewGitHub'
+            );
+        }
+    );
+	menu.addItem(
+        'Download source and libraries from GitHub',
+        function () {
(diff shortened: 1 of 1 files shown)


cycomachead:
IMO, we should go from 3 links in the menu to one that points here:
https://github.com/jmoenig/Snap--Build-Your-Own-Blocks/releases/latest

The zip file linked there will always be the version the user is using, and subsequently allows people to get to GitHub, if they desire.

jmoenig:
Can you check whether this suggestion has now been implemented in a way that makes sense to you? Thanks!

brianharvey:
I vote no on removing the links to the manual and the website!  Many people who feel the same way I do about git will nevertheless want to read the manual.  And the link to the website will be super important once the social site is up.

Speaking of which, when it _is_ up, I propose keeping the existing page as a sort of super-about linked to from the About box ("Click for more information").

cycomachead:
No one was suggesting removing the links to website of manual (by 3 links, I meant that this PR adds 2 more that point to GitHub.)

Jens recent change addresses the issue, I think.

cycomachead:
Thanks for the pull request, gdavid04, but if you feel like there's something else we can do to make finding the repo easier, let us know!

Issue #2108, opened 2018-06-02
Feature Request (easy, low priority): letter ( ) of ( ) should have the same 1,last,random as item ( ) of ( )


ddgarcia:
This wouldn't change anyone's existing use of the block (hopefully), but people who would use it in the future would find it works just like item, giving easy access to the last letter in the sentence and a random element in the sentence...

jmoenig:
Good idea, Dan, I like it. Assigning myself to it...

jmoenig:
... of course, thinking about this some more, it's the old ```length of``` discussion all over. We could simply eliminate either the green reporters in the ```operators``` category or the red ones in ```lists```. Personally I'd opt (again) to get rid of the red list ones...

brianharvey:
Nyet!  (I'm in the middle of a book about how the former soviet republics adapted to capitalism.)

Yes, a string is an array of characters, and so all the operations on strings have corresponding operations on arrays.

If we wanted _users_ to think of strings as arrays, we would invent another datatype Character to be used in blocks such as Unicode conversion.  How you arrange primitives depends on how you want users to think, not on how things are implemented.

But we want users to think about strings as _words_, or as _sentences_ if they're made of words.  That's why the green block is named LETTER rather than ITEM; the latter would be ambiguous, because it might mean the third letter or the third word.  (It also explains why LETTER and not CHARACTER.  We want users to think about words and sentences.)

In a perfect world, sentences would be implemented as lists of words.  But lists don't _look_ like sentences; they're these funny boxes with the items arranged vertically.  So we use strings for both words and sentences, and that's why we can't have generic functions FIRST and BUTFIRST that look at the type of their input to know whether to report the first letter or the first word.  We can't distinguish a word from a one-word-long sentence by looking at the internal type.  So we need FIRST LETTER and FIRST WORD functions.

In God's language, we have LIST-REF and STRING-REF as separate functions.

jmoenig:
Okay, okay. OTOH I do believe that having fewer blocks would make Snap! less intimidating and more inviting, and one way to achieve this would be to let blocks handle closely related / similar operations on different data types, such as ```item / letter of``` and ```length of```. But I get it that you disagree and I'm not going to push it. How do you feel about Dan's original suggestion?

brianharvey:
Sure, that's fine.

[...]

PR #2121, opened 2018-06-12
Updating letter block translations

jguille2:
Hi,
As I said in #2108, I've done the updates for the translation files.

Only two comments:

- Tamil (lang-ta) translations had a problem with the order of the variables.  I've tried to fix this.
- lang-pt_BR had codification problems. I convert its unknown codification to UTF-8


jmoenig:
Thank you sooooo much, @jguille2 , this is a *big* help!! You're fantastic!!

PR #2123, opened 2018-06-15
Automatically Determine the Cloud URL

cycomachead:
This will make it easier to host Snap! on different domains, like cs10.org
or on a staging server.

What it will do is automatically lookup the URL that matches the domain.
Additionally, we can now specify a meta tag to force the snap-cloud URL to be
a different domain. This will be helpful for developing (where Snap! can be loaded from localhost) or if 3rd party users host their own cloud APIs.


cycomachead:
I pushed this to http://cs10.org/snap/snap.html -- you can look at `SnapCloud.url` and see that it picked the CS10.org domain from the list, since that's the best option there.

jmoenig:
Thank you, Michael, this looks good to me. How about you, @bromagosa ?
Do yo want me to pull it and include it in the upcoming release?

jmoenig:
(does it work offline, too, like, when I run Snap totally without any web server?)

cycomachead:
> (does it work offline, too, like, when I run Snap totally without any web server?)

Offline, it defaults to berkeley. If you run a localhost server, it will default to berkeley _unless_ the port is 8080, which is the default for the cloud server in testing. We can remove that if others find it problematic and have the cloud server set the meta tag. 

Pulling it in 4.2 would be nice if there will be a gap -- it will help with some website stuff that might happen soon.

snap.html
```diff
@@ -25,6 +25,7 @@
 		<script type="text/javascript">
 			var world;
 			window.onload = function () {
+        SnapCloud = new Cloud(Cloud.determineCloudDomain());
```
cycomachead: Because this is global, it seemed a little more obvious to be here than slightly hidden in cloud.js (since most other files don't create globals that are instances.)

jmoenig: yeah, I wouldn't mind if it's in ```cloud.js```, in fact I might even prefer it there, keeping the current ````snap.html```` the way it is now, very concise with as little code as possible...

jmoenig:
Okay! Looks good. Let me know when it's ready to pull. I'm planning to make the release sometime Monday morning my time so it's up when school starts in the U.S. (and when the Berkeley servers are hopefully up again, too)

bromagosa:
Looks good to me too :)

jmoenig:
Thanks, Bernat!
@cycomachead if you get around to it, I'd prefer initializing the global ```SnapCloud``` reference in ```cloud.js```  rather than in ```snap.html```` itself. Otherwise I can do it...
Thanks, all!

cycomachead:
The challenge with moving that is that `determineCloudURL` needs to be called once the `document.head` is available to read from. 

Otherwise, it's not clear to me that `SnapCloud` even needs to be global. It could just be `this.cloud` on the `IDE_Morph`, since the IDE is the only object that would need to access the cloud.

jmoenig:
hmmm.... are you suggesting we make SnapCloud a property of the IDE? That would be fine with me!

cycomachead:
Yeah, let me try something real quick. :)

jmoenig:
looks like the only other thing we'd need to tweak would be @brianharvey 's Eisenbergification library (to get the username, which would simply be deferrred to the IDE)...

cycomachead:
Ok...that was a bit more work than I expected. I'll push the library change in a second. 

I have noticed two things:
* the dev/ version seems to be broken right now
* try opening a saved project, then doing a save-as -- I get an error, and I don't think it's from this code.

jmoenig:
Thanks, Michael!
@bromagosa , the issue Michael is observing seems to be related to trying to display the recover button. Could you maybe have a look at this, is this something I broke? Thanks!

bromagosa:
Ouch! How did that happen? O_O

Let me fix it, one minute.

jmoenig:
Okay, thanks, guys! I just pulled both of your PRs, but abbenay at the UC Berkeley server seems to be already down or currently not working, so I cannot update the dev version there. bummer. @brianharvey : The Berkeley web server seems to be still up, so maybe abbenay not being available is a different issue?

brianharvey:
Oops.  I'll look into it this afternoon, but I'm going to have to turn it off for the power service anyway.

PR #2126, opened 2018-06-16
Fixing storage issues with created svgs

### LINKED ISSUE #2125: Problems with svg images generated by vector editor
Hi!

I reported (by mail) some issues about new Vector Editor... but it was still in process. Now I see the _v4.2 RC_ commit and then, I want to report it here... and I will try to fix them (I think it was only a problem with svg headers).

- I can create svg costumes and edit svg costumes created by vector editor. But I can't edit other svgs (from Snap! library or imported)
- I can export svg costumes created, but the file is not right. I can't edit it with a desktop vector editor.
- And so, exported svg costumes can not be imported in other projects (they are not fine svgs).

I'll take a look about this. Do you have a date for the 4.2 release? Depending this date I'll try to check more things... and PR a Catalan language update.

jguille2:
These little changes fix #2125 
- As Jens said, general svgs (also svgs from Snap! library) are (for now) not editable. We can only edit svgs created with Snap! vector editor.
- But now I can export "snap svg costumes" with a right format. So...
  - Exported "snap svgs" can be imported to other projects (and they are still editable).
  - I can open all exported svgs with a desktop vector editor.
  - Svgs edited with a vector editor can be imported into Snap! projects... but now they are "general svgs" (not created by Snap!) and then, they are not editable.

This is all!

### DIFF
--- a/sketch.js
+++ b/sketch.js
@@ -138,7 +138,7 @@ VectorShape.prototype.imageURL = function () {
 
     svg.children = [ this.asSVG() ];
 
-    return 'data:image/svg+xml, ' + svg;
+    return 'data:image/svg+xml;base64,' + svg;
 };
 
 VectorShape.prototype.copy = function (newShape) {
@@ -1525,7 +1525,7 @@ VectorPaintEditorMorph.prototype.getSVG = function () {
 
     svg.children = this.shapes.map(function (shape) { return shape.asSVG(); });
 
-    return svg;
+    return window.btoa(svg);
 };
 
 VectorPaintEditorMorph.prototype.getBounds = function (shapeCollection) {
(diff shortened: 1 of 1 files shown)


brianharvey:
@jmoenig @jguille2 @bromagosa What are our long-term plans about SVGs?  Is someone sometime going to make it so we can import and edit arbitrary costumes as vectors?  This is an area in which we are dramatically inferior to Scratch, so it should be a high priority for Jens. :-)  If it's too hard to write our own, maybe we should think about third-party code?

jmoenig:
Thanks, Joan!

@brianharvey : Currently our strategy is to support arbitrary SVGs as costumes, which we've done for a long time. But now we also have a real SVG sketching editor that makes "real" SVGs, but with a limited set of shapes/primitives. Expanding that set of primitives is "just work". If someone creates an SVG using only our shapes that can be edited in our editor.

In the mid- to long term my priority wouldn't be so much extending the vector paint editor, but supporting SVG pen trails. *That's* much cooler, and I'm really impatient for it!

PR #2129, opened 2018-06-21
Fix to "my" translation

MMSequeira:
It was breaking the "set" block. This is urgent!


jmoenig:
thanks, Manuel! I'll pull and integrate this within the next days.

PR #2138, opened 2018-07-03
Fixing cache issues with url block

jguille2:
Hi,
After some Snap4Arduino issues about URL block (like [this last one](https://github.com/bromagosa/Snap4Arduino/issues/247) I've been studying this.

Some behaviors are not predictable... and sometimes the issue is in the Server side... but I see many times Snap requests have a "Caution: Provisional headers shown" warning... and this is a bad behavior because it was returning cached values without any request to the server.

I see browsers (Firefox and Chrome) has not this issue, and they have a `Cache-Control: max-age=0` header. Then I've tested, and adding this header to our request (into URL block) the issue is gone.

Thanks

Joan

### DIFF
--- a/threads.js
+++ b/threads.js
@@ -2301,6 +2301,7 @@ Process.prototype.reportURL = function (url) {
         }
         this.httpRequest = new XMLHttpRequest();
         this.httpRequest.open("GET", url, true);
+        this.httpRequest.setRequestHeader('Cache-Control', 'max-age=0');
         this.httpRequest.send(null);
         if (this.context.isCustomCommand) {
             // special case when ignoring the result, e.g. when


jmoenig:
Excellent analysis, @jguille2 , thank you!

jmoenig:
Ugh, @jguille2 , we're running into some serious issues with this code. Apparently it's causing other local servers, such as Birdbrain's robot server for the Finch robot and Hummingbird kits to throw CORS exceptions. Removing the cache-control headers "fixes" these issues again. Since this affects a large portion of our users I'm going to have to comment the cache-control header out again for a quick maintenance release. I really hope you can find another way to address the issue in Snap4Arduino. Which browsers did you say exhibit the unpredictable behavior?

bromagosa:
No browsers, Snap4Arduino runs on desktop over nwjs.io. We can always add this as a decorator to the URL block in Snap4Arduino, don't worry about it :)

jguille2:
Yes @bromagosa,

About browsers, the idea was add the same header to the _xmlhttprequest_ that browsers add in their "manual" requests (location submits).

I guess the problem with the Birdbrain devices could be in the server side (with BirdBrainRobotServer responses) but, of course, it is important that those devices run ok.

I will re-study this... but I will have to check BirdBrain devices before submitting anything. The idea is not to use a decorator... because it was reported in Snap4Arduino, but it was a Snap! issue (_url block_ sometimes showing cached data without any request to the server).

I'll continue... Bernat, maybe I will ask you if you have any BirdBrain device to test it.

Joan

PR #2172, opened 2018-08-07
Some Speedups to Importing Blocks

cycomachead:
Eliminates a bunch of redraws which checking for duplicate blocks.

I really hate adding `silently` as a control parameter, but it seems like the common thing to do?


cycomachead:
boo. :( I started off the wrong branch... will fix later.

cycomachead:
This seems to no longer be necessary!

PR #2178, opened 2018-08-09
Keep tools.xml in the libraries/ directory.

cycomachead:
we should only need to serve one copy, this just updates the path Snap!
uses when it looks to import tools from the menu.


cycomachead:
I also decided to fix the label stage issue so it gives you a usable error message now.

jmoenig:
Thanks, Michael, this looks good!

PR #2220, opened 2018-10-03
Fix bug in letter of with numeric inputs

cycomachead:
Coerce input to string before checking length in LETTER OF

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -2706,14 +2706,14 @@ Process.prototype.reportLetter = function (idx, string) {
     if (string instanceof List) { // catch a common user error
         return '';
     }
+    str = isNil(string) ? '' : string.toString();
     if (this.inputOption(idx) === 'any') {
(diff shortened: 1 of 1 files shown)


src/threads.js
```diff
@@ -2706,14 +2706,14 @@ Process.prototype.reportLetter = function (idx, string) {
     if (string instanceof List) { // catch a common user error
         return '';
     }
+    str = isNil(string) ? '' : string.toString();
```
jmoenig: we probably want to declare this as a local variable (```var str 0 ...```), so it doesn't accidentally become global...

jmoenig:
Thanks for the quick fix, Michael!

src/threads.js
cycomachead: I think github cuts it off from the view, but the bar is rare in the existing definition. :)

jmoenig: wow, you're right. That's funny!

Issue #2229, opened 2018-10-23
Typo in help for 'sum' from variadic library


ToonTalk:
Takes any number of numbers as inputs (use the left and right arrowheads to adjust the number of input slots) and reports the result of adding them all, so
ISUM (4) (100) (8))
reports 112.

should be

(SUM (4) (100) (8))

brianharvey:
Oops!

jasonappah:
Should this be closed?

PR #2232, opened 2018-10-23
Typo in help for sum from variadic library

belushkin:
I fixed typo in issue #2229


brianharvey:
Thanks for calling this to my attention.  But the proposed fix isn't quite right either; the errant I should be replaced by an open parenthesis, not just deleted.  I'll install my fixed version in the repo.

PR #2233, opened 2018-10-25
Add predicate for the existence of a variable

kephas:


kephas:
It's too bad that dependencies are a problem. Maybe adding proper modules could be a priority? (I know it would help me **A LOT**)

brianharvey:
As soon as things slow down :-) we're going to make "export blocks" export the transitive closure of the set you select.  That will solve the problem, I think.

kephas:
But is there any plan for going further? Something like folders in the IDE would avoid having dozens of unrelated blocks mixed together. Also, blocks could use variables that wouldn't clutter the user's namespace.

brianharvey:
Umm.  I wouldn't say a _plan_ exactly, but we certainly want to improve the state of extensions to Snap<i>!</i>.

Issue #2244, opened 2018-11-01
Unclear if there is a programmatic way to make a clone permanent


ToonTalk:
I see the documentation for how to make change a clone from temporary to permanent but don't see a way for a program to do so.

kephas:
If you run this to produce a deleted sprite:
![untitled script pic 1](https://user-images.githubusercontent.com/762421/47858859-3de1dc80-dded-11e8-919f-15533426be66.png)

The following code should report true:
![untitled script pic 2](https://user-images.githubusercontent.com/762421/47858912-5b16ab00-dded-11e8-9e19-666d4b694a8a.png)

(you'll find `safely try` in the **Catch errors in a script** library)

brianharvey:
Responding to the original issue:  This is a deliberate decision by Jens.   Permanent clones are much more costly in efficiency than temporary ones, mainly because of having to update them in the sprite corral.  So Jens wants you to have to make each permanent clone deliberately, not have your program create a zillion in a loop.

ToonTalk:
Here's my use case. The temporary clones are board positions (9 or 64 typically) and if while playing the game one saves the project the clones are destroyed. Or someone presses the stop button and then wants to play some more.

jmoenig:
yes, the problem I'm trying to avoid is a sorcerer's apprentice one, where a script creates a zillion permanent clones that can't be deleted programmatically. Now, y'all are going to say "just let permanent clones be deletable programmatically", but that's going to make kids loose their work really quickly and unrecoverably.

In the meantime, there *is* an experimental way to change the "temparary" setting of a clone, but it's only available in dev mode:

<img width="191" alt="screen shot 2018-11-02 at 12 42 24" src="https://user-images.githubusercontent.com/2596285/47913659-2feb9480-de9d-11e8-8bb5-4b26af48ae2f.png">

I put it in for exactly the use case you're describing, where you want to create 64 chess figures, and don't want to do it all by hand.

Das that help?

ToonTalk:
Yes, it does, thanks.

My personal preference is that rather than disabling dangerous features to put in the effort that warnings are produced when they are used and when they cause trouble that a clear error message helps users avoid the problem next time. E.g. "Warning: Snap! is nearly out of memory [or running very slowly] because you've created 23...45 sprites and turned off their temporary status". But I'm not suggesting that addressing this should have higher priority than many other outstanding issues especially because it might not be easy to implement.

PR #2245, opened 2018-11-01
Added two predicates that report if code raises an error

kephas:


kephas:
Because I had to use that idiom several times in the last weeks.

brianharvey:
I'm sure the use case is vivid in your mind, but without an example, to me what you said is just "it's important because I think it's important"!  Could you send a project URL?  Thanks!  :)

kephas:
I actually used it in #2233 to test the existence of a variable without JS and in [my comment](https://github.com/jmoenig/Snap/issues/2244#issuecomment-435064084) to #2244.

brianharvey:
I guess my question would be whether your solution is more or less comprehensible than using SAFELY TRY explicitly.  Let me think about it...

PR #2279, opened 2018-12-04
added Cloud >> updateProjectName method

bromagosa:
Needed to allow users to rename their projects from the social site.

### DIFF
--- a/src/cloud.js
+++ b/src/cloud.js
@@ -749,3 +749,22 @@ Cloud.prototype.updateNotes = function (
     );
 };
 
+Cloud.prototype.updateProjectName = function (
+	projectName,
+    newName,
+    onSuccess,
+    onError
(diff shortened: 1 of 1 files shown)


src/cloud.js
```diff
@@ -749,3 +749,22 @@ Cloud.prototype.updateNotes = function (
...
+    onError
+) {
+    this.withCredentialsRequest(
+        'POST',
+        '/projects/%username/' +
+            encodeURIComponent(projectName) +
+            '/metadata',
+        onSuccess,
+        onError,
+        'Could not update project notes',
```
cycomachead: Shouldn't this say, "Could not update project name"?

bromagosa: It should indeed. Thanks for noticing!

PR #2287, opened 2018-12-11
fix: can't play sounds by name if name is a number

bromagosa:
This used not to work:

imatge
number sound test script pic

This PR fixes it :)


jmoenig:
Thanks, Bernat!

PR #2289, opened 2018-12-12
Resolved conflicts in PR #544 (volume blocks)

bromagosa:


jmoenig:
Nice! This seems like a bunch of big changes, we'll want to thoroughly test them, right?
Did you add a mute button? Are you sure we want our own mute button rather than the one that's on your computer?
Thank you, I'm looking forward to playing with this.
Another question: Is the volume global, per sprite or even per script?

bromagosa:
I haven't really looked at the code in depth, I just resolved the conflicts
on this old PR by @gubolin so it can be merged again and reviewed. It needs
testing, especially given that Snap! has changed quite a lot since this PR
was originally issued (instruments were added, for example).

I would not add a global mute button, nope. That's what [ stop all sounds ]
is there for :)

(I believe the volume is like in Scratch, per object.)

jguille2:
Hi,
Some comments about this volume/mute feature:

- I see "volume" is here an sprite property. Then "play sounds" and "play notes" are modified to use it. But now, we have some libraries ("Text to speech" and "Audio comp") that play sounds. So we have to decide if this "volume" is only for "Sounds" (like "size" is for the costumes) or it is really an sprite property, and then, we will have to update these libraries to add the volume feature.

- We have to check that code, because it stores volume property into the Sound object. This is not right, because the volume is only a sprite property. This problem causes, for example, that "Import sounds" feature is broken with this PR. When you want to import a Sound (for example, from the sounds library), the import action is broken because "volume" property is not defined. The problem is in the "Sound.prototype.copy" function.

- About mute, different comments:
  - First, the same comment about "other sounds" in Snap. If we want a really "mute" feature, we have to consider all Snap! sounds (text to speech.... and even the play action into the Sounds tab.
  - I'm not sure about this mute feature... but maybe it is useful thinking in the embed presentation (for our next social web). If we think in pages with different Snap! projects, maybe a "mute" feature can be useful. Then I like a button and also a url flag for the initial value (mute or not).
  - A last comment... maybe the "mute" icon is not clear.

Joan

bromagosa:
Closing as this feature has been added to dev already :+1:

PR #2292, opened 2018-12-18
Changes needed for the remix functionality in our new social site

bromagosa:


bromagosa:
Here's what the project page looks like for a project that has been remixed:

imatge

And here's what it looks like for the remix:

imatge

brianharvey:
If you remix a remix, who gets credited up in the corner of the remix's page?  If I have to pick one, I'd go with the original author, although ideal would be

remixed from FOO by UserN ... from FOOLY by User1.

That's an actual ellipsis!  So we credit the most recent author and the original author, no matter how many remixes are in between.

bromagosa:
Good point! This affects the backend and site though, not the editor.

PR #2293, opened 2018-12-18
add string library

brianharvey:
substring by index or by content.  Sets a global flag for case-dependent vs case-independent search, in parentThatIsA(IDE_Morph). I hope that's okay...


jmoenig:
Nice, thank you!
I don't like having a library add a global variable, but in this case it's really just another attribute to the IDE, so, since it's you, Brian, it's okay :-)

PR #2309, opened 2019-01-16
minor change because of new user roles

bromagosa:
This is a tiny change that does not affect the editor. We're moving to a more complete user permission system for the social site, where users have roles.


jmoenig:
sure (great!!). I'll pull it in about an hour. Thanks!

brianharvey:
What's a "user role"?

jmoenig:
I guess something allows for more variety than can be expressed by a Boolean :-)

bromagosa:
Yep. We have admins, moderators, reviewers, regular users and banned users, in order or godness.

PR #2315, opened 2019-01-24
Updates to cloud.js

cycomachead:
* Add the /api/v1 prefix to API routes
* fix file size checking
* make cloud.js work independently of morphic / rest of snap
* and ended up reducing some duplication in the gui for saving projects.

### DIFF
--- a/src/cloud.js
+++ b/src/cloud.js
@@ -376,7 +376,6 @@ Cloud.prototype.changePassword = function (
         onError,
         'Could not change password'
     );
-
 };
 
 Cloud.prototype.resetPassword = function (username, onSuccess, onError) {
@@ -403,6 +402,8 @@ Cloud.prototype.resendVerification = function (username, onSuccess, onError) {
 // Projects
 
 Cloud.prototype.saveProject = function (projectName, body, onSuccess, onError) {
+    // Expects a body object with the following paramters:
+    // xml, media, thumbnail, remixID (optional), notes (optional)
     var myself = this;
     this.checkCredentials(
         function (username) {
--- a/src/cloud.js
+++ b/src/cloud.js
@@ -51,6 +51,9 @@ Cloud.prototype.init = function () {
     this.username = null;
 };
 
+// Projects larger than this are rejected.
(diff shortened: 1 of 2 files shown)


src/cloud.js
```diff
@@ -28,10 +28,12 @@
 */
 
 // Global settings /////////////////////////////////////////////////////
+// cloud.js should be able to exist indepent of Snap!
```
cycomachead: cloud.js exists for more than just Snap<em>!</em> now. 

I'm starting to think this file might belong in it's own repo (like morphic).

src/gui.js
```diff
@@ -5583,17 +5583,84 @@ IDE_Morph.prototype.logout = function () {
...
+    };
+    this.serializer.isCollectingMedia = false;
+    this.serializer.flushMedia();
+
+    return body;
+}
+
+IDE_Morph.prototype.verifyProject = function (body) {
+    // Ensure the project is less than 10MB and serializes correctly.
+    var encodedBody = JSON.stringify(body);
```
cycomachead: You'll note that the 10MB limit is on the entire body not just the project + media. Mostly this means we lose ~.4KB to the thumbnail + JSON compared to before, but up to ~1MB based on logs I saw. We could possibly raise the cloud server limit by a bit to loosen things.

jmoenig:
Thanks, @cycomachead ! At first glance this looks alright to me. Did you test it / try it out yet? @bromagosa are you also okay with this? I'm willing to pull this right now. Will it break anything? The reason I'm asking is that I want to use the dev version next Tuesday for a 3-hour workshop at a large conference in Karlsruhe (https://www.learntec.de) and wouldn't want things to break...

cycomachead:
I've been testing it locally. Cmd-S and using the project dialog. It saved to the real cloud and to my local cloud. 

I did just push a small fix on my math because uh... 2e20 is not 2**20! 😳 time for bed!

jmoenig:
Haha! thanks. So, I guess it's okay to pull it, right?

cycomachead:
Yep, thanks!

PR #2318, opened 2019-01-30
Setup client-side error tracking

cycomachead:
This uses Rollbar to track client-side errors.
After a small tweak from the back-end, the errors will also be associated with a
user, but even without that we will have visibility into bugs.

The main thing right to is to investigate any reports of project saving issues.

I put this in a `lib/` folder, and then moved fileSaver and sha512 there since all of this code is separate from Snap!.


cycomachead:
@jmoenig @brianharvey I invited you both to Rollbar if you want to see the errors that might show up. 

(If you accept the invite you might want to turn off email notifications. I have no idea what the volume will be.)

PR #1214, opened 2016-05-10
IME composition support

ubertao:
This PR brings [IME](https://en.wikipedia.org/wiki/Input_method) support to Snap and fix various issues including #4 and #1047.

There are already other PRs trying to solve same issue:
- PR #1165 - The approach is similar but closed with some remaining issues.
- PR #1103 - Might still be a good choice as the changes are smaller. But I haven't figured out whether it can support "phrase composition" well.

The design of this PR is outlined below:

**Text input/edit involves 3 components:**
- A `StringMorph` or `TextMorph`
- The global `CursorMorph`
- The global `<textarea>` element

**What states do they maintain:**
- `StringMorph/TextMorph`
  - _text content_ being edited, updated upon <code>input</code> event fired by `<textarea>`
  - _text selection_ (via `startMark/endMark`), updated by mouse or keyboard cursor movement or text selection event
- `CursorMorph`
  - _cursor position_, including its `(x, y)` and `slot#`
  - it also manages `<textarea>`'s lifecycle
- `<textarea>`
  - _text content_ (`value` property) being edited, tells `StringMorph/TextMorph` to update its _text content_
  - _text selection_ (`selectionStart/selectionEnd` properties), updated by `CursorMorph` when cursor position is changed.
  - _placement_, to keep IME's candidate window close to cursor, updated when cursor position is changed.

**What mouse events do they handle:**
- `StringMorph/TextMorph`: Mouse click/drag events: update _text selection_ and tells `CursorMorph` to update its _cursor position_.
- `CursorMorph`: None.
- `<textarea>`: None.

**What keyboard events do they handle:**
- `StringMorph/TextMorph`: None.
- `CursorMorph`
  - Cursor movement keys (arrows, `Home`, `End`, etc.), tells `textarea` to update its text selection.
  - `Tab`, navigates to prev/next editable text field
  - `Enter`, ends the edit of a <code>StringMorph</code>
  - `ESC`, aborts the edit.
- `<textarea>`: All but cursor movement keys.

PR #2321, opened 2019-02-06
Use browser's textarea to support input from IME

swiperthefox:
#### Summary

Included in this PR, is another attempt of the "ghost textarea" method of IME support, as discussed in:

* [#1103](https://github.com/jmoenig/Snap/pull/1103) Enable IME Composing for some diacritics on Mac OS #4 X & ideograms #1047 & Android virtual keyboard
* [#1214](https://github.com/jmoenig/Snap/pull/1214) IME composition support

I tried the patches submitted in #1214 with the newest source tree, it kind of works. But since a lot of code changed in morphic.js, the merge is not very clean and includes a lot guess work. This is a new attempt, borrowed some code of the initialization function of #1214, and then toke a different road from #1214 .

The main difference is that in this PR, the `textarea` take care of handling all keyboard events, so the `CursorMorph` can do much less work, mostly acts as a visual representation of the cursor. As a result, the information flow between the `textarea` and `CursorMorph` is in single direction and minimized.

#### Motivation
On the one hand, I think some features like local variables, ability of creating reporter blocks, first class lists, together with the block based programming, make a great platform to teach kid  modular programming.

On the other hand, being able to input, display their native languages helps a lot in engaging the students in learning programming. The browser environment that Snap! runs in already have a good support for this, so I really hope Snap! can support it too.

In short: I like it, I want it to be better.
####  The design
 - Upon an edit session starts, a hidden `textarea` is created, initialized with the contents of the `Morph` to be edited.
 - The `textarea` will have keyboard focus all the time, handles all keyboard events: navigation, selection and insertion/deletion.
 - After each `keyup` event, the result content and status of selection is updated for the target `Morph`, and the caret position is updated for the `CursorMorph`.
 - The target `Morph` handles all the mouse events, which will only change the cursor position and status of selection.
 - After each mouse events, the status of selection, the position of cursor are reported to the `textarea`.

#### Improvements made by these changes:
 - The main goal is archived:
    Now user can input texts in languages that needs an input method.
 - As side effects, two more bugs are fixed:
   * Shift+click always select all text, does not select text as expected.
   * Numeric input slots accept invalid inputs like "10-2" (but treat it as 10, which is the returned value by parseFloat, I guess), "10.0.2", so one can put "10-2" in sqrt block, and got '3.16...'(sqrt(10)) back
 
#### Behavior change that might affect other part of system:
 *  Method `WorldMorph.edit`: I added a guard at the start of the function, so that  a new edit session is created only if the target morph is different from the one that is currently being edited. This is related to the above mentioned  "shift+click" bug, which is caused by always creating new edit session, even for the target Morph that is being edited.

     As far as I can tell, this changes does not affect other parts of the program.
 
####  Features that are not ported
 *    In the handling of ctrl(cmd)-keys, some special combinations of keys are  supported, for example `ctrl + F12` will insert '{'. Is there any device that needs these combinations? I've never heard of shortcut like this. If needed, I think they can be ported by synthetic events or directly the value of `textarea`.

### DIFF
--- a/snap.html
+++ b/snap.html
@@ -5,8 +5,10 @@
         <title>Snap! Build Your Own Blocks 5 - Beta -</title>
         <link rel="shortcut icon" href="src/favicon.ico">
         <script type="text/javascript" src="src/morphic.js?version=2019-01-10"></script>
+        <script type="text/javascript" src="src/imeChangesToMorphic.js"></script>
         <script type="text/javascript" src="src/widgets.js?version=2018-10-02"></script>
-        <script type="text/javascript" src="src/blocks.js?version=2019-02-06"></script>
+        <script type="text/javascript" src="src/blocks.js?version=2019-01-28"></script>
+        <script type="text/javascript" src="src/imeChangesToBlocks.js"></script>
         <script type="text/javascript" src="src/threads.js?version=2019-01-28"></script>
         <script type="text/javascript" src="src/objects.js?version=2019-01-28"></script>
         <script type="text/javascript" src="src/gui.js?version=2019-01-28"></script>
--- a/src/imeChangesToBlocks.js
+++ b/src/imeChangesToBlocks.js
@@ -1,5 +1,5 @@
 /**
- * Changes made in this file related to a bug that "shift-click" does not
+ * Changes made in this file related to a bug that "shift+click" does not
  * select text as expected.
  *
  * This reason for the bug is that in the previous version of the following
--- a/src/imeChangesToBlocks.js
+++ b/src/imeChangesToBlocks.js
@@ -1,0 +1,34 @@
+/**
+ * Changes made in this file related to a bug that "shift-click" does not
+ * select text as expected.
+ *
+ * This reason for the bug is that in the previous version of the following
+ * function, stopEditing() is always called after a mousedown event, then a
+ * new editing session started, and lost the selection status. The change I
+ * made is only to start a new editor if the target is not being edit now.
+ *
+ * I think the stopEditing() call should be made in WorldMorph.edit, if
+ * needed, since WorldMorph holds information to make the decision.
+ * WorldMorph.edit should check that:
+ * 1. If the target morph is being editing now, then do nothing.
+ * 2. If it is editing other morph, stop it.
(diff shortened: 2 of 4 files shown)


bromagosa:
I haven't had the time to look at the code, but one thing I notice is that you're creating two new files that override morph and block methods, instead of modifying these methods directly in the files where they live.

If this was a Snap! extension that'd be a good way to do it (so as to not touch the Snap! source), but you want this to be integrated into Snap!, so I'd suggest moving everything from `imeChanges*.js` into `morphic.js` and `blocks.js`.

On the other hand, thanks for tackling this long requested issue!

swiperthefox:
Thanks for your feedback.

I started by changing src/morphic.js, and when I think I am ready to submit a PR, I checked the [CONTRIBTUTING.md](https://github.com/jmoenig/Snap/blob/master/CONTRIBUTING.md), and that doc suggests to contribute via change set file. It's something new to me, but since that document is quite new (updated 4 months ago), I think I should follow the instruction and switch to change sets.

I'll put the changes back to the original files, and update this PR.

Before re submit, I have a question about the git history. The git history of my development of this feature is not very clean: sometime I just make a commit so that I can work on it at another place, with a different computer later. If this PR is accepted and merged, will my git history just disappear or it will be merged too? Should I clean the history before submit the PR? (since it's a one man's work for now, I think it's ok to change the history before it goes public?)

Thanks for any suggestions.

bromagosa:
Uhm. I believe that file is meant for people who don't use Git. Maybe that should be specified somewhere...

Regarding git history: PRs can be squashed into a single commit when merging, but I don't think Jens cares much about the commit history :)

brianharvey:
The discussion of changesets in that file is from back when Jens was refusing to use git (not because of how screwy the software is, but because he didn't and still doesn't want the entire world adding their favorite feature to Snap<i>!</i>).  What's different now is having colleagues such as Bernat working very actively on Snap<i>!</i>, so it no longer works for Jens to do all the merging by hand.  @jmoenig should rewrite the Contributing document...

swiperthefox:
Thanks for the intro of history.

I merge the change set files into the original files, and created a new PR #2322 .

PR #2322, opened 2019-02-08
Provide IME support to text editing using hidden textarea

swiperthefox:
#### Summary

Included in this PR, is another attempt of the "ghost textarea" method of IME support, as discussed in:

* [#1103](https://github.com/jmoenig/Snap/pull/1103) Enable IME Composing for some diacritics on Mac OS #4 X & ideograms #1047 & Android virtual keyboard
* [#1214](https://github.com/jmoenig/Snap/pull/1214) IME composition support

I tried the patches submitted in #1214 with the newest source tree, it kind of works. But since a lot of code changed in morphic.js, the merge is not very clean and includes a lot guess work. This is a new attempt, borrowed some code of the initialization function of #1214, and then toke a different road from #1214 .

The main difference is that in this PR, the `textarea` take care of handling all keyboard events, so the `CursorMorph` can do much less work, mostly acts as a visual representation of the cursor. As a result, the information flow between the `textarea` and `CursorMorph` is in single direction and minimized.

#### Motivation
On the one hand, I think some features like local variables, ability of creating reporter blocks, first class lists, together with the block based programming, make a great platform to teach kid  modular programming.

On the other hand, being able to input, display their native languages helps a lot in engaging the students in learning programming. The browser environment that Snap! runs in already have a good support for this, so I really hope Snap! can support it too.

In short: I like it, I want it to be better.
####  The design
 - Upon an edit session starts, a hidden `textarea` is created, initialized with the contents of the `Morph` to be edited.
 - The `textarea` will have keyboard focus all the time, handles all keyboard events: navigation, selection and insertion/deletion.
 - After each `keyup` event, the result content and status of selection is updated for the target `Morph`, and the caret position is updated for the `CursorMorph`.
 - The target `Morph` handles all the mouse events, which will only change the cursor position and status of selection.
 - After each mouse events, the status of selection, the position of cursor are reported to the `textarea`.

#### Improvements made by these changes:
 - The main goal is archived:
    Now user can input texts in languages that needs an input method.
 - As side effects, two more bugs are fixed:
   * Shift+click always select all text, does not select text as expected.
   * Numeric input slots accept invalid inputs like "10-2" (but treat it as 10, which is the returned value by parseFloat, I guess), "10.0.2", so one can put "10-2" in sqrt block, and got '3.16...'(sqrt(10)) back
 
#### Behavior change that might affect other part of system:
 *  Method `WorldMorph.edit`: I added a guard at the start of the function, so that  a new edit session is created only if the target morph is different from the one that is currently being edited. This is related to the above mentioned  "shift+click" bug, which is caused by always creating new edit session, even for the target Morph that is being edited.

     As far as I can tell, this changes does not affect other parts of the program.
 
####  Features that are not ported
 *    In the handling of ctrl(cmd)-keys, some special combinations of keys are  supported, for example `ctrl + F12` will insert '{'. Is there any device that needs these combinations? I've never heard of shortcut like this. If needed, I think they can be ported by synthetic events or directly the value of `textarea`.

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -5418,7 +5418,7 @@ CursorMorph.prototype.initializeTextarea = function () {
         }
     });
 
-    // For other keyboard events, let the textarea element handle other key 
+    // For other keyboard events, first let the textarea element handle the
     // events, then we take its state and update the target morph and cursor
     // morph accordingly.
     this.textarea.addEventListener('keyup', function (event) {
@@ -5498,7 +5498,9 @@ CursorMorph.prototype.syncTextareaSelectionWith = function (targetMorph) {
     var start = targetMorph.startMark;
     var end = targetMorph.endMark;
 
-    if (start <= end) {
+    if (start === end) {
+        this.textarea.setSelectionRange(this.slot, this.slot, 'none');
+    } else if (start < end) {
         this.textarea.setSelectionRange(start, end, 'forward');
     } else {
         this.textarea.setSelectionRange(end, start, 'backward');
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -5357,73 +5357,153 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
         this.target.setAlignmentToLeft();
     }
     this.gotoSlot(this.slot);
-    this.initializeClipboardHandler();
+    this.initializeTextarea();
 };
 
-CursorMorph.prototype.initializeClipboardHandler = function () {
-    // Add hidden text box for copying and pasting
-    var myself = this,
-        wrrld = this.target.world();
+CursorMorph.prototype.initializeTextarea = function () {
+    var myself = this;
 
-    this.clipboardHandler = document.createElement('textarea');
-    this.clipboardHandler.style.position = 'absolute';
-    this.clipboardHandler.style.top = window.outerHeight;
-    this.clipboardHandler.style.right = '101%'; // placed just out of view
+    this.textarea = document.createElement('textarea');
+    this.textarea.style.zIndex = -1;
+    this.textarea.style.position = 'absolute';
+    this.textarea.wrap = "off";
+    this.textarea.style.overflow = "hidden";
+    this.textarea.style.fontSize = this.target.fontSize + 'px';
+    this.textarea.autofocus = true;
+    this.textarea.value = this.target.text;
+    document.body.appendChild(this.textarea);
+    this.updateTextAreaPosition();
+    this.syncTextareaSelectionWith(this.target);
 
-    document.body.appendChild(this.clipboardHandler);
 
-    this.clipboardHandler.value = this.target.selection();
-    this.clipboardHandler.focus();
-    this.clipboardHandler.select();
+    /* The following keyboard events causes special actions in Snap, so we
+    don't want the textarea to handle it:
+    - ctrl-d, ctrl-i and ctrl-p: doit, inspect it and print it
+    - tab: goto next text field
+    - esc: discard the editing
+    - enter / shift+enter: accept the editing
+    */
+    this.textarea.addEventListener('keydown', function (event) {
+        // other part of the world need to know the current key
+        myself.world().currentKey = event.keyCode;
 
-    this.clipboardHandler.addEventListener(
-        'keypress',
-        function (event) {
-            myself.processKeyPress(event);
-            this.value = myself.target.selection();
-            this.select();
-        },
-        false
-    );
+        var keyName = event.key;
+        var shift = event.shiftKey;
+        var singleLineText = myself.target instanceof StringMorph;
 
-    this.clipboardHandler.addEventListener(
-        'keydown',
-        function (event) {
-            myself.processKeyDown(event);
-            if (event.shiftKey) {
-                wrrld.currentKey = 16;
+        if (!isNil(myself.target.receiver) &&
+                    (event.ctrlKey || event.metaKey)) {
+            if (keyName === 'd') {
+                myself.target.doIt();
+            } else if (keyName === 'i') {
+                myself.target.inspectIt();
+            } else if (keyName === 'p') {
+                myself.target.showIt();
             }
-            this.value = myself.target.selection();
-            this.select();
-
-            // Make sure tab prevents default
-            if (event.key === 'U+0009' ||
-                    event.key === 'Tab') {
-                myself.processKeyPress(event);
-                event.preventDefault();
+            event.preventDefault();
+        } else if (keyName === 'Tab' || keyName === 'U+0009') {
+            if (shift) {
+                myself.target.backTab(myself.target);
+            } else {
+                myself.target.tab(myself.target);
             }
-        },
-        false
-    );
+            event.preventDefault();
+            myself.target.escalateEvent('reactToEdit', myself.target);
+        } else if (keyName === 'Escape') {
+            myself.cancel();
+        } else if (keyName === "Enter" && (singleLineText || shift)) {
+            myself.accept();
+        } else {
+            myself.target.escalateEvent('reactToKeystroke', event);
+        }
+    });
 
-    this.clipboardHandler.addEventListener(
-        'keyup',
-        function (event) {
-            wrrld.currentKey = null;
-        },
-        false
-    );
+    // For other keyboard events, let the textarea element handle other key 
+    // events, then we take its state and update the target morph and cursor
+    // morph accordingly.
+    this.textarea.addEventListener('keyup', function (event) {
+        myself.world().currentKey = null;
 
-    this.clipboardHandler.addEventListener(
-        'input',
-        function (event) {
-            if (this.value === '') {
-                myself.gotoSlot(myself.target.selectionStartSlot());
-                myself.target.deleteSelection();
+        var target = myself.target;
+        var textarea = myself.textarea;
+        var filteredContent;
+        
+        // filter invalid chars for numeric fields
+        function filterText (content) {
+            var points = 0;
+            var result = '';
+            for (var i=0; i < content.length; i++) {
+                var ch = content.charAt(i);
+                var valid = (
+                    ('0' <= ch && ch <= '9') || // digits
+                    (i === 0 && ch === '-')  || // leading '-'
+                    (ch === '.' && points === 0) // at most '.'
+                );
+                if (valid) {
+                    result += ch;
+                    if (ch === '.') points++;
+                }
             }
-        },
-        false
-    );
+            return result;
+        }
+        
+        if (target.isNumeric) {
+            filteredContent = filterText(textarea.value);
+        } else {
+            filteredContent = textarea.value;
+        }
+        
+        if (filteredContent.length < textarea.value.length) {
+            textarea.value = filteredContent;
+            var caret = Math.min(textarea.selectionStart, filteredContent.length);
+            textarea.selectionEnd = caret;
+            textarea.selectionStart = caret;
+        }
+        // target morph: copy the content and selection status to the target.        
+        target.text = filteredContent;
+        if (textarea.selectionStart === textarea.selectionEnd) {
+            target.startMark = null;
+            target.endMark = null;
+        } else {
+            if (textarea.selectionDirection === 'backward') {
+                target.startMark = textarea.selectionEnd;
+                target.endMark = textarea.selectionStart;
+            } else {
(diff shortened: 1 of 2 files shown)


bromagosa:
I can't exactly pin down what's going on, but in text areas like project notes, selecting all text, then moving to other inputs, then coming back and re-selecting all text causes partial disappearance of the contents.

I can't seem to find a minimum reproducible case, though :(

swiperthefox:
Thanks for the feedback. I can't reproduce the problem, but I do find two issues related to the projectNote editor:

1. When opening project note editor, the textarea does not get the focus.

I've figured out the problem, it's because I assumed every editing sesssion will start with "selectAll", so I put the code of setting focus in that method.

2. The "shift + enter" way of accepting the editing result will lost all the editing:

open project note editor, do some editing, then press "shift enter", reopen the editor, all change are lost. 

but clicking "ok" does not have this problem and will save all the changes. I think this might relate to what you've found, since switching to other fields is an implicit way of stopping the edit.

I'll investigate more.

swiperthefox:
OK, the problem is just a few lines away, here is the code for handling "ok".
```
IDE_Morph.prototype.editProjectNotes = function () {
    ...
    var text = new TextMorph(this.projectNotes || ''),
    ...
    dialog.ok = function () {
        myself.projectNotes = text.text;
        ok.call(this);
    };
```
So `text` (the TextMorph being edited) is only a temporary holder of the note, click "ok" will save the content to the `projectNotes` field of the IDE_Morph, but "shift-enter" only tells the `text` morph to keep the result, on the next call of `editProjectNotes`, the content of `text` will be overwritten.

swiperthefox:
@bromagosa, I submitted another PR #2323, which fixed the bug related to "shift enter" key press. Please check if the problem you mentioned is fixed.

swiperthefox:
At the start of editing the project note, the CursorMorph's position is at the end of text, but the cursor position of the underlying `textarea` element is at the start of text.

Commit e972ae0 fixed this.

src/morphic.js
```diff
@@ -5357,73 +5357,154 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
...
-    this.clipboardHandler = document.createElement('textarea');
-    this.clipboardHandler.style.position = 'absolute';
-    this.clipboardHandler.style.top = window.outerHeight;
-    this.clipboardHandler.style.right = '101%'; // placed just out of view
+    this.textarea = document.createElement('textarea');
+    this.textarea.style.zIndex = -1;
+    this.textarea.style.position = 'absolute';
+    this.textarea.wrap = "off";
+    this.textarea.style.overflow = "hidden";
+    this.textarea.style.fontSize = this.target.fontSize + 'px';
```
cycomachead: Why does the font size need to be set? If the element is hidden, shouldn't this be unnecessary?

src/morphic.js
```diff
@@ -5357,73 +5357,154 @@ CursorMorph.prototype.init = function (aStringOrTextMorph) {
...
-        wrrld = this.target.world();
+CursorMorph.prototype.initializeTextarea = function () {
+    var myself = this;
 
-    this.clipboardHandler = document.createElement('textarea');
-    this.clipboardHandler.style.position = 'absolute';
-    this.clipboardHandler.style.top = window.outerHeight;
-    this.clipboardHandler.style.right = '101%'; // placed just out of view
+    this.textarea = document.createElement('textarea');
+    this.textarea.style.zIndex = -1;
```
cycomachead: Minor thing:
For the style elements that are static, I think it's clearer to write

```suggestion
    this.textarea.setAttribute('style', 'z-index: -1; position: absolute; overflow: hidden')'; 
```

which keeps it all together, then `wrap` and `autofocus` can be next to each other (since they are both attributes specific to the element.)

src/morphic.js
swiperthefox: Thanks for your feed back.

I included this line of code to align the floating IME window to the position of caret: 

1. Users expect that the IME window appears near the caret (of the text morphic), but the OS/browser will put the IME window near to the caret of underlying textarea.

2. By setting the font size of the text morphic and the textarea to the same, we can minimize the distance of the two carets.

I recall that I added this when I noticed a huge offset of the IME window. I tried to comment off this line to take a screenshot that showcasing the problem, but in my current setting, there is no noticeable difference with/without this code: in both cases, the position of IME window are not perfect but acceptable.

src/morphic.js
swiperthefox: Thanks for your suggestion, I'll change it in next PR.

I've never thought this way: "this.textarea.style.zIndex=1" is changing the "style" attribute of the element.

ToonTalk:
I'm involved in a project that would greatly benefit from Chinese input. For testing we are using copy and paste instead (which is far from ideal).  My questions are

1. any estimate when this will all be working?
2. if it will be more than a couple months can we, as a fallback, use this branch instead of the standard release?

swiperthefox:
> I'm involved in a project that would greatly benefit from Chinese input. For testing we are using copy and paste instead (which is far from ideal). My questions are
> 
>     1. any estimate when this will all be working?
> 
>     2. if it will be more than a couple months can we, as a fallback, use this branch instead of the standard release?

I hope this feature will be added to snap! soon. Before that, the way I'd like to "patch" the current snap! site is through browser addons like greasymonkey. For example, [Here](https://gist.github.com/swiperthefox/48996720c0e82b3d02560188a4b5006f) is a violentmonkey script.

ToonTalk:
Great，thanks for the script. In Chrome it works well with one minor problem. The display is updated only after I introduce a subsequent character. e.g.

image

If I type anything then the "za" in this example becomes a Chinese character. (In contrast when I select a Chinese character via IME here in the browser I instantly see the character.)

I also tried this in FireFox I get the error 

Script error in [Greasemonkey script Violentmonkey Scripts/ADD IME TO SNAP]:
ReferenceError: InputSlotMorph is not defined

swiperthefox:
> If I type anything then the "za" in this example becomes a Chinese character. (In contrast when I select a Chinese character via IME here in the browser I instantly see the character.)

The synchronization between the morphic and underlying input filed happens after the input field receives a "keyup" event. It looks like the browser is not generating this "keyup" event when you select the Chinese character. I don't know why this happens, it might depends on the browser, OS and the IME method. I tested it with Linux+fcitx(IME)+Chrome/Firefox, OS X+System IME+Chrome, and can't reproduce the problem.
 
An (ugly) workaround is to show the input filed, instead of hiding it. Make the following changes to the "initializeTextaea" method:

`this.textarea.style.zIndex = 1000; // show this textarea on top of the canvas
`
> I also tried this in FireFox I get the error
> 
> Script error in [Greasemonkey script Violentmonkey Scripts/ADD IME TO SNAP]:
> ReferenceError: InputSlotMorph is not defined

I am not 100% sure, but maybe this is because of Greasemonkey's sandboxing of the script, the script don't have access to the global namespace of `window`.

ToonTalk:
The synchronization problem occurs in Windows 10 both in Chrome and FireFox.

Thanks for your help with this. While not ideal I think we can live with this for now.

I can think of a few ugly workarounds such as checking a few times a second while the text morph is the focus. Maybe upon entry use setTimeout to repeatedly check until exiting the text morph. 

Oh well.

DarDoro:
Most problems can be solved by using *"input"* event (as a  *"onchange"* substitute) instead of *"keyup"*. 
Candidate window can be handled by mouse with instant feedback. Android virtual keyboard should work even for ASCII input (maybe voice input too, barcode scanning etc.).
*"keyup"* can be reduced to handle cursor movement, selection and non visual keys.

ToonTalk:
Perhaps there should be a listener for compositionend events.

> The compositionend event is fired when a text composition system such as an input method editor completes or cancels the current composition session.

From https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event

DarDoro:
@Toontalk, indeed . This way it is working in #1103. But "composition*" events designed to support IME have some flaws in browsers. For example, somtimes browsers "forgot" to emit any useful event for numbers and punctuation. No "composition*" nor "key*" event only textarea update.

swiperthefox:
> @ToonTalk, indeed . This way it is working in #1103. But "composition*" events designed to support IME have some flaws in browsers. For example, somtimes browsers "forgot" to emit any useful event for numbers and punctuation. No "composition*" nor "key*" event only textarea update.

@ToonTalk, could you please test Dardoro's idea, thanks. You only need to replace "keyup" to "input" in the following line:

     [in src/morphic.js, near line 5424]
     this.textarea.addEventListener('keyup', function (event) {
         ...

Since I can't reproduce the problem you mentioned, I can't tell if this will fixed it. I've tried this change, and it works on my environment.

DarDoro:
Handling for "keyup" can be duplicated or better reduced only to handle cursor and selection.

 ```
this.textarea.addEventListener('keyup', function (event) {
        myself.world().currentKey = null;

        var target = myself.target;
        var textarea = myself.textarea;
        // target morph: copy the content and selection status to the target.        
        if (textarea.selectionStart === textarea.selectionEnd) {
            target.startMark = null;
            target.endMark = null;
        } else {
            if (textarea.selectionDirection === 'backward') {
                target.startMark = textarea.selectionEnd;
                target.endMark = textarea.selectionStart;
            } else {
                target.startMark = textarea.selectionStart;
                target.endMark = textarea.selectionEnd;
            }
        }
        target.changed();
        target.drawNew();
        target.changed();

        // cursor morph: copy the caret position to cursor morph.
        myself.gotoSlot(textarea.selectionStart);

        myself.updateTextAreaPosition();
    });
```

swiperthefox:
Thanks for the suggestion. That's what I end up to.
> Handling for "keyup" can be duplicated or better reduced only to handle cursor and selection.

ToonTalk:
It's working fine now. Thanks!

I made [these changes](https://github.com/jmoenig/Snap/pull/2322/commits/d8e2bc31290849a4753e779cb91d49df85d38032) to my copy of violentmonkey_script.user.js

jmoenig:
What's the state of this PR?
Should we incorporate this into Snap next? Are there other ways or is this the best one? What are your takes on this, @DarDoro (I think you also suggested this or a similar solution), @ToonTalk (what are your experiences with this so far?), @bromagosa and @cycomachead

jmoenig:
I'm also interested in the relationship of this PR to @DarDoro's #1103 PR. Are these similar strategies, has #1103 been superseded by this or vice-versa. Tnx

jmoenig:
Thank you @ToonTalk ! So, it looks like this should probably go into Snap's main branch real soon.

DarDoro:
> I'm also interested in the relationship of this PR to @DarDoro's #1103 PR. Are these similar strategies, has #1103 been superseded by this or vice-versa. Tnx

Different approach virtually no similarity.

jmoenig:
Thanks, everybody for your feedback. Playing with this it looks and feels great. I'll be looking into this more closely over the next time, but this is becoming a priority for v5.1. I'm already marking this "blessed by jens", but I might end up not directly pulling this PR but integrating it in other ways. Thank you, @swiperthefox , for this excellent piece of pioneering!

swiperthefox:
> Thanks, everybody for your feedback. Playing with this it looks and feels great. I'll be looking into this more closely over the next time, but this is becoming a priority for v5.1. I'm already marking this "blessed by jens", but I might end up not directly pulling this PR but integrating it in other ways. Thank you, @swiperthefox , for this excellent piece of pioneering!

@jmoenig No problem. Thanks for creating Snap!. I am glad to be able to contribute something to this great project.

jmoenig:
thank you, @swiperthefox ! Hey, if you tell me your actual name I'd like to credit you in the release notes and in the contributors tab...

swiperthefox:
Hi, Jens,

Thanks for including the IME support, now one more reason to get my kids
into Snap!. I am glad I can help with this.

My name is Zhenlei Jia. It's my honor to be included. Thanks.

Best,
Zhenlei

Jens Mönig <notifications@github.com> 于2019年7月22日周一 下午12:40写道：

jmoenig:
Hi Zhenlei @swiperthefox !
yay!! thank you, I've just added your name to both the credits in Morphic and Snap, and we've released the patch with your contribution in v5.0.6 https://github.com/jmoenig/Snap/releases/tag/v5.0.6
I'm excited about this, as it makes Snap! more accessible to children in many, many countries, plus it finally makes it work on Android tablets.
Thanks again!
-Jens

PR #2329, opened 2019-02-15
fix vector editor colors

brianharvey:
Reverse edge and fill, rename to edge and fill, fix
left-button/right-button selection, all good things, no bad things.

### DIFF
--- a/src/sketch.js
+++ b/src/sketch.js
@@ -1261,7 +1261,7 @@ VectorPaintEditorMorph.prototype.populatePropertiesMenu = function () {
         myself = this,
         pc = this.propertiesControls,
         alpen = new AlignmentMorph("row", this.padding),
-        alignColor = new AlignmentMorph("row", this.padding);
+        alignColor = new AlignmentMorph("row", this.padding),
         alignNames = new AlignmentMorph("row", this.padding);
 
     pc.primaryColorViewer = new Morph();
--- a/src/sketch.js
+++ b/src/sketch.js
@@ -1020,7 +1020,7 @@ VectorPaintEditorMorph.prototype.convertToBitmap = function () {
 
 VectorPaintEditorMorph.prototype.buildScaleBox = function () {
     var myself = this;
-    ['Top', 'Bottom', 'Up', 'Down'].forEach(function (label) {
+    ['Frontmost', 'Backmost', 'Front', 'Back'].forEach(function (label) {
         myself.scaleBox.add(
             myself.pushButton(
                 label,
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks! I remembered there was something different for lines, but apparently it's not. Great! This might take a while for me to pull and integrate. I also want to make it easier to translate the color labels, and update some translations.

src/sketch.js
```diff
@@ -1339,12 +1339,12 @@ VectorPaintEditorMorph.prototype.populatePropertiesMenu = function () {
...
             );
 
-    alignColor.add(pc.primaryColorViewer);
     alignColor.add(pc.secondaryColorViewer);
+    alignColor.add(pc.primaryColorViewer);
     alignColor.fixLayout();
 
     c.add(pc.colorpicker);
-    c.add(new TextMorph(localize('Primary color      Secondary color')));
+    c.add(new TextMorph(localize('    Edge color             Fill color')));
```
jmoenig: this is the part I would love to change. Using magic spaces to align what should be two individual labels doesn't work well for translations. But that's not your change, so don't worry about it. I'll take care of turning this into separate strings/morphs when I integrate your changes. Thanks!

brianharvey:
There _was_ something different for lines, and for brush painting, but I fixed it!  :-)

brianharvey:
This was fun, learning about Morphic...

src/sketch.js
```diff
@@ -1262,6 +1262,7 @@ VectorPaintEditorMorph.prototype.populatePropertiesMenu = function () {
         pc = this.propertiesControls,
         alpen = new AlignmentMorph("row", this.padding),
         alignColor = new AlignmentMorph("row", this.padding);
+        alignNames = new AlignmentMorph("row", this.padding);
```
jmoenig: excellent! That's what I would've done :-)
Thanks!

src/sketch.js
```diff
@@ -1343,9 +1344,19 @@ VectorPaintEditorMorph.prototype.populatePropertiesMenu = function () {
     alignColor.add(pc.primaryColorViewer);
     alignColor.fixLayout();
 
+    alignNames.add(new TextMorph(localize('Edge color'),
```
jmoenig: nice! Now this is done the right way. Awesome!

jmoenig:
Thank you, now it's perfect!

src/sketch.js
bromagosa: The semicolon at the end of the `alignColor` line should be a comma, I believe :)

jmoenig: oh yeah. Please always test and JSHint/Lint your code.

brianharvey: oops.  I tested it a lot but I never think about linting it, because (unlike the original C lint) 3/4 of what it says is arbitrary standards of taste.  :-(  I'll try to remember.

jmoenig: That's certainly true of Crockford's JSLint, which has become totally unusable because it forces Doug's personal taste on everybody. You'll find JSHint (https://jshint.com) much more tolerant and usable.

PR #2335, opened 2019-02-20
Fixes do play until JS error

bromagosa:
Fixes this JS error:

captura de pantalla de 2019-02-20 09-18-29

I had noticed it before, but this time it popped up while I was recording a video and I thought it'd be easier to fix it than to explain what the error meant :)


bromagosa:
Oh wow. What happened here!

PR #2337, opened 2019-02-21
fix pipette bug

brianharvey:
Shift-click on pipette button makes it work on fill color instead of
edge.  (The shiftness is the click on the button itself, unlike other
buttons, but pipette is different anyway because when you let go it
switches back to the old button selection.)

### DIFF
--- a/src/paint.js
+++ b/src/paint.js
@@ -495,7 +495,7 @@ PaintEditorMorph.prototype.getUserColor = function () {
     var myself = this,
         world = this.world(),
         hand = world.hand,
-		shifted = this.paper.isShiftPressed();
+        shifted = this.paper.isShiftPressed();
         posInDocument = getDocumentPositionOf(world.worldCanvas),
         mouseMoveBak = hand.processMouseMove,
         mouseDownBak = hand.processMouseDown,
@@ -514,7 +514,7 @@ PaintEditorMorph.prototype.getUserColor = function () {
             return;
         }
         color.a = 255;
-        myself.propertiesControls.colorpicker.action(color,shifted);
+        myself.propertiesControls.colorpicker.action(color, shifted);
     };
 
     hand.processMouseDown = nop;
--- a/src/paint.js
+++ b/src/paint.js
@@ -495,6 +495,7 @@ PaintEditorMorph.prototype.getUserColor = function () {
     var myself = this,
         world = this.world(),
         hand = world.hand,
+		shifted = this.paper.isShiftPressed();
         posInDocument = getDocumentPositionOf(world.worldCanvas),
         mouseMoveBak = hand.processMouseMove,
         mouseDownBak = hand.processMouseDown,
@@ -513,7 +514,7 @@ PaintEditorMorph.prototype.getUserColor = function () {
             return;
         }
         color.a = 255;
-        myself.propertiesControls.colorpicker.action(color);
+        myself.propertiesControls.colorpicker.action(color,shifted);
     };
 
     hand.processMouseDown = nop;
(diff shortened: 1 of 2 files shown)


jmoenig:
thanks, Brian!

jmoenig:
Aw man, that fix didn't work and it took me hours to track down how to make it work, because - let's face it - the whole architecture of paint.js and - as a derivative - sketch.js just sucks incredibly.

Can we for the moment please stop submitting pull-requestst to these two modules of code? I'm not kidding. Thanks!

jmoenig:
basically your change to reverse the meaning of "secondary" and "primary" for certain figures broke a lot of stuff. That's exactly what I warned you about and why I didn't want you to touch the damned code. I'd been there and tried it and discovered that it's a rat hole. But you claimed it was all "easy" and "all good". Turns out, it wasn't. So, please, let's not do any quick'n'easy stuff anymore before the release.

jmoenig:
You can see that it used to work when you try it in the current non-beta version of Snap!

brianharvey:
Jens, I'm really sorry this is causing you trouble.  But I tested all the 
constraint stuff, and it works for me!  So I don't understand why it 
doesn't work for you.

(Just by the way, I didn't say "easy" about the pipette bug.  That took me 
a long time to find and fix, because the actual work is done in the bitmap 
editor code.  But after the long time, it was a tiny change to fix it.)

jmoenig:
No need to be sorry, Brian. That code is - as I've mentioned before - terrible in many ways. The architecture is byzantine, and it's extremely hard to find out which code is responsible for what. As it turns out, you *thought* that the pipette bug was due to things governed by the bitmap editor, but upon closer inspection the part that was broken was, in fact, in the vector editor part. So, nevermind the effort. I'm glad we've managed to at least improve the vector editor a bit for the upcoming release. But let's stop here and not introduce more bugs while trying to improve things.

bromagosa:
Yep. I got bitten hard by the editor too. I was like "hey, let's start this over and redo the vector editor entirely. It can't take longer than a week or two", and there I was two months later ready to jump off a window.

The absolute worst part is, IMO, the handling of local coordinates.

brianharvey:
The next time we have five minutes with nothing to do, one of you is going to have to show me what my code breaks.  I swear I tested everything!  And, you know how it goes, I think (thought?) I finally _understand_ the code.

jmoenig:
Well, you noticed later that the pipette was no longer reacting to holding the shift-key down. That was what you're code broke. Then you tried to fix that but didn't find the right function where it was happening, so you made it so that the user had to press shift while mouse-clicking on the pipette button, and then it was *permanently* sampling the fill color even once you let go of the shift key. The thing you oversaw was that when you changed the meaning of "primary" and "secondary" *that* change affected how the pipette works.

jmoenig:
But - again, this ain't your fault here. The code *is* evil and awful, and the architecture is what happens when you let teenager write production code :-)

brianharvey:
Oh!  But wait!  That change was deliberate.  The thing is, the other buttons just set a tool mode; they don't actually do anything when clicked.  So it makes sense to sample the shiftedness when you actually use the tool.  But the pipette tool immediately changes a color (by default, the edge color) when you click it, and keeps changing that color as you move the mouse around.  So, the color it changes as you're moving the mouse has to be choosable when you click the pipette tool.  I think it's really confusing if you change the edge color while moving the mouse around and then at the last second you shift-click and the edge color goes back to what it was and the fill color changes instead.

By the way, something I want to change later is that hitting ESC while in pipette mode restores the previous color (the _status quo ante_ as we lawyers say...).

Whew, I thought I was crazy, and this is just a disagreement about the design.

jmoenig:
No, that can't have been deliberate, because you edited the wrong function and broke another one. This was about renaming labels, not about changing how the UI works. C'mon!

brianharvey:
Okay, I'm back to "you have to show me."  But it doesn't have to be soon.

jmoenig:
will do :-)

PR #2338, opened 2019-02-22
Allow loading downloaded scripts that contain custom blocks

bromagosa:
When working on Paul's project I constantly need to copy snippets from a project to another, and importing a script that includes custom blocks was giving out an error.

This doesn't pull the _definition_ of the custom block, but it lets you import it if the definition is there.

Not sure if you'll like the way I detect whether we're importing a single script. I wanted to touch as little code from store.js as possible, so this seemed like an okay way to me.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -1085,6 +1085,14 @@ SnapSerializer.prototype.loadScript = function (model, object) {
     // private
     var topBlock, block, nextBlock,
         myself = this;
+
+    // Check whether we're importing a single script, not a script as part of a
+    // whole project
+    if (!this.project.stage) {
+        this.project.stage = object.parentThatIsA(StageMorph);
+        this.project.targetStage = this.project.stage;
+    }
+
     model.children.forEach(function (child) {
         nextBlock = myself.loadBlock(child, false, object);
         if (!nextBlock) {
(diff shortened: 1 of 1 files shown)


brianharvey:
I vote for not _ever_ putting one of those hideous red "obsolete" blocks in deserialized code!  (a) They're very intimidating; (b) the name "obsolete" is no longer applicable; (c) the check for the existence of a block matching this invocation should happen at runtime (and, sure, be cached after the first success so it doesn't slow down the execution).

bromagosa:
That's a different kind of issue, though, Brian. Let's not start hijacking conversations :pray:

:)

brianharvey:
But it's not different!  I'm (I think) proposing a simplification of the piece of code including your patch, namely, removing it entirely.  :P

bromagosa:
This gave out an error saying "receiver is undefined":

imatge

So it's totally different, because there were no obsolete blocks to start with.

brianharvey:
Oh, I see.  But the loadBlock method is the one that says Obsolete; that's why I thought that was what you meant.

bromagosa:
This refers to `loadScript`, not `loadBlock`. The only change in `loadBlock` is an accidental enter (that other methods do include, so I haven't removed) between the var definition and the following line of code.

(And as I feared we've totally hijacked the thread)

brianharvey:
Oh.  Sorry!

jmoenig:
looks good to me. Thanks, Bernat!

PR #2339, opened 2019-02-23
current state of manuals

brianharvey:
5.0 is not ready!!  Especially the two undocumented libraries on page
87.  But the whole thing needs to be indexed again.


jmoenig:
Okay, thanks! So, I shouldn't be pulling this until it's done, right?

brianharvey:
Right!  This is my first priority; it should be done soon.

jmoenig:
Thanks for the clarification. Yes, the sooner the better, but still: No rush! The manual is already in pretty good shape and you've got plenty of stuff to handle with the lawyers this week.

jmoenig:
bumping this thread. @brianharvey should I now pull this or wait some more? Thanks!

jmoenig:
Thank you, @brianharvey ! Now reading...

jmoenig:
I sent you - only - two very minor quibbles. This is looking GREAT!

PR #2341, opened 2019-02-24
Adding 'localizes' to allow some string translations for Snap5

jguille2:
Hi!
Adding some "localize"s to allow their string translations.

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -5412,20 +5412,20 @@ IDE_Morph.prototype.initializeCloud = function () {
                     myself.source = 'cloud';
                     if (!isNil(response.days_left)) {
                         new DialogBoxMorph().inform(
-                            'Unverified account: ' +
+                            localize('Unverified account: ') +
                             response.days_left +
-                            ' days left',
-                            'You are now logged in, and your account\n' +
-                            'is enabled for three days.\n' +
-                            'Please use the verification link that\n' +
-                            'was sent to your email address when you\n' +
-                            'signed up.\n\n' +
-                            'If you cannot find that email, please\n' +
-                            'check your spam folder. If you still\n' +
-                            'cannot find it, please use the "Resend\n' +
-                            'Verification Email..." option in the cloud\n' +
-                            'menu.\n\n' +
-                            'You have ' + response.days_left + ' days left.',
+                            localize(' days left'),
+                            localize('You are now logged in, and your account\n') +
+                            localize('is enabled for three days.\n') +
+                            localize('Please use the verification link that\n') +
+                            localize('was sent to your email address when you\n') +
+                            localize('signed up.\n\n') +
+                            localize('If you cannot find that email, please\n') +
+                            localize('check your spam folder. If you still\n') +
+                            localize('cannot find it, please use the "Resend\n') +
+                            localize('Verification Email..." option in the cloud\n') +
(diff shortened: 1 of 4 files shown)


src/gui.js
```diff
@@ -5412,20 +5412,20 @@ IDE_Morph.prototype.initializeCloud = function () {
...
-                            'Please use the verification link that\n' +
-                            'was sent to your email address when you\n' +
-                            'signed up.\n\n' +
-                            'If you cannot find that email, please\n' +
-                            'check your spam folder. If you still\n' +
-                            'cannot find it, please use the "Resend\n' +
-                            'Verification Email..." option in the cloud\n' +
-                            'menu.\n\n' +
-                            'You have ' + response.days_left + ' days left.',
+                            localize(' days left'),
```
jmoenig: hmm... for long strings such as the ones in this PR it might make sense to leave it as a single large one, instead of breaking it up into several substrings. I can imagine some languages have a different sentence structure...

src/paint.js
```diff
@@ -237,8 +237,8 @@ PaintEditorMorph.prototype.buildEdits = function () {
         function () {
             if (myself.paper.undoBuffer.length > 0) {
                 myself.ide.confirm(
-                    'This will erase your current drawing.\n' +
-                    'Are you sure you want to continue?',
+                    localize('This will erase your current drawing.\n') +
```
jmoenig: hmmm.... I don't think these strings need to be explicitly localized at all here. When I look at the code for the ```IDE_Morph >> confirm```  method it already has a call to ```localize``` in it.

jmoenig (commented):
the English translations dictionary has a section for handling long strings. The idea is that we don't directly put long strings into the source code but short ones that we in turn look up in the respective dictionaries. I seem to have not abided by this mechanism myself.

src/gui.js
jguille2: @jmoenig
I think long strings can be broken into smaller ones. It is good to avoid long strings, both in code and in translations (translations are also code).

But yes, for me the mistake here is to break the message in lines. We can break long messages in sentences. That will be good to have shorter lines and shorter strings... and also good to take the advantage of the work done: some sentences will be repeated more than once, and they will need only one translation (sentences like "Are you sure?").

I added the "localizes" and I didn't want to change the origin code. But if you want, I will update the PR cutting the messages by sentences.

src/paint.js
jguille2: @jmoenig 
Yes, but the issue here is the same as the previous comment. I know ide.confirm has already a call to localize, but here we have again message made up with different strings. If we want that the translation dictionary was the same that the code dictionary (the same strings), I have to call a "localize" for each shorter string.

jguille2:
Hi @jmoenig,

Is this right for you?

Now, those long strings are still splitted (like the original code was), by splitted in sentences (not by lines).

Other possibility is to split a bit more, using phrases (maybe one sentence has more than one phrase... I'm not sure if these words are in English the best... but I think you know my idea). Then, the strings will be shorter, and I think the translation will not be a problem... but we have those "\n" inside that are strange (Yes translators can move them between strings... but then, the idea of using those strings in different contexts, is not clear).

Ok. If you agree, I will update our (Catalan) language pack with these strings, and also with the newest changes (microphone blocks...)

Thanks

Joan

PR #2351, opened 2019-03-10
costumes: add full alphabet

brianharvey:
Not the digits because they aren't all available in a single font in
Scratch.

### DIFF
--- a/Costumes/COSTUMES
+++ b/Costumes/COSTUMES
@@ -1,41 +1,42 @@
+a-block.svg	a block
 abby-a.svg	abby a
 abby-b.svg	abby b
 abby-c.svg	abby c
 abby-d.svg	abby d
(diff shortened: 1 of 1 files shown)


jmoenig:
I'm going to close the PR, because adding additional costumes to the catalog is going to further slow down importing costumes. Instead I tried to address the underlying issue in two ways:

1) I also took out the "b block" costume from the catalog, presuming it shouldn't have been in there in the first place c4e9518da2e4019867c583b329e60a015fc28e08

2) I added a small "text costumes" library that lets you quickly create arbitrary word- letter- or number-costumes yourself, directly inside Snap! ecdfdf75389c8eae495f237c4b2b0d9e2eee9d95

Hope that's okay with you.

brianharvey:
Whatever.

PR #2352, opened 2019-03-10
VEE example no methods

brianharvey:
Changing the procedures in this project back from local (methods) to
global, just to remove a small barrier to understanding (the location
icon).  Also at some point we seem to have lost the
show-code-on-space-bar feature.


jmoenig:
thank you!

PR #2355, opened 2019-03-14
Atomic HOF names

brianharvey:
Changes the names to match the non-flash versions.


jmoenig:
Reluctant about this. The atomic versions are for adults. I never liked plastering the labels of KEEP with "items such that" or "items of" when speaking of lists.

brianharvey:
> The atomic versions are for adults.

Why do you say that?  You use them demoing for kids.  The kids, some of them anyway, have seen the tools and will find what you're doing easier to understand if the names agree.  And your COMBINE wording is just bad English.  (Pointed out by a kid on the Scratch forum.)

jmoenig:
I hate COMBINE and KEEP. If it were for me I'd use FOLD and FILTER. That's what students will find on Wikipedia

https://en.wikipedia.org/wiki/Fold_(higher-order_function)
https://en.wikipedia.org/wiki/Filter_(higher-order_function)

So I'm just naming those atomic HOFs that way so you *can* know which corresponds to which.

brianharvey:
Geez, it's pretty late in the game to hate COMBINE and KEEP!

When I read the Wikipedia articles, I see things like REMOVE-IF, REMOVE-IF-NOT, FOLDL, and FOLDR as actual procedure names (as opposed to Wikipedia page names).  So I don't think the appeal to authority is working for me here.

You can of course call procedures anything you want, for your own use.  But if we are providing something as a library, I think that, absent a really compelling reason otherwise, we should be consistent _with ourselves_.   Look, for example, at the stream library.  When I first wrote those functions, as a project for myself, I used the Scheme names.  But when I made a library, I changed to using our names (with the word "stream" added where appropriate).

jmoenig:
Whaddaya mean with "pretty late"? I've been against ```keep``` and ```combine``` forever!

But okay, okay. Did you just change the names or anything else?

brianharvey:
Just the names.  Thanks.

jmoenig:
Aw man, I hate it. What's wrong with SORT, why doesn't it have ITEMS OF anywhere? :-)

brianharvey:
:-)

jmoenig:
now online in dev. You probably need to clear cached images and files in your browser to get the latest library version. Later I'm going to push yet another version that copies the translations from tools over...

brianharvey:
Gotcha.  Thanks.

PR #2358, opened 2019-03-17
Another (the last?) Catalan language pack update before Snap5

jguille2:
Hi @jmoenig !

- Updating Catalan language translation pack with the last features and changes for Snap5 (microphon...) and also added the _Animation_ library Catalan translation.
- Other missing strings added (libraries names and descriptions, symbols names...)
- Waiting for decisions on #2341 . I've updated the translation according to the latest changes on this PR, spliting those strings by phrases (not by lines). I hope that is right now and that PR can also be merged.

Thanks!
Joan


jmoenig:
Thanks Joan @jguille2 ! I'll pull this asap, once the issue of what the final naming for the microphone dropdown's items is going to be (https://github.com/jmoenig/Snap/issues/2354)
Oh, and I promise to think about the long string :-)

jguille2:
Ok Jens.

If you want, you could leave it unmerged until (and before) next release. Then, I will continue committing this PR with the new changes. I don't want to be so heavy, making PRs all the time about this... but without a calendar... I'm trying to do the best (also in our translation) for the release date (also to update Snap4Arduino with it).

About the other PR (long strings translation), we can think deeper later (maybe in Olot we can work and share some Snap! themes: translations, videoMotion... (colors no! :disappointed_relieved:  )... But also, I think that PR is just a simple implementation of our current sytem, to allow some translations, and it is not a commitment for future decisions.

Continue...

Joan

PR #2361, opened 2019-03-24
make-vars fix: check that name is a string

brianharvey:
Our forum is full of kids who want to push the limits of Snap! and try
to break it...


jmoenig:
it's mostly one kid and these efforts show why not every carefully designed feature of the IDE should be turned into a quick-n-dirty block.

brianharvey:
Well, you won't be surprised that I disagree; the problem with this particular library is that I wrote it for an adult (who needed it yesterday; you're right about that part) and so didn't think about bulletproofing it.  I guess I have to learn to imagine a teenager looking over my shoulder as I program.  :-)  But the real point is that a kid who writes WHEN <1 = 1>, FOREVER... knows that he's disclaiming any expectation that he won't kill his browser.  He'd probably like nothing better than to kill his browser and tell us about it.  :-)  (By the way, I thought clicking the stop sign would prevent WHEN blocks from running until a script is started some other way, but that seems not to be the case.)

jmoenig:
RE:
> (By the way, I thought clicking the stop sign would prevent WHEN blocks from running until a script is started some other way, but that seems not to be the case.)

 ```WHEN``` hat blocks fire whenever their predicate evaluates to ```true``` and their script isn't already running (basically ```WHEN``` hats are always thread safe).

untitled script pic (8)

You can click the red stop sign button to stop *all* ```WHEN``` hats from firing. Pressing the red stop button again or pressing the green flag button resets it (not just running any script).

PR #2369, opened 2019-04-07
threads.js fixes #2368

### LINKED ISSUE #2368: (unicode (unicode of []) as letter) runs forever
untitled script pic
This (the string input is blank) will run forever (presumably - I let it run for 3 minutes and it didn't finish). This occurs both in the stable and the beta.

Using Chrome 73.0.3683.86 on Windows 7

brianharvey:


src/threads.js
```diff
@@ -2817,9 +2817,9 @@ Process.prototype.reportUnicode = function (string) {
     var str = isNil(string) ? '\u0000' : string.toString();
 
     if (str.codePointAt) { // support for Unicode in newer browsers.
-        return str.codePointAt(0);
+        return str.codePointAt(0) || 0;
```
jmoenig: why zero and not an empty string?

brianharvey: In general, the block returns a number, so it should return a number in this case also.  And I thought line 2817 was an attempt to return zero, by creating a string with a null character in it!  So I just made that work.

jmoenig: Oh wait! it's the *reportUnicode* function, duh. You're right, nevermind. Thanks!

jmoenig:
thanks! Online in the beta now...

PR #2376, opened 2019-04-16
Fetching a user's public collections now works

bromagosa:


jmoenig:
excellent, thank you!

PR #2379, opened 2019-04-20
Costumes

brianharvey:
The svg problems were almost all about the bounding box being wrong.
The bitmap problems were all being double size.

We should make editorial decisions about Scratch 2 vs Scratch 3 costumes, and real people vs cartoons, and all that, but this is just to get it so nothing is terrible.

One political decision: I removed the white breakdancer, deemed culturally insensitive if there isn't also a black one.


jmoenig:
Yessss!! this is awesome, Brian!

PR #2396, opened 2019-05-18
Catalan translation update with video motion blocks

jguille2:
Maybe the last before Snap! 5 release :smile_cat: 
Thanks,
Joan

### DIFF
--- a/locale/lang-ca.js
+++ b/locale/lang-ca.js
@@ -577,6 +577,18 @@ SnapTranslator.dict.ca = {
         'mode turbo',
     'flat line ends':
         'puntes de llapis planes',
+    'video %vid on %self':
+        '%vid del vídeo en %self',
+    'motion':
+        'moviment',
+    'snap':
+        'instantània',
+    'set video transparency to %n':
+        'fixa la transparència del vídeo a %n',
+    'video capture':
+        'captura de vídeo',
+    'mirror video':
+        'mirall sobre el vídeo',
     'filtered for %clr':
         'filtrat per a %clr',
     'stack size':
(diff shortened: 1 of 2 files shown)


brianharvey:
Good point -- do we have a mailing list of translation contributors?  They should all be asked to update before 5.0...

jguille2:
Oops! I forgot to update date changes info.

jguille2:
And yes, Brian. I think this is a pending job.
We have translator contacts in our "locale" files. But I think these is not a current list (with people active doing this translating job). And before sending mails some things should be defined.

But if you (Brian and Jens) agree, I can do this coordination work. I share my thoughts here:
- Initial tasks:
  - Contact translators and define a list of Snap! languages and their current translation coordinator.
  - Add a "Snap! translation" category in our Forums. This should be the place to coordinate translations: a category tracked by translators to coordinate their job with Snap! Team and also to handle requests from users from each language.
  - And to do this, I think we should establish a calendar rule to facilitate their work. I just want to say that before each release, it would be necessary to announce it to the forum a while before (perhaps a week or two) so that all those who want, can push the updated translations. At least, to do this before major releases.
- Other tasks
  - I want to facilitate the work of translators by creating myself all the translation files. Always keeping all the work done but with all the updated strings (those that are not, they will appear in English - English). This will greatly facilitate the work, avoid errors and also facilitate the maintenance of the code (when we want to update the original strings).
  - To do this, we will need some discussions and decisions. Maybe reopening something (no all) of #1551 and remaking some local tool like [this](https://github.com/jguille2/SA5/tree/master/snapTranslations) that was made with this idea.

Continue...
Joan

brianharvey:
If you want to take this on, go for it!  I can't imagine how anyone would object.  Thank you so much!

PR #2398, opened 2019-05-27
Improve help for list utilities library.

brianharvey:
Also remove EMPTY? (now primitive) and improve the algorithm for sort
(splitting the list in one pass instead of two).


jmoenig:
Thanks! Will pull asap...

PR #2401, opened 2019-06-01
new help screens for new blocks

brianharvey:


jmoenig:
Thanks! Will pull asap. Now working on making the new reporters both watchable and inheritable...

PR #2408, opened 2019-06-08
Fixing 'Stage' name fixed in World Map library

jguille2:
Hi,
WorldMap library is not running when stage name is not "Stage".

_Update_ block is now
mapUdate1
 and this PR change it to
mapUdate2


jmoenig:
Brilliant analysis and solution, as usual! Thank, you, @jguille2 . Will pull asap.
Hey, if I'm not responding during the next 2 weeks it's because I'll be on vacation and probably offline or someplace far away. Don't hesitate to fix more bugs you encounter :-)
Thanks!

jmoenig:
merged, I also updated the World Map demo example with your fix. It's also online now, but you'll have to clear your cached files in the browser to get your fixed version.
Again, thank you!

PR #2410, opened 2019-06-09
Proposal for WorldMap library

jguille2:
After fixing some issues on the WorldMap library, I want to propose some changes:

- The World Map is always there virtually (with its location, zoom and style). Now that is the case. So, the new "main" blocks are **show** and **hide** to use this feature over the stage. We still have the "map as costume" to use it wherever you want.
- And now, I can manage this virtual map (changing location, zoom, style... moving it, going to current location...) without forcing its visualization. If the map is visible, it will be updated. And if the map is hidden, the virtual map will be updated, but it will be not displayed (yet)
- If I hide WorldMap, the stage will be take its last costume (what it had before the map). So, that "lastCostume" is stored inside the worldMap object. We can continue using the virtual map (doing calculations, modifying parametres...) and re-viewing whenever we want.
- Longitude and Latitude are now shown always in the same order (lon-lat). This is for input slots, lists and also for the order inside the palette. This order has been chosen for its relation with x and y coordinates.
- "change Map by x: and y: pixels" has been fixed to have the same behavior over x and y coordinates.
- All the blocks (current and new) are translated to Deutsch and Catalan

Then, this proposal is:
MapBlocksProposal

If you want to test it...

Joan


jguille2:
Only a last change.
"change Map by x: and y: pixels" is changed for "move" because it seems to be clear (we are moving the map)
So the last library block will be
moveMap
and someone will have to verify if the German translation is fine
moveMapDeutsch

That's all. Thanks!

Joan

jguille2:
Hi!
I only upload xml library file, to be able to try it more easily.
[maps_module.xml.zip](https://github.com/jmoenig/Snap/files/3270770/maps_module.xml.zip)

jguille2:
Or even more easily [here](https://snap.berkeley.edu/snapsource/dev/snap.html#present:Username=jguille2&ProjectName=WorldMapLibrary%20Proposal)

Joan

brianharvey:
I like this.  Especially show/hide/shown.

While we're renaming things, I keep thinking "current location" means the location of the current _sprite_; what if we change "current" to "user's"?

In the "distance in km to" block, "Stage" in the menu is useless, but it would be useful to change that option to "Origin" or "Map Center."

"Move" is better than "change," but it's still not obvious what moves relative to what.  Can we get one of those fist icons you see in map software when you're dragging the mouse on the map?  It would go before "move."

jguille2:
Hi Brian,
I try to update it with your proposals. So...
- **current location** to **user location**
Then, the renamed blocks are
userLocation1
Jens will have to check my "german attempts" :)
userLocation2
I try to use emojis with the "gps-location" icon
location-pointer2
but I think there is no any good emoji to this.
- **distance in km** block. I have not changed the menus (I can't without adding another custom menu) but I have changed their behavior. **Now "Stage" is the "center of the map"**. So, "distance from sprite to stage" is the distance of that sprite to the center of the map. No more "-1" results.

- Emoji for the "move map" block. Not easy! Your proposal (fist icon) is not an emoji (oncoming fist and raised fist is not appropriate). I want this other
draggingIcon2
but it is not either an emoji.
It's not perfect... but I use this other to follow your idea...
moveMap


Continue... You can test again with the [updated project](https://snap.berkeley.edu/snapsource/dev/snap.html#present:Username=jguille2&ProjectName=WorldMapLibrary%20Proposal)

Joan

brianharvey:
I found "back of hand" 🤚U+1F91A, UTF-8 F0 9F A4...

It's really quite astonishing that this isn't in the "technical symbols" section of Unicode!

jguille2:
Then, it was
moveMap2
I await your decisions to commit this change.

Are the other changes right for you?

brianharvey:
Fine with me. Jens?

jmoenig:
I‘m vacationing in Tbilisi this week, sorry. But imo this has been getting worse instead of better. I likes Joan‘s first version.

Von meinem iPhone gesendet

PR #2432, opened 2019-06-29
Support horizontal scrolling (hold shift)

towerofnix:
This PR makes Snap! scroll elements horizontally when you hold shift - it's a standard interaction in almost all desktop and web apps. It also make it so Snap! uses the standardized `'wheel'` event, instead of non-standard/obsolete alternatives.

In actuality, this PR is extremely similar to one I made for Scratch 3.0: LLK/scratch-blocks#1670.

Tested in Firefox and Chromium! Definitely worth checking on a device with a trackpad (i.e. a macbook) before merging; I haven't got access to one ATM, so I haven't tried myself.

/cc @cycomachead!

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -11456,16 +11456,22 @@ HandMorph.prototype.processMouseScroll = function (event) {
         morph = morph.parent;
     }
     if (morph) {
+        var x = event.deltaX;
+        var y = event.deltaY;
+        if (event.shiftKey && event.deltaX === 0) {
+            // Scroll horizontally (based on vertical scroll delta). This is
+            // needed for some browser/system combinations which do not set
+            // deltaX.
+            x = y;
+            y = 0; // Don't scroll vertically.
+        }
+
+        // Multiplier variable, to account for both pixel and line deltaModes.
+        var multiplier = event.deltaMode === 0x1 ? -1/3 : -1/53;
+
         morph.mouseScroll(
-            (event.detail / -3) || (
-                Object.prototype.hasOwnProperty.call(
-                    event,
-                    'wheelDeltaY'
-                ) ?
-                        event.wheelDeltaY / 120 :
-                        event.wheelDelta / 120
-            ),
-            event.wheelDeltaX / 120 || 0
+            y * multiplier,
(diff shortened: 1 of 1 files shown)


cycomachead:
Well I certainly like that this adds features while reducing code. :)

Minor thing but do you have a link to the reasons for the multiplier values?

towerofnix:
>Minor thing but do you have a link to the reasons for the multiplier values?

There's no hard resource explaining it, but there's some discussion in LLK/scratch-blocks#1190. In actuality we were already basically doing this math in the existing Snap! code - note we were dividing by 120 or 3 depending on if the event was from Chrome or Firefox. The point is just to adjust the value to a range that Snap! is roughly expecting, instead of some really giant value like 50, which would result in a very big scroll distance. :P

jmoenig:
Thank you!

towerofnix:
No problem! :tada:

src/morphic.js
```diff
@@ -11456,16 +11456,22 @@ HandMorph.prototype.processMouseScroll = function (event) {
...
+        if (event.shiftKey && event.deltaX === 0) {
+            // Scroll horizontally (based on vertical scroll delta). This is
+            // needed for some browser/system combinations which do not set
+            // deltaX.
+            x = y;
+            y = 0; // Don't scroll vertically.
+        }
+
+        // Multiplier variable, to account for both pixel and line deltaModes.
+        var multiplier = event.deltaMode === 0x1 ? -1/3 : -1/53;
```
cycomachead:
```suggestion
        var DOM_DELTA_LINE = 1
        var multiplier = event.deltaMode === DOM_DELTA_LINE ? -1/3 : -1/53;
```

cycomachead: This is my only real suggestion on the code. The rest looks fine.

cycomachead:
Sorry for the delay, tested and test seems fine on a Mac with a trackpad.

towerofnix:
No worries! Thanks for the review. I made the fix you suggested (plus a semicolon missing from your suggested changes :P).

cycomachead:
Good catch. 👀

PR #2433, opened 2019-06-29
fixes collection grid page size

bromagosa:
A stray uppercase character and a missing | were causing collection grid sizes to be ignored by the backend:

image

This fixes it :)

### DIFF
--- a/src/cloud.js
+++ b/src/cloud.js
@@ -982,7 +982,7 @@ Cloud.prototype.getUserCollections = function (
             '/collections?' +
             this.encodeDict({
                 page: page || '',
-                pageSize: page ? pageSize | 16 : '',
+                pagesize: page ? pageSize || 16 : '',
                 matchtext: searchTerm ? encodeURIComponent(searchTerm) : ''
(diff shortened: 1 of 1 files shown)


src/cloud.js
```diff
@@ -982,7 +982,7 @@ Cloud.prototype.getUserCollections = function (
             '/collections?' +
             this.encodeDict({
                 page: page || '',
-                pageSize: page ? pageSize | 16 : '',
+                pagesize: page ? pageSize || 16 : '',
```
jmoenig: haha! JSHint was complaining about this but I thought to myself "geez, Bernat surely knows something I don't" :-)

brianharvey:
Sigh, languages with syntax are such a pain!

But, that reminds me, we should invent bitwise logical operators, in the math functions block.

jmoenig:
:-)

There's not much to invent for bitwise operators, we just add them. The only question is, do we put them into a library or do we make them primitives? I'm - hesitantly - leaning towards a library.

PR #2455, opened 2019-07-10
Removing 'any message' from messageSpecialOptionMenu

jguille2:
This PR changes 'message' _specialOptionMenu to avoid 'any message' option.

It really changes from "messagesReceivedMenu"  to " messagesMenu" _choice_ .

The reason is that "any message" has no sense (I think) here. Our custom blocks will treat as a "string message" and it will not have the possibilities of our hat "When I receive" block

messageSpecialOption

Thanks

Joan

### DIFF
--- a/src/byob.js
+++ b/src/byob.js
@@ -324,7 +324,7 @@ CustomBlockDefinition.prototype.dropDownMenuOf = function (inputName) {
             fname = this.declarations.get(inputName)[2].slice(2);
             if (contains(
                 [
-                    'messagesReceivedMenu',
+                    'messagesMenu',
                     'objectsMenu',
                     'costumesMenu',
                     'soundsMenu',
@@ -2655,7 +2655,7 @@ BlockLabelFragment.prototype.hasOptions = function () {
 BlockLabelFragment.prototype.hasSpecialMenu = function () {
     return contains(
         [
-            '§_messagesReceivedMenu',
+            '§_messagesMenu',
             '§_objectsMenu',
             '§_costumesMenu',
             '§_soundsMenu',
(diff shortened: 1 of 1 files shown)


jguille2:
Hi Jens,
I've added a new commit for backward support. Projects on Snap 5.0.0-5.0.3 can have used that "messageReceivedMenu"

Now, this is not an option (the "message" option is really "messageMenu") but is some project had that (the project does not store "message"; it stores "§_messagesMenu"/"§_messagesReceivedMenu") it still goes ok with it.

Joan

jmoenig:
excellent, thanks, Joan! I don't think there will be many - if any - projects with such a saved option, but it's certainly nice (and right) to respect prior official releases.

jguille2:
Yes!
I've just seen these details because I'm building the "new SnapJr" for the Citilab team. Now I've removed all the "hacks" and it is a "simple" Snap! project.
If you want to take a look ... It's still a draft: [SnapJr 5.beta](http://snap4arduino.rocks/dev/#loadJr)

Ah! and I've also implemented new specialOptionMenus for Snap4Arduino, to use the "autodetection pins" dropdowns.

specialOptionMenusS4A
 
Continue...

Joan

jmoenig:
Oh, that's really cool! I love SnapJr!!

DyslexicAwe:
> Oh, that's really cool! I love SnapJr !!

Me, too!! @jguille2 here is my first 'bug report'. I set the ChoosePenColor (for lack of a better name) block to red color, but the trail was still black. The size, when I tried to set it by SetPenSize was over-ridden, too, by the PenDown because it contains besides the pen_down, also the set_pen_color and set_pen_size primitives, I have remixed them. You can see it here https://snap.berkeley.edu/project?user=kinestheticlearning&project=SnapJr_by_Joan

jguille2:
Hi @DyslexicAwe 
Thanks for your comment.

Really, it was a first draft. If you have been playing with this [Snap4ArduinoJr-5.0-draft link](http://snap4arduino.rocks/dev/#loadJr) these days, a lot of changes have been done.

That job is finish (for now). I think current version (that link) will be the released (Snap4Arduino 5.0 will be release in a few days!)

You can take a look. Thanks!

Joan

PR #2457, opened 2019-07-17
fixes bug when fetching collection list from community site

bromagosa:
This was totally my fault. I can't remember when I broke this, but I did :disappointed:

It'd be great if it could be integrated quickly, as right now nobody can add any projects to collections because of this... :(


bromagosa:
This is what they're getting now:

imatge

jmoenig:
excellent! I'll pull this for a maintenance release later when I find a few minutes, hang on!

PR #2467, opened 2019-07-29
fixes importing OGGs in Firefox

bromagosa:
As reported in the forum: https://forum.snap.berkeley.edu/t/ogg-sound-files/681/6

This addresses this bug in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1240259

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -11771,7 +11771,10 @@ HandMorph.prototype.processDrop = function (event) {
                 readSVG(file);
             } else if (file.type.indexOf("image") === 0) {
                 readImage(file);
-            } else if (file.type.indexOf("audio") === 0) {
+            } else if (file.type.indexOf("audio") === 0 ||
+                    file.type.indexOf("ogg") > -1) {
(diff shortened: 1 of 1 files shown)


cycomachead:
Dang, the OGG format is pretty weird.... it can actually contain videos too. Though, I still think this makes the most sense - because it's not like Snap<em>!</em> could play a video if someone just dropped one it. (Though, a VideoSprite would be cool, I suppose.)

jmoenig:
Yes, when I discovered that OGG also supports video I thought that Firefox is probably doing the right thing by identifying every OGG file as - potential - video, since there's apparently no way we can tell what's actually inside. I don't have any strong feeling against this workaround, except that who know what people end up importing and what happens then. My personal take would be to let the browser decide what it thinks it know how to handle. Except that for MS Windows even that doesn't work. Have I mentioned lately that I'm in love with browsers? :-)
Marking this a "blessed" to be pulled shortly, thanks, Bernat!

PR #2470, opened 2019-08-05
Accounts for the scrollbars in ScrollFrameMorphs.

### LINKED ISSUE #638: View Doesn't Account For Scrollbars when setting bounds
(Ooops, didn't mean to hit enter that quickly!)

When you are viewing a list of projects and there are scrollbars, the scrollbar size isn't taken into account and covers up part of the content. Look at the last item of this list. (The last few letters are long project names are also cut off.)

covered

### LINKED ISSUE #804: Scroll Views don't Account for Scrollbar height
In the project list view, the scrollbars overlap with the content of the view. This is making it really difficult to click on text that is (in this case) the last item in a list.

### LINKED ISSUE #2468: Can't see last library
We want to use the last library in Snap, but it's not visible due to the scroll bar. 

image

Can this be fixed quickly, so we can use it ASAP?

cycomachead:
This ensures that the stopping point of the scrollbar is beyond the length of
the contents that it is scrolling, and that when scrolling is calculated, it
will now account for the presence of a scrollbar. This allows
you to scroll a little extra so you can always see all content.

Fixes #2468
Fixes #638
Fixes #804

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -1042,7 +1042,7 @@
     canvasses for simple shapes in order to save system resources and
     optimize performance. Examples are costumes and backgrounds in Snap.
     In Morphic you can create new canvas elements using
-    
+
         newCanvas(extentPoint [, nonRetinaFlag])
 
     If retina support is enabled such new canvasses will automatically be
@@ -1083,12 +1083,12 @@
     stepping mechanism.
 
     For an example how to use animations look at how the Morph's methods
-    
+
         glideTo()
         fadeTo()
 
     and
-    
+
         slideBackTo()
 
     are implemented.
@@ -1448,7 +1448,7 @@ function copy(target) {
     canvasses for simple shapes in order to save system resources and
     optimize performance. Examples are costumes and backgrounds in Snap.
     In Morphic you can create new canvas elements using
-    
+
         newCanvas(extentPoint [, nonRetinaFlag])
 
     If retina support is enabled such new canvasses will automatically be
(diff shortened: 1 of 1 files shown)


src/morphic.js
```diff
@@ -1042,7 +1042,7 @@
     canvasses for simple shapes in order to save system resources and
     optimize performance. Examples are costumes and backgrounds in Snap.
     In Morphic you can create new canvas elements using
-    
+
```
cycomachead: Sigh, this is my editor... but it's probably not bad to clean this up...

src/morphic.js
```diff
@@ -10578,8 +10578,7 @@ ScrollFrameMorph.prototype.adjustScrollBars = function () {
         vHeight = this.height() - this.scrollBarSize;
 
     this.changed();
-    if (this.contents.width() > this.width() +
-            MorphicPreferences.scrollBarSize) {
+    if (this.contents.width() > this.width()) {
```
cycomachead: So, `MorphicPreferences.scrollBarSize` was definitely wrong since a ScrollFrame could have its own scrollbar size. However, I don't think we actually want to account for the scrollbar here. If the content is any wider than the view we should just show a scrollbar. Account for the width of scrollbar would make sense if we were always adding padding on the right/bottom.

jmoenig:
very nice, thank you, Michael!
I'm gonna bless this for the next minor release (5.1.) which I'll do in September (after my vacation). There's going to be one more patch release sometime this week, but I'm only going to put things in it that obviously shouldn't break anything or have (next to) no side effects, like Simon's Chinese translation update, and Bernat's OGG patch. This will go into 5.1. along with the new help system and other updates. Yay!

jmoenig:
Okay, I've just given this some banging, and I like it so much I'lm going to pull it right now for this week's patch. Thank you, Michael!

cycomachead:
Ah, nice thanks for catching that!

PR #2471, opened 2019-08-06
Fix auto-playing audio not working on load

bromagosa:
Chrome doesn't want to play audio before some sort of user interaction happens, but we still need to call `resume()` on the audio context so that Chrome re-checks whether a user interaction has happened.


jmoenig:
yep, will pull within the next days as discussed :-)

PR #2473, opened 2019-08-09
Update threads.js

michaelolifant:
add factorial into monadic


jmoenig:
factorial is nice and easy to write yourself in Snap!, especially when doing it recursively as you have done. If we ever needed to make it a primitive we'd use a non-recursive optimized version. Why do you think we need to add factorial?

Issue #2476, opened 2019-08-18
Untouchable sprite with 1px dimension


DarDoro:
Untouchable sprite with 1px dimension.
Sprite with one of the dimension = 1px is untouchable. 
Showcase project 
https://snap.berkeley.edu/snap/snap.html#present:Username=dardoro&ProjectName=untouchables
Every sprite in this project touch each other so sensing should report "touched" for every case.

More generic problem causing this behaviour: first row and column of every sprite is ignored during touching tests.

Reported at forum https://forum.snap.berkeley.edu/t/color-touching-color/724

jmoenig:
thanks to your merged PR we can now close this. Thank you, @DarDoro, for this good catch!

PR #2477, opened 2019-08-18
Sprite touching : first row, column included, Issue #2476

DarDoro:
Sprite "touching" detection now start at (0,0 not (1,1 . #2476


jmoenig:
thank you, @DarDoro, good catch!

PR #2479, opened 2019-08-22
Fix: "ask" spawns as many prompt dialogs as number of clones

bromagosa:
The ask block currently spawns as many prompt dialogs as clones an object has. This fixes it :)


bromagosa:
Closed because it doesn't actually fix it. I don't know how I checked! :(

PR #2481, opened 2019-08-26
Catalan locale update

bromagosa:
It also includes translations for the list utilities, big data and World map libraries, as I needed those to be in Catalan for the course I'm writing.


jmoenig:
thank you, @bromagosa !

PR #2488, opened 2019-09-09
Remove extra comma in Galician translation

bromagosa:
It was preventing the translation from loading at all. Reported by @mbouzada.

### DIFF
--- a/locale/lang-gl.js
+++ b/locale/lang-gl.js
@@ -1024,7 +1024,7 @@ SnapTranslator.dict.gl = {
     'detach from':
         'soltar de',
     'detach all parts':
-        'soltar todas as partes',able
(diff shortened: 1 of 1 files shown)


bromagosa:
Pinging @jmoenig.

The current Galician file is broken and won't load at all, so Galician kids haven't been able to use Snap! in their language for quite a bit now :(

jmoenig:
Ah, yes, that's totally on my radar (as are the other open issues and pull-requests). Thing is, I'm - somewhat superstitiously - pausing all commits and code changes until after Google addresses the ugly ```getImageData()``` bug. Hang on!

jmoenig:
Okay, screw it, I'm not gonna wait for Google anymore...

PR #2497, opened 2019-09-25
Show a link to the community page for current project

cycomachead:
If you click the cloud menu, then you can get to the community
page and have it open in a new window.

This uses the URL parameters so it's not dependent upon being
logged in, and doesn't track any state.

Screen Shot 2019-09-25 at 2 47 55 PM

### DIFF
--- a/src/cloud.js
+++ b/src/cloud.js
@@ -47,8 +47,8 @@ function Cloud() {
 }
 
 Cloud.prototype.init = function () {
-    this.urlBasePath = '/api/v1';
-    this.url = this.determineCloudDomain() + this.urlBasePath;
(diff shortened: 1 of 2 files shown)


src/cloud.js
```diff
@@ -47,8 +47,8 @@ function Cloud() {
 }
 
 Cloud.prototype.init = function () {
-    this.urlBasePath = '/api/v1';
-    this.url = this.determineCloudDomain() + this.urlBasePath;
+    this.apiBasePath = '/api/v1';
```
cycomachead: Not strictly necessary, but I think it's helpful to differentiate the API routes from other pages.

bromagosa:
LGTM! :)

Issue #2492, opened 2019-09-14
Editing costume type (SVG <-> bitmap) not working properly


coproc:
Changing a costume's type (SVG <-> bitmap) while editing it does not work as expected. The original costume is not changed. When changing from bitmap to SVG the newly created costume is appended at the end of the list of costumes of the sprite. When changing from SVG to bitmap the newly created costume will only be selected as the active one (the sprite will "wear" it), but the new costume is not added to the available costumes of the sprite.

What could to be done to fix that:
- When a costume's type changes while editing, a new costume (Costume or SVG_Costume) is already created. But the old costume should also be removed from the sprite's costumes.
- The newly created costume should be added with the old name at the position in the costumes' array of the old costume.

coproc:
I have [a fix](https://github.com/coproc/Snap/tree/edit-costume-type) in my fork of Snap. But I guess my work needs a review/some feedback for further refinement before you would accept a pull request. Who can do that?

Let me give you an idea what I have been doing:
One of the problems was the asymmetric handling of bitmap and vector costumes. The vector costume was created by the bitmap editor when changing to vector type. But at this place the context of the editing (new sprite or new costume or edit existing costume) is not known. The existing callbacks only worked for some use cases (actually only for creating a new costume).

So I decided to create all editor callbacks in the gui events, where the context is known. When the costume type is changed inside an editor, the callbacks are just forwarded to the new editor. To allow this, it was also necessary to generate the costume after editing (from the data sent to the callback by the editor).

Any (constructive) feedback is welcome!

PR #2500, opened 2019-09-27
Replace costume when type changed during edit

coproc:
As discussed with bromagosa at SnapCon I propose these changes to fix issue #2492 ("Editing costume type (SVG <-> bitmap) not working properly").

The following will improve:
* new costumes will always have a unique name of the form 'Untitled' or 'Untitled(1)' ... (currently new SVG costumes will have default names of the form 'costume1', 'costume2', etc. not even guaranteeing a unique name)
* when editing a costume and changing its type while editing, the original costume will be changed as expected in all cases (currently when editing a SVG costume and changing its type to bitmap, the original costume will remain unchanged, although the sprite will wear the changed costume; and when editing a bitmap costume and changing its type to SVG the edited costume will be added as new costume)
* the code of VectorPaintEditorMorph does not depend on the Costume class any more
* the Costume class does not depend on PaintEditorMorph or VectorPaintEditorMorph any more

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -2364,25 +2364,32 @@ IDE_Morph.prototype.addNewSprite = function () {
 
 IDE_Morph.prototype.paintNewSprite = function () {
     var sprite = new SpriteMorph(this.globalVariables),
-        cos = new Costume(),
-        myself = this;
-
+        cosName = sprite.newCostumeName(localize('Untitled')),
+        myself = this,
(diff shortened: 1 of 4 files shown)


bromagosa:
Just to recap, this fixes https://github.com/jmoenig/Snap/issues/2492, among other issues :)

(thanks @coproc!)

Issue #2259, opened 2018-11-12
Typo in help for 'of'


ToonTalk:
drgging instead of dragging

brianharvey:
Thanks.

jasonappah:
It looks like the issue is still present, so I'll go ahead and fix that and make a PR if that's cool.

Here's a screenshot that I just took:

![Screen Shot 2019-10-07 at 19 52 02](https://user-images.githubusercontent.com/30608521/66359261-04d6b400-e93c-11e9-9f87-e4533bc7a04e.png)

PR #2503, opened 2019-10-08
Fixed typo in help screen of the reportAttribute block

jasonappah:
#2259 I saw that it hadn't been done, so I just went ahead and added the 'a' in Photoshop. Is there somewhere else I also need to update this?


jmoenig:
thank you!

Issue #1823, opened 2017-08-08
Manual PDF opens tab with title 'Microsoft Word - Snap! Manual'


ToonTalk:
In Chrome Version 60.0.3112.90 (Official Build) (64-bit)
![image](https://user-images.githubusercontent.com/2138696/29080047-b1bfc792-7c56-11e7-9926-ef255c3f4ce2.png)

After clicking on home page link to http://snap.berkeley.edu/SnapManual.pdf

This is particularly confusing when many tabs are open and the tab is abbreviated as just 'Microsoft Word'.

jasonappah:
I'll go ahead and make a PR for this. PDFs have a file name, and then they also have their own title attribute, which I edited in Adobe Acrobat and resaved. For some reason, when you export PDFs from Word, Word by default prefixes the title with "Microsoft Word -". I believe there is a way to disable this function in Word, but for now I will just reupload the corrected file.

PR #2504, opened 2019-10-08
Edited the PDF title of the manual.

jasonappah:
See #1823.


Barthdry:
I don't think the Snap Manual's name is required to be changed

brianharvey:
Too late -- there have been a lot of changes to the manual since that file to keep up with maintenance releases.  I'll try to remember to fix the pdf title next time I make a pdf.

PR #2514, opened 2019-11-04
No internal canvas error if inputs to new costume are 0, just return an empty costume

bromagosa:
The internal canvas error is not very human friendly. I think this block should do what the stretch block does: just return an empty costume.

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -5109,6 +5109,9 @@ Process.prototype.reportNewCostume = function (pixels, width, height, name) {
     }
     width = Math.abs(Math.floor(+width));
     height = Math.abs(Math.floor(+height));
+    if (width <= 0 || height <= 0) {
+        return new Costume();
+    }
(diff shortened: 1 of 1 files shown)


jmoenig:
interesting and a good idea. Have you tried this fix, does it actually work? I'm wondering, because the new Costume is also going to have a zero extent, aren't we getting the same error then?

bromagosa:
Yep, I tried it and it does work :)

jmoenig:
Excellent, thank you. Pulling now. Should we do a quick release?

bromagosa:
Sure, let's do it :)

Here's the fix working, by the way: https://youtu.be/0uAawt2a7Gg?t=359

jmoenig:
Oooh, cool video!
Okay, I'm preparing the release now. It has another pretty important fix, so that variable names such as "top", "item", "value" etc. will no longer be translated in the SET block's drop-down menu. That actually bit us yesterday at our workshop in Vienna :-)

Issue #2518, opened 2019-11-05
Menu translations are broken in 5.3.1


jguille2:
Hi Jens @jmoenig ,

[This commit](https://github.com/jmoenig/Snap/commit/c9b06bcb0cd1ffacb32b39d23091df73769a2d41) **breaks menu translation**.
![translationBroken](https://user-images.githubusercontent.com/2926315/68250370-0ff72f80-0021-11ea-9a39-de7c4ef3f45f.png)

MenuMorph.addItem localized by default, and now many items (most) are not translated.

I wanted to PR the solution... but I doubt between two solutions...

- Keep current behavior (_addItem_ function) and add `localize( )` to every menu item on _gui.js_

- Fix _addItem_ function. Then my proposal would be:
    + On _morphic.js - MenuMorph.prototype.addItem_ add a `translatable` param. If this is `true` or not defined (as up to now), that item is translated (as it was before). If `translatable` param is `false`, then applied the new code (not translating if items are not "marked").
     + On _blocks.js - InputSlotMorph.prototype.menuFromDict_ we can use the new param, and call `menu.addItem( )` with that `translatable = false`

If I can do anything about this... but this is a recent change, and I'm sure Jens wants to fix it and decide the right behavior.
Aside the problem, Jens, what's your idea about that "marked" options (to allow "consciously decided translations" of options)? I thought that this "mark" would be able to use on custom block options, but I don't know how to use it.

Joan

jmoenig:
OMG. This is bad. Working on it right now. Thanks, Joan!

jmoenig:
Fixed! Phew...

jguille2:
Hi Jens @jmoenig 
I'm sorry but think this is not fixed yet.

Now, the behavior is again the "original". Menus are translated but dropdown menus are also again translated.
I think we don't need all these "new two length arrays", because the "original solution" (checking 1-dim arrays) was fine. The problem was it was into the "general" addItem, and we need to be applied only on dropdowns (now possible thanks to the verbatim param).

I will try to make a proposal. Continue...

Joan

jmoenig:
Hi Joan, are you sure?
Which dropdowns are still translated? I thought I'd gotten them all, did I miss one?
Thanks!

jmoenig:
Ah, I see, you're referring to the drop-down menus in custom blocks. Okay, will fix!

jmoenig:
[...]

PR #2519, opened 2019-11-06
Fixing dropdown tranlation

jguille2:
Hi Jens!

You know, #2518 comments...

It's a proposal to fix the dropdown menus and the translation behavior, using your new param (verbatim) and avoiding all those changes on values (those _[name, false]_ arrays)

I've also added four new strings to Catalan locale.

Feel free to do what was better (I can resend Catalan changes)

As I said, the most changes are undoing latest changes, and the final code is more similar to yesterday code: only the new _verbatim_ param in _morphic-addItem_ and its use inside _blocks-dropdown creation_.

Joan

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -24,7 +24,6 @@
 * blocks: don't translate message names in drow-down menus
 * blocks: don't translate names in dropdowns
 * help: fixed size of MY help screen, thanks, Brian
-* byob: don't translate items in dropdowns
 * prepared release
 
 ## v5.3.1:
--- a/locale/lang-ca.js
+++ b/locale/lang-ca.js
@@ -185,7 +185,7 @@ SnapTranslator.dict.ca = {
     'translator_e-mail':
         'bernat@snap4arduino.rocks, jguille2@xtec.cat', // optional
     'last_changed':
-        '2019-11-05', // this, too, will appear in the Translators tab
+        '2019-11-06', // this, too, will appear in the Translators tab
 
     // GUI
     // control bar:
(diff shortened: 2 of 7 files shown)


jmoenig:
Thank you, Joan. I'll have a close look!

jmoenig:
at first sight this looks really, nice! I didn't see you reverting my changes to ```byob.js```, though. Surely we don't want those, either, right?

jguille2:
Hi Jens,
I've added a fix for submenus translation (current behavior neither work fine with them).
I look at _byob.js_ changes and recheck also everything.

Joan

jmoenig:
Thank you, Joan. As I've said, I like your way a lot, it's really what I was trying to accomplish yesterday and it doesn't need the kludge of introducing yet another marker (the two-item array). So, once you've got it all together I'll gladly pull all of this!
This might take me a few days, because for the rest of the week we're pretty busy with workshops again, until Monday. So, if I find the time I'll do it in between my activities, otherwise please don't be discouraged, I *will* pull this when I find the time. Thank you!

jguille2:
I've checket it.
Yes, these last two commits ([this](https://github.com/jmoenig/Snap/commit/9882d3dad724b79461bd95ed3cd71858398ea5ed) and [this](https://github.com/jmoenig/Snap/commit/1b5846992b1969bb6a770e420ce41a11e18fdc96)) are done after my PR.

So, if you want my proposal and you don't revert anything, we must clean these (on _byob.js_ and on _blocks.js_)

Also, if you want this line, I recheck everything to avoid more issues.

Continue...

jguille2:
Hi Jens,
Now this proposal is directly mergeable (undoing other changes made on master tree).

I will check thoroughly all these changes to avoid new surprises. Surely I will find a case not considered in any of the two proposals.

I will post here to confirm after checking.

Joan

jguille2:
Hi @jmoenig ,

Only posting that I have already done all my tests.

This PR is fine and ready to merge. All translatable and untranslatable strings are right.

I won't add more commits here.

Thanks,

Joan

jmoenig:
Thank you, Joan! Will pull as soon as I find the time, probably on Tuesday.

PR #2523, opened 2019-11-10
Rename help screen for ANSWER

cycomachead:
Not sure when this happened, but the help screen has the wrong name.
Noticed while looking through server logs!


jmoenig:
Thank you, Michael, for catching this. Will pull asap, probably on Tuesday...

Issue #2525, opened 2019-11-11
"Garbage Collection" for clones


sorenbug:
Is there a way to automatically delete clones that no longer have a reference? Or am I going to have to delete them myself? I'm running into the clone limit building a ray marcher because I've implemented vector instances as objects.

brianharvey:
I'm not sure what you mean by "clones that no longer have a reference."  All sprites have references from the display manager; clones have additional references from their parent.  I'm guessing you have some data structure of your own with pointers to clones.  And, yeah, if removing the clone from that data structure means it's useless, you'll have to delete it yourself.

jguille2:
Hi Anthony @sorenbug,
I think you have enough tools in Snap! to do what you want.
Some comments...
- You know you have all sprite clones referenced  by the "my clones" reporter. So, you can delete all your clones doing
![deleteClones](https://user-images.githubusercontent.com/2926315/68629488-58609280-04e4-11ea-9fe9-81df32cb0c01.png)
- Then, if you want to delete non-referenced clones, you only have to filter that clones list.
- For example, if you have a list of referenced clones, your not referenced clones are in
![notReferencedClonesList](https://user-images.githubusercontent.com/2926315/68629680-f2283f80-04e4-11ea-93f5-fe8ce10374b3.png)
- And if you have only one referenced clone, then
![notReferencedCloneOne](https://user-images.githubusercontent.com/2926315/68629733-2ac81900-04e5-11ea-9cf3-20f5db5e8d11.png)
- If you have different references, you can build a list of references... or do any other action you need.

Joan

sorenbug:
> P.S. I tried reading your project but it's way too complicated for me to work out what you mean about references.

Some clone creations are bounded to a certain scope but stay alive outside of that certain scope (see the local reporter `march` in the `Camera` sprite, and how it creates and scales `Vector3::Cartesian`s without destroying them).
<hr>
I think a good solution to this would be a `using` block, sort of like this, except the `close λ` would still run even if the commands inside error:

![image](https://user-images.githubusercontent.com/7751296/68631245-f598d800-049e-11ea-87d5-90638e53942a.png)

The only remaining concern if the blocks above were to work would be continuations. For example:

![image](https://user-images.githubusercontent.com/7751296/68631517-f4b47600-049f-11ea-8375-b8d0edf14f75.png)

If I were to model the _intention_ of this code in Scala
```scala
Vector // a singleton object representing
[...]

PR #2526, opened 2019-11-12
Fixing issue with 'constructor' variable name

jguille2:
Fixing problem with "constructor" variable name related to #2525

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -5922,7 +5922,7 @@ VariableFrame.prototype.silentFind = function (name) {
     answer the closest variable frame containing
     the specified variable. Otherwise return null.
 */
-    if (this.vars[name] !== undefined) {
+    if (this.vars[name] !== undefined && !(this.vars[name] instanceof Function)) {
         return this;
     }
     if (this.parentFrame) {


jmoenig:
yeah, this would work. What do you think about replacing the current ```vars``` object / dictionary with a ```Map``` structure? That's probably a lot cleaner and might even be a little faster. I'm leaning towards this...

src/threads.js
```diff
@@ -5922,7 +5922,7 @@ VariableFrame.prototype.silentFind = function (name) {
     answer the closest variable frame containing
     the specified variable. Otherwise return null.
 */
-    if (this.vars[name] !== undefined) {
+    if (this.vars[name] !== undefined && !(this.vars[name] instanceof Function)) {
```
jmoenig: I guess the idea is that every entry in the ```vars``` dictionary has to be of type ```Variable```. So, we could easily check that, which might be safer and easier. Let me investigate....

jmoenig:
I've decided to check for ```Variable``` type entries only, which is easier and seems to solve the issue. Thank you, @jguille2, for your excellent-as-always analysis of the underlying issue, you're a fantastic help!

jguille2:
Yes Jens,
Checking for  `Variable` type is clearly the right solution, and also is solving that issue.
I really appreciate your words. For me, it is a joy to be part of this great project.
Joan

PR #2528, opened 2019-11-15
fix broken recursion in variadic library

brianharvey:


jmoenig:
we both worked on this at the same time. Closing this

PR #2534, opened 2019-12-15
Add reporter version of safely try

brianharvey:


jmoenig:
Thanks! Will pull asap for 5.4

PR #2540, opened 2019-12-22
Some help screen issues

brianharvey:
Most of these reflect recent changes in the actual blocks. There are three things I'd like Dylan to work on:

changePenHSVA_graph.png: This picture, used in several help screens, includes annotations as pixels. Could they be made translatable?

changeScale.png: descender of "g" cut off. I noticed it on this screen, but it's a general problem.

doSwitchToCostume.png: I added a script, and as a result, the previous row extends past the right edge of the frame. I thought maybe it was becasuse of retina, but when I disabled retina mode the same thing happened when I compiled the xml.

Thanks!


brianharvey:
reportGetSoundAttribute.png: <br/>, more control over arrows
	Also, titles have moved to below the heading block.

reportPenTrailsAsCostume.xml: Script comes out too wide when compiled. 
	Scale="0.6" doesn't affect it.

reportIfElse.png: Script too wide. And no zebra coloring!

fork.png: xml is correct, png is missing a letter in "draw"

evaluate.png:  In the phrase "the input variable function" the word "function" should be in courier font.  (More generally, it'd be nice to be able to put HTML tags for font, color, etc. inside text.)  Oh, and up on top there's supposed to be an arrow between the two blocks.  (There are a few more like this in other blocks.)

doTellTo.xml: The last script had the first (sprite) input to tell empty.  I put in "Sprite(2)" as it should be, but when I compile it, the script is too wide.

reportAskFor.xml: I changed a word in the text, and when I compiled, all the scripts are too wide.

createClone.png:  This time the help is too /tall/, so the bottom edge is clipped.  (I didn't generate this picture; it came this way in the PR.)

PR #2544, opened 2019-12-28
Brian's help

brianharvey:
Hi


cycomachead:
@brianharvey @jmoenig @jadga-h 

Obviously, this is super old and the _code_ is now out of date... but what it we just take the images from this PR and update them? Or should we just copy these to the manual? They at least look quite a bit nicer since they have been re-rendered.

PR #2547, opened 2020-01-03
Minor Catalan update

bromagosa:


jmoenig:
Yay, thank you, Bernat!
So, we might end up with a minor release soon regardless :-)

PR #2552, opened 2020-01-16
Update lang-de.js

SJatzlau:
Changed translation for "when I start as a clone".

The German translation "wenn ich geklont werde" suggests that the object being cloned executes the attached script. 

However, in reality, that code is executed by the object that is created THROUGH the cloning process, NOT the original prototype. The English block text reflects this, but the German one did not!


jmoenig:
excellent point, Sven! Will pull asap for the next patch.
Thanks!

PR #2556, opened 2020-02-26
Custom js dropdowns

bromagosa:
JS functions inside the Input Options dialog are interpreted as such :)


jmoenig:
as discussed, thank you, Bernat!

PR #2561, opened 2020-03-20
Include a new library : MQTT

TopNiz:
This library fully functional helps deal with MQTT messaging protocol through Snap!.
It is also a good OOP example


TopNiz:
Additional demo projects are also added

jmoenig:
Wow, Thank you!

TopNiz:
It's a pleasure.
The demo project may lead to some confusions. So I have enhanced it to be completely functional and intuitive.
When this pull-request is merged, I'll proceed with and additional pull-request with the new demo project.

PR #2563, opened 2020-03-21
Fixing color.a value at pipette feature

jguille2:
Hi!
Fixing [forum issue](https://forum.snap.berkeley.edu/t/costume-editor-crashes/1651) about costume editor crashes using the _paint bucket_ tool.

The problem is in the _pipette_ feature.
Just fixing color.a value, that has a 0-1 range (not 0-255)

Joan


jmoenig:
Thanks, Joan. I'll pull this for the next release!

PR #2573, opened 2020-04-23
Add Hummingbird library to support EXCITE project

brianharvey:
This is for the BJC-for-robotics project we're cooperating with.


jmoenig:
Thanks, will pull for the next release!

PR #2579, opened 2020-05-07
Fixing Stretching to larger dimensions (Morphic2)

jguille2:
It is #2574 rebased to Morphic2
(tested - no conflicts - fast-forward mergeable)

If this way is right, I will continue rebasing and closing old PRs

Rewriting old notes...

Hi Jens,
Related to [this issue from Snap! forum](https://forum.snap.berkeley.edu/t/bug-found-in-stretch-block/1799)

- `Costume.prototype.stretched` creates a `new Costume` and then, this runs `shrinkToFit(this.maxExtent())`. So, new stretched costumes can not bigger than Stage dimensions.
- And, in addition to this limitation, the problem is that _Costume center_ is calculated with the unlimited dimensions.
- This solution is adding a parameter _noFit_ to _Costume function_. This ensures that it does not damage any other code. In fact, only two lines at _objects.js_:

```
function Costume(canvas, name, rotationCenter, noFit ) {
...
if (!noFit) {this.shrinkToFit(this.maxExtent()); }
```
and then
```
Costume.prototype.stretched = function (w, h) {
...
    stretched = new Costume(
        canvas,
        this.name,
        center,
        true
    );
```
That's all!
Joan

### DIFF
--- a/src/objects.js
+++ b/src/objects.js
@@ -9507,10 +9507,10 @@ SpriteBubbleMorph.prototype.fixLayout = function () {
 
 // Costume instance creation
 
-function Costume(canvas, name, rotationCenter) {
+function Costume(canvas, name, rotationCenter, noFit) {
     this.contents = canvas ? normalizeCanvas(canvas, true)
             : newCanvas(null, true);
-    this.shrinkToFit(this.maxExtent());
+    if (!noFit) {this.shrinkToFit(this.maxExtent()); }
     this.name = name || null;
     this.rotationCenter = rotationCenter || this.center();
     this.version = Date.now(); // for observer optimization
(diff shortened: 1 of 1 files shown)


jmoenig:
Hi @jguille2 , can you rebase this back to "main"? I think I'd like to get rid of this branch first and then merge everything directly into main. Thanks!

jguille2:
Hi!
Not necessary. I close this, because we still have #2574. This way it was only a proposal :)

And only comment that (aside branchs) I like rebasing sometimes, because it makes clearer histories. When PRs are quite old (to appear really when they were added) or when PRs have multiple commits (sometimes, fixes doing and undoing) to add them with a single commit.

Joan

jmoenig:
Thank you, Joan! I might need your help later...

PR #2580, opened 2020-05-08
filtering out typical BJC, Teals etc projects from published project list

bromagosa:


brianharvey:
This is a very mild sort of filtering; the projects still appear if you specify a username.  Teachers aren't going to be satisfied by this, I think.

PR #2582, opened 2020-05-10
fixing the translation of 'set y to %n'.

abee2:
missing 'y' in the translation of japanese and hiragana.


jmoenig:
Whoa, thank you, Abe-san! I will pull this asap for the next major release.
BTW there are now a number of new blocks with missing Japanese translations. Maybe we can team up to identify and update them...

abee2:
Thank you for the quick response. I'm sorry for letting the problem sit there for 9 years.
If you can add the words to be translated, I will do my best to accommodate.

PR #2584, opened 2020-05-10
Fixing dropping reporters in variadic C-shaped input slots

### LINKED ISSUE #2583: Dropping reporter in variadic C-shaped input slot displays weirdly
untitled script pic (2)

Only happens when variadic; single C-slots correctly fall back into the title line.  Fails for me in Snap<i>!</i> 5 and 6.

jguille2:
Hi Jens and Brian!
This fixes #2583 

I've tried different ways, but I think this is the most coherent (robust) behavior.
Comments...
- You know CSlots are basically for command blocks (and stacks of command blocks). But in custom blocks (not in primitives) these slots allow to snap a reporter. I guess the idea is to snap a reporter/variable that have commands. Anyway, this action (snapping) change the block shape (from a CSlot to a "normal" input).

- The problem is with a "multiple input-Cshaped slot". Here, CSlotMorphs parent is not a Block... their parent is a MultiArgMorph. The problem is here.

- And we can change all the CSlotMorphs (of that MultiArgMorph) to a "reporter slot" (to fix the visual problem reported by Brian) But if the other CSlots have commands, I will have a problem loosing previous code.

Then, the solution is easy. Avoid dropping reporters into CSlots when (only when) this input is a multiple Cslots one.

Just this:
+ Single CSlot continue with current behavior, taking reporters.
dropping1

+ And multiple Cslots do not accept reporters, but yes, we can set a reporter as the list value of all those Cslots.
dropping2

Joan

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -6582,7 +6582,11 @@ ScriptsMorph.prototype.closestInput = function (reporter, hand) {
         }
         return true;
     }
-
+    //checking Target to avoid variadic C-shaped slots for reporters
+    function targetChecked(target) {
+        return (target instanceof CSlotMorph && target.parent instanceof MultiArgMorph) ? null : target;
+    }
+            
     if (this.isPreferringEmptySlots) {
         if (hand) {
             handPos = hand.position();
@@ -6601,7 +6605,7 @@ ScriptsMorph.prototype.closestInput = function (reporter, hand) {
                                 !contains(blackList, input)
             );
             if (target) {
-                return target;
+                return targetChecked(target);
             }
         }
         target = detect(
@@ -6617,7 +6621,7 @@ ScriptsMorph.prototype.closestInput = function (reporter, hand) {
                                 touchingVariadicArrowsIfAny(input)
         );
         if (target) {
-            return target;
+            return targetChecked(target);
         }
     }
 
@@ -6632,10 +6636,10 @@ ScriptsMorph.prototype.closestInput = function (reporter, hand) {
                             !contains(blackList, input)
         );
         if (target) {
-            return target;
+            return targetChecked(target);
         }
     }
-    return detect(
+    target = detect(
         all,
         input => (input !== reporter) &&
             !input.isLocked() &&
@@ -6643,6 +6647,7 @@ ScriptsMorph.prototype.closestInput = function (reporter, hand) {
                     !(input.parent instanceof PrototypeHatBlockMorph) &&
                         !contains(blackList, input)
     );
+    return targetChecked(target);
 };
 
 ScriptsMorph.prototype.closestBlock = function (comment, hand) {


brianharvey:
Well, this would eliminate the bizarre appearance, but at the cost of breaking what the user was trying to do, namely, display a list of all his custom blocks.

Of course there are other ways to do that, by using a LIST block and explicitly ringing all the blocks or scripts.  Or making a custom block with a variadic Reporter slot, better yet.

So maybe this is okay?

jguille2:
Hi Brian,
I don't understand what you say about "trying to display a list of all his custom blocks". Maybe I explained myself wrong... or maybe I'm missing something.

This fix only avoid to add reporters to a CSlot (and only if this CSlot is a multi-input-Cslot).

As a user, what I would expect is to never be able to do that, as is the case with primitives, but I kept current  behavior on single-cslots-input.

If you want to test it here:

[http://jguille2.github.io/snap6Dropping/snap.html](http://jguille2.github.io/snap6Dropping/snap.html)

Joan

jmoenig:
Hi Joan and Brian.

Great minds... ;-) This was my first idea when I saw the issue, to simply prevent reporters from being embedded into sub-slots. There might be an even easier solution for that. Let me think about it a little.

OTOH I can also see that it would be nice to let users express complex scripts that would require them to drop reporters into individual C-subslots, so maybe actually solving the underlying issue might be a good idea, also.

Hmm... I need to think about this. But thank you, Joan! I might decide just pull your fix...

brianharvey:
I don't hate Joan's fix!  I'm just reporting that the particular user who reported this problem wants to be able to mix commands and reporters in the slots of a variadic C-shaped input.  We could decide to disallow that.  I have no opinion.

It occurred to me that we might consider inventing reporter C-slots.  I haven't thought that through at all.

But also, we keep running into situations in which the command/reporter distinction works against us, and down the road, When Things Slow Down™, we might consider ways to invent a general Procedure category that would encompass commands and reporters.  For example, Berkeley Logo has a special form called .MAYBEOUTPUT that evaluates its input; if that evaluation produces a value, .MAYBEOUTPUT reports it, but if not, it continues without erroring.  It's not something end users use much, but it's been super helpful to me in inventing library tools.

jguille2:
Thanks Brian,

I continue with some comments... but it is not to argue, just to have the right criteria (maybe for other issues).
- For me, "run" and "call" commands are good examples of our command/reporter  differentiation/confusion. Those rings (command or reporter rings) are only hints (suggestions). Because you can put inside any code.
- This ambiguity can be disputed, but is consistent both for primitives and custom blocks.
- But for me, beyond fixing current issue, I would expect that CSlots behavior was the same in primitives and custom blocks. Only commands.
- Then, despite this, I think for custom single cslots it's good to do this conversion (setting a reporter, change its shape, stopping to be a cslot). But it is only a helper! Because on our Snap! methafor, CSlots are only for commands (like primitives). So, for variadic cslots, I think it's clear. If some body wants a list of reporter inputs, he must choose a reporter slot, not a cslot. **The option says "Command (C-shape)"!!** 

Joan

jmoenig:
I've decided to follow Joan's suggestion and disabled dropping reporters into multi-C sub-slots for now, using a slightly simpler mechanism (isStatic). Thanks! https://github.com/jmoenig/Snap/commit/eaefe734abaa35142dda3ba5a31a31b6340142d9

jguille2:
Wow!
I don't see this!! I wasted my time following _snap_ function and fighting with Morph parents... and your solution is just simpler an elegant!

The only comment is I like more the other behavior. With this, if you bring (close) a reporter to the command block, it wants to go to the _MultiArgMorph_ (as the list of inputs) and this delete all the commands inside all the other Cslots. It's a bit dangerous.
The other way, the reporter must go to the "arrows zone" to _snap_ as a MultiArg, and nothing appears dragging it over the other slots (the same behavior that "normal" Cslots (primitives))

Joan

jguille2:
Hi @jmoenig ,

I don't want to be annoying... but I only want to explain visually what I pointed about "reduce the danger" of the currentDev implementation.

With our last current code:

testingInputsCurrentdev

An alternative (this is the result of my PR... but maybe we can check that code, but the result...)
testingInputsAlternative2

That's all,
Joan

jmoenig:
Hi Joan, yes, you're right, of course. The current "fix" is incomplete and not what we want. As soon as I get out from under the current flurry of urgent stuff bogging me down I'll revisit this. Re-opening your PR to keep this on the list. Thanks, you're never "annoying" at all!!!

jmoenig:
Hi Joan, I've attempted another fix here 88a5612934991ad12356c85c111adc3515401d06 let's see whether that works better. Thanks!

jguille2:
Hi Jens,
I've tested it... and now it's perfect!!

Thanks,
Joan

Issue #1657, opened 2017-02-09
checkbox not updated for watchers when watchers removed from stage


gengshenghong:
Steps to reproduce:
1. in "Sensing", find "answer" (for example only) block, check the checkbox and it will show the watcher on stage
![image](https://cloud.githubusercontent.com/assets/1018468/22768106/b7c50cd0-eeba-11e6-883e-1178d29a5a18.png)

2. on the stage, drag the watcher of "answer" to palette area and it will be deleted as expected.

3. check the status of "answer" block in palette, the checkbox is checked still (not refreshed), expected behaviour is to uncheck it.
In fact, if you check the checkbox again, checkbox will not change and still checked, watchers will pop up on stage again.

PR #2586, opened 2020-05-12
FixingWatcherStatusRefresh

### LINKED ISSUE #2585: Checkbox for "mouse x" watcher not behaving nicely
Create two sprites and show the motion palette in both. In one check the box to watch "mouse x". Then go back to the other sprie and the checkbox isn't checked. But clicking on it removes the watch.

jguille2:
Hi Jens!
This fixes #2585 (I see #1657 too)
The problem is watcher status changes were not updated on other sprites when required (for global variables and global reporters).

I also want to fix "sprite local var" representation on Palette, when there is also a global var with its own name. But it's better in another PR...

Joan


jmoenig:
Excellent, @jguille2 , thanks for this long-awaited fix!

PR #2587, opened 2020-05-13
Fixing localVars drawing on the palette

jguille2:
Hi Jens,
(Just commented on other PR and on the working papers about the new make-variables lib)

Behavior:
On Snap! palette, global and sprite vars are distinguishable. Fine! But when I have the same name for a global and for a sprite var, it’s not running well.
The shape looks global but really, it is the sprite var!
And yes! It must be the local one.
I can’t have both because we can only use the sprite var on the scripting area (the closest scope). But then, we need to draw the “local sprite shape”.

Code:
Really easy fix. Function `SpriteMorph.prototype.allLocalVariableNames` was filtering (with a more complex code) and really it must report "all local variables". I keep this function because it offers the "sort" option and then, changes are minimal. Anyway, nobody else call this function so we haven't any other problem with this change.

Joan

### DIFF
--- a/src/objects.js
+++ b/src/objects.js
@@ -6794,17 +6794,13 @@ SpriteMorph.prototype.hasSpriteVariable = function (varName) {
 };
 
 SpriteMorph.prototype.allLocalVariableNames = function (sorted) {
-    var exceptGlobals = this.globalVariables(),
-    	globalNames = exceptGlobals.names(),
-     	data;
+    var data;
 
     function alphabetically(x, y) {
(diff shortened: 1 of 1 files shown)


jmoenig:
Hi @jguille2 , good catch, thank you! Reading through my code I don't understand why I filtered out the globals in the first place, especially since the `reachableGlobalVariableNames()` function already addresses the issue of double-definitions. I'm pretty sure this had something to do with adding inheritance, but it's clearly wrong. Let's pull this and find out whether it breaks something else, which doesn't seem likely, as the function isn't called anywhere else.
Thanks!

PR #2591, opened 2020-05-19
Fixing slider position for lower values

jguille2:
Hi Jens,

It's a simple fix for [this forum bug report](https://forum.snap.berkeley.edu/t/slider-bug/1978).
Maybe you want to write it different... but this is tested and point (as a helper) to the code to change.

One question aside this, would it be better to set the default "slider min" to 0?
Variables default value is (or shows) 0, and changing the watcher to a slider, we have a 0 value in "the 1 position". If you click to the sliderbuttom (without movement) that value changes to 1 (because it is in "that 1 position").
Aside this comment, maybe is "natural" a default slider from 0 to 100. Isn't it?

If you agree this, I can set this 0 in this PR... to make it simpler :)

Joan

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -6933,18 +6933,24 @@ SliderMorph.prototype.fixLayout = function () {
         bh = Math.max(bw, Math.round(this.height() * this.ratio()));
         this.button.setExtent(new Point(bw, bh));
         posX = 1;
-        posY = Math.min(
-            Math.round((this.value - this.start) * this.unitSize()),
-            this.height() - this.button.height()
+        posY = Math.max(
+            Math.min(
+                Math.round((this.value - this.start) * this.unitSize()),
+                this.height() - this.button.height()
+            ),
+            0
         );
     } else {
         bh = this.height() - 2;
         bw  = Math.max(bh, Math.round(this.width() * this.ratio()));
         this.button.setExtent(new Point(bw, bh));
         posY = 1;
-        posX = Math.min(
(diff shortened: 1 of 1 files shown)


jmoenig:
Oooh, that was quick, thanks, Joan!

jguille2:
Hi @jmoenig , you've been fast too :smile:

Just for not forgetting the question, what do you think about the comment about "default slider min"?

jmoenig:
I think this fix is fine as it is, let's see how it fares...
Again thanks for the quick and perfect solution"

jguille2:
Hi Jens (@jmoenig)
Excuse my insistence :pray: but I see it so clearly ... and I think I'm explaining it wrong ...

Aside my past comments (about the success of having the default slider at 0-100), I guess current behavior (1-100) is only a small code "error".
See this...
At _objects.js_ ln 11195 we create the new sliders:
```javascript
        this.sliderMorph = new SliderMorph(
            0,
            100,
            0,
            20,
            'horizontal'
        );
```
And I think we want that (0-100) range.

But the slider builder at _morphic.js_ ln 6869 does
```javascript
function SliderMorph(start, stop, value, size, orientation, color) {
    this.init(
        start || 1,
        stop || 100,
        value || 50,
        size || 10,
        orientation || 'vertical',
        color
    );
}
```
And that `start || 1` is `0 || 1` and then 1. But we want 0, don't we?

I stop. Thanks,
Joan

PR #2600, opened 2020-06-04
Repair vee.xml

brianharvey:
Somehow a block got lost.  Dunno how or when.


jmoenig:
Hmm... how strange! Thanks for the catch...

jguille2:
Fine!
But there is no mystery at all :smile: 
We have this from [March 2019](https://github.com/jmoenig/Snap/commit/ac15df15360c57c2a91a43b26044f7b491f86b8f#diff-8ff9747c766529605a24802bf25fd206). Before this, _vee_ and the other _drawing_ blocks were local commands and we had not that second sprite.
Joan

jmoenig:
Haha! Well, look at that, Joan, great find!

PR #2602, opened 2020-06-08
Update bignums library

djsrv:
I've updated the library for hyper blocks, added some missing options to reportMonadic (neg, lg, and 2^), and switched the !== checks for NaN to the ES2015 Number.isNaN, which should be clearer.


brianharvey:
Thanks, Dylan!

jmoenig:
Thank you, pulling now...

PR #2606, opened 2020-06-10
remove pixels library

brianharvey:
We decided to do this, right?


jmoenig:
yes, but then I found that there's still the `snap photo` reporter in there, which some projects and kids love to use, so I've just removed the `show picture` command and left the library in for now.

PR #2607, opened 2020-06-11
typo in Catalan translation

bromagosa:
Just a minor typo in a source string.

### DIFF
--- a/locale/lang-ca.js
+++ b/locale/lang-ca.js
@@ -1164,7 +1164,7 @@ SnapTranslator.dict.ca = {
     // Project Manager
     'Untitled':
         'Sense títol',
-    'Open un Project':
+    'Open Project':
         'Obre projecte',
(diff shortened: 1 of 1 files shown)


jguille2:
Thanks Bernat,

But I see now this string is twice
(https://github.com/jmoenig/Snap/blob/master/locale/lang-ca.js#L1167)
and
(https://github.com/jmoenig/Snap/blob/master/locale/lang-ca.js#L2087)

Don't worry about this. I'm waiting for the latest changes on v6 and I'll submit all Catalan updates (I'm waiting to do a single commit and only one PR). Then, I'll also fix this.
If you agree I'll write "Obre un projecte".

Joan

bromagosa:
Then it should have been removed, you're right.

In any case, the key I corrected was the Catalan/English mix "Open un Project", which was obviously wrong :P

PR #2611, opened 2020-06-16
color is touching color should not be sprite only

space-elephant:
I found this when I was making the wiki article for sensing.


jmoenig:
No, that's totally intentional and also matches Scratch's example

PR #2613, opened 2020-06-18
Remove script type declaration

R4356th:
### Changes

Removes script type declaration from index.html and snap.html.

### Cause of Changes

The default scripting language for browsers is JavaScript. So, keeping declarations is unnecessary and a waste of storage.

### Test Coverage

Locally tested by editing the files with Notepad and running them on my browser (Firefox) afterwards.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <script type="text/javascript">
+    <script>
     if (window.location) {
       window.location = 'snap.html' + window.location.hash;
     }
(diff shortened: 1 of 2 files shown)


jmoenig:
actually ... that's a good idea!

It used to be that browsers kinda wanted you to specify the MIME type for scripts and all the guidelines heavily encouraged them, but since HTML5 have become embraced by all browsers we should be able to finally safely remove them.

Thanks!

R4356th:
Thanks, @jmoenig! Happy to help!

PR #2618, opened 2020-06-25
Add noscript tag with instructions on how to enable JavaScript

R4356th:
### Changes

Adds noscript tag with instructions on how to enable JavaScript.

### Cause of Changes

Users who have disabled JavaScript mistakenly or intentionally are unable to see anything currently. They might think that it is a Snap<i>!</i> bug, which it is not. Also, young users may not know how to enable JavaScript.

### Test Coverage

Locally tested by editing the files with Notepad and running them on my browser (Chrome) afterwards, and I can assure you that they do not appear when JavaScript is enabled.

### DIFF
--- a/snap.html
+++ b/snap.html
@@ -40,6 +40,6 @@
     </head>
     <body style="margin: 0;">
         <canvas id="world" tabindex="1" style="position: absolute;"></canvas>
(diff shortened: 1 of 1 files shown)


jmoenig:
naw, enabling JS and having a browser that supports canvas are both obvious prerequisites. Let's keep it simple. But thanks for the consideration!

R4356th:
Alright but what I was thinking was that children might be unable to figure that out.

PR #2620, opened 2020-06-27
Update Bengali translation

R4356th:
### Changes

Adds missing strings to Bengali translation, fixes spelling mistakes and spacing.

### Cause of Changes

The translation had got outdated.


R4356th:
@jmoenig, library does not have a good equivalent in Bengali. If I translate it, the meaning will get altered and will mean book library. I have changed that to block library for that reason.

locale/lang-bn.js
```diff
@@ -440,9 +440,9 @@ SnapTranslator.dict.bn = {
         'পুল',
```
R4356th, line 440: What are all these pull and push doing here?

jmoenig:
This looks like it's coming along great! Let me know if you need any information and / or help...

jmoenig:
How is the state of this PR? Are you ready to submit? I'd love to get your translation update into the release, and that's coming up in a few days. Keep in mind that I'll need some more time to test the final version. Anyway, thank you for your work in keeping Snap! up to date in Bengali! I really appreciate this.

R4356th:
@jmoenig, I am trying to find equivalents for some strings. And I am encountering a problem, GitHub Pages says that the build was succesfully deployed but when you go [here](r4356th.github.io/Snap/) and choose Bengali as the language, the strings are not translated! Without that, I should be able to finish it by tomorrow (at my timezone).

jmoenig:
don't worry about the build, your browser probably hangs on to cached versions of your translation. If you want to see your changes as you're making them you might need to clear your browser's cache every time, or - perhaps better yet - just open snap from your repository locally on your computer (i.e. not via a web server such as github pages).

Anyway, also don't worry if you don't find *all* the strings. We can always add more later. Just let me know when you're ready and I'll pull it. Early next week will be fine :-)

Thanks!

R4356th:
Alright, @jmoenig. I think I will locally test them with Git. Thank you for your feedback.

R4356th:
This should be (because sadly my browser is taking too long to clear all caches, I cannot check it) stable enough after just one more commit (I clicked the Ready for review button mistakenly).

R4356th:
Ready for review, @jmoenig!

locale/lang-bn.js
R4356th: I have removed this in 6d9e097.

jmoenig:
looks great, will pull tomorrow, thank you!

PR #2630, opened 2020-07-08
Fixing make-variables Lib showing and hident script vars

jguille2:
Hi!
Fixing make-variables lib, to show/hide also script variables.

It was more elegant, using only Snap! blocks, but now they use JS (similar to primitives code) to access to the scope of the custom block context.
Joan


jmoenig:
Thanks, Joan!

PR #2634, opened 2020-07-09
Update threads.js

brianharvey:
Fix foreach bug for hybrid lists from forum.


jmoenig:
thanks, Brian, this looks good!

PR #2640, opened 2020-07-13
fixing issues (constraining and switching) on PaintEditor

jguille2:
Hi Jens (@jmoenig),

This is about [that forum bug report about PaintEditors constrain feature](https://forum.snap.berkeley.edu/t/paint-editor-shift-does-not-constrain/2310/2).
This is tested, but I want to test a little more (because there are quite a few changes and a lot of actions to test (switching between editors, painting, saving, using tools, keyboard events...). Anyway, this PR helps us to share these changes and check if those ways are right...
A summary:
- To allow constraining with the _shift_ key, on "direct" _VectorPaintEditor_ we have to use `currentKey` and not `event.shiftKey`. It is the same behavior than _PaintEditor_.
 ```
-       this.shift = event.shiftKey;
-        this.ctrl = event.ctrlKey;
+        myself.shift = myself.world().currentKey === 16;
+        myself.ctrl = event.ctrlKey;
```
- To allow constraining with the _shift_ key after switching between editors (bitmap <-> vector) we have to call the "switcher" asynchronously, because the _confirm dialog_ is capturing the _keyboard listener_. So, we have to run _openIn_ functions after closing those dialogs.
on paint.js
```
            if (myself.paper.undoBuffer.length > 0) {
                myself.ide.confirm(
                    'This will erase your current drawing.\n' +
                    'Are you sure you want to continue?',
                    'Switch to vector editor?',
                    () => {
                        setTimeout(() => {myself.switchToVector()});
                    }
                );
            } else {
                myself.switchToVector();
            }
```
and on sketch.js
```
                if (myself.shapes.length > 0) {
                    myself.ide.confirm(
                        'This will convert your vector objects into\n' +
                        'bitmaps, and you will not be able to convert\n' +
                        'them back into vector drawings.\n' +
                        'Are you sure you want to continue?',
                        'Convert to bitmap?',
                        () => {
                            setTimeout(() => {myself.convertToBitmap()});
                        }
                    );
                } else {
                    myself.convertToBitmap();
                }
```

- To allow saving after switching editors several times (bitmap -> vector -> bitmap...) we have to add the "onsubmit action" to _convertToBitmap_ function. The reason is we are not creating a new costume (we are using the one converted from our vectorEditor) but maybe we are not editing a previous costume from our wardrobe.
The result of this is:
  + As current version, if we edit a svg costume, changing it to bitmap. Our editor will create a new costume (not deleting the previous svg costume). This is debatable.... but now (and in v5 too) is doing nothing... and it's the same occurs when we change a costume from bitmap to svg (this creates a new costume without deleting previous one).
  + Creating a new costume, and switching from vector to bitmap, saving is running (not now in current version).

- And the other changes are only typos and little bugs without errors (_this_ to _myself, functions parameters, changes to _arrow functions_ format...)

I want to test more (to avoid possible errors) and I will confirm here...

Joan

### DIFF
--- a/src/objects.js
+++ b/src/objects.js
@@ -9973,7 +9973,8 @@ SVG_Costume.prototype.edit = function (
     oncancel,
     onsubmit
 ) {
-    var editor = new VectorPaintEditorMorph();
+    var editor = new VectorPaintEditorMorph(),
+        myself = this;
 
     editor.oncancel = oncancel || nop;
     editor.openIn(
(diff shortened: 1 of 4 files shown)


jmoenig:
That's great, Joan. These changes all make sense to me and I'm ready to pull them when you tell me to :-)
Thanks you!

jmoenig:
Naw, @jguille2 , this looks fine, I'm pulling it right now! Thank you so much :-)

PR #2642, opened 2020-07-15
Update bignumbers.xml

brianharvey:
Fix path to libraries (no more snapsource)


jmoenig:
excellent, thank you!

PR #2647, opened 2020-07-17
Add favicon specification for iOS

### LINKED ISSUE #2646: iOS requires "apple-touch-icon" for favicon
iOS requires the "apple-touch-icon" type to be inside link tag for favicon. You can read about this [here](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4).

R4356th:
### Resolves

Resolves #2646.

### Changes

Adds favicon specification for iOS.

### Cause of Changes

Read #2646 and https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4.


jguille2:
Hi @jmoenig 
I have no apple devices to test, but I think this is not right.
- _apple-touch-icon_ is not for favicons, is for the _home screen app icon_. Maybe our _Snap! logo_ is better there that the _lambda favicon_.  I used it in my PWA tests.
- Anyway, their src has to be a png file. And it should have a bigger size (not 32x32px like our favicon). Usually 180x180px. Doing this in the PWA style (with a manifest) you can add different sizes for different devices (iphones, ipads...)

I can propose alternatives... but I think it's better to talk (someday) about a pwa manifest that does this job with more possibilities.

Joan

R4356th:
Hi, @jguille2. I think I know an alternative. I will try that but I do not have an iOS device, too.

R4356th:
In the meantime, @jmoenig, every OS lets you have pngs as the favicon, to my knowledge. So, you could just use one single png file for every OS.

jmoenig:
Thanks, @jguille2 !

R4356th:
> 
>     * _apple-touch-icon_ is not for favicons, is for the _home screen app icon_. Maybe our _Snap! logo_ is better there that the _lambda favicon_.  I used it in my PWA tests.
> 
>     * Anyway, their src has to be a png file. And it should have a bigger size (not 32x32px like our favicon). Usually 180x180px. Doing this in the PWA style (with a manifest) you can add different sizes for different devices (iphones, ipads...)
> 
@jguille2, Apple Developer Documentation says, "To specify an icon for a single webpage or replace the website icon with a webpage-specific icon, add a link element to the webpage, as in:

<code>link rel="apple-touch-icon" href="/custom_icon.png"</code>"

jguille2:
Hi @R4356th ,

But 'this icon' is not the 'favicon browser icon', is the icon for the home screen page.

And there are a bunch of _meta_ tags and possibilities. And I think it has no sense to consider element by element. If is needed, the better way is to consider the "global pack" (if we need a set of icons with different sizes, or a manifest, to check different browsers and devices compatibility...)
Joan

R4356th:
> 
> And there are a bunch of _meta_ tags and possibilities. And I think it has no sense to consider element by element. If is needed, the better way is to consider the "global pack" (if we need a set of icons with different sizes, or a manifest, to check different browsers and devices compatibility...)
> 
Are you talking about <code>meta name="og:image" content="src/favicon.ico"/</code>?

PR #2648, opened 2020-07-17
Add viewport

R4356th:
### Changes

Adds viewport.

### Cause of Changes

This single line should be able to make mobile and tablet experience better. You can read how that works [here](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag).

### Extra Note

GitHub is showing the diff incorrectly; the indentation is alright.


jmoenig:
@bromagosa this is for you to review ;-)

snap.html
```diff
@@ -2,6 +2,7 @@
 <html>
     <head>
         <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
```
towerofnix, line 5: It's not a huge deal, but (almost) the rest of this file uses spaces and this is a tab, so maybe you could replace the tab with eight manually-typed spaces (or switch your editor to use spaces instead of tabs)?

R4356th: I do not think this is an issue because it looks alright when you view the file.

R4356th: Never mind. I just fixed this while resolving the conflict.

bromagosa:
@jmoenig this looks good to me. Just tested it in Android and it does make it significantly more usable. Prevents the page from automatically zooming all over the place :)

A phone is still not the ideal environment for Snap!, but an Android tablet is actually not bad at all!

We should find a way to automatically hide the virtual keyboard after the user's finished typing, but that's a different topic.

TLDR; this is a simple change that significantly improves usability in Android. Should be pulled :)

jmoenig:
Thanks for the test and review, @bromagosa, and thanks, @R4356th for the PR. Will pull for the next patch.

PR #2663, opened 2020-07-26
More cleanup and help texts

brianharvey:
Merge multi-line into strings, merge bigger data into frequency thingo.


jmoenig:
yay, thanks!

PR #2671, opened 2020-08-04
Update strings.xml

brianharvey:
Adding TEXT_CONTAINS_ and TO LOWER CASE to strings library.


jmoenig:
did you remember to take the `lowercase` block from the Frequency Distribution Analysis library? Because that version has already been translated to Portuguese...

brianharvey:
No, but I copied over the translation.  I use toLocaleLowerCase() which is supposed to work better in Turkish and mostly otherwise be the same.

jmoenig:
Thanks! Will pull for the next patch

PR #2672, opened 2020-08-05
fixed green flag size in embed mode

bromagosa:
It now looks like it did before v6:

imatge

This is how it looks before the fix:

imatge


jmoenig:
Oh, nice, thanks, Bernat!

PR #2675, opened 2020-08-06
find broadcast senders and receivers from block context menu

bromagosa:
Inspired by what Scratch 2.0 used to have, and of course by "senders / implementors" in The Language.

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -3093,12 +3093,13 @@ BlockMorph.prototype.userMenu = function () {
         return menu;
     }
     if (contains(
-        ['doBroadcast', 'doBroadcastAndWait', 'receiveMessage'],
+        ['doBroadcast', 'doSend', 'doBroadcastAndWait', 'receiveMessage',
+            'receiveOnClone', 'receiveGo'],
         this.selector
     )) {
         menu.addLine();
         menu.addItem(
-            (this.selector === 'receiveMessage' ?
+            (this.selector.indexOf('receive') === 0 ?
                 "show senders" : "show receivers"),
             'showMessageUsers'
         );
@@ -3129,15 +3130,31 @@ BlockMorph.prototype.userMenu = function () {
 BlockMorph.prototype.showMessageUsers = function () {
     var ide = this.parentThatIsA(IDE_Morph),
         corral = ide.corral,
-        getter = (this.selector === 'receiveMessage') ?
-            'allSendersOf' : 'allHatBlocksFor';
-        message = this.inputs()[0].evaluate();
+        getter = (this.selector.indexOf('receive') === 0) ?
+            'allSendersOf' : 'allHatBlocksFor',
+        inputs = this.inputs(),
+        message, receiverName;
+
+    if (this.selector === 'receiveGo') {
+        message = '__shout__go__';
+    } else if (this.selector === 'receiveOnClone') {
+        message = '__clone__init__';
+    } else if (inputs[0] instanceof InputSlotMorph) {
+        message = inputs[0].evaluate();
+    }
+
+    if (((this.selector === 'doSend') && inputs[1] instanceof InputSlotMorph)) {
+        receiverName = this.inputs()[1].evaluate();
+    } else if (this.selector.indexOf('receive') === 0) {
+        receiverName = this.scriptTarget().name;
+    }
 
     if (message !== '') {
         corral.frame.contents.children.concat(corral.stageIcon).forEach(
             icon => {
                 if (icon.object &&
-                    (icon.object[getter](message).length > 0)
+                    ((this.selector !== 'doSend' || receiverName === icon.object.name) &&
+                    (icon.object[getter](message, receiverName).length > 0))
                 ) {
                     icon.flash();
                 }
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -3092,6 +3092,17 @@ BlockMorph.prototype.userMenu = function () {
         }
         return menu;
     }
+    if (contains(
+        ['doBroadcast', 'doBroadcastAndWait', 'receiveMessage'],
+        this.selector
+    )) {
+        menu.addLine();
+        menu.addItem(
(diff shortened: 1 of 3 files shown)


src/blocks.js
```diff
@@ -3092,6 +3092,17 @@ BlockMorph.prototype.userMenu = function () {
         }
         return menu;
     }
+    if (contains(
+        ['doBroadcast', 'doBroadcastAndWait', 'receiveMessage'],
```
jmoenig, line 3096: what about 'doSend', shouldn't that also be part of this list?

src/objects.js
```diff
@@ -5788,6 +5788,24 @@ SpriteMorph.prototype.allMessageNames = function () {
     return msgs;
 };
 
+SpriteMorph.prototype.allSendersOf = function (message) {
+    if (typeof message === 'number') {message = message.toString(); }
+    return this.scripts.children.filter(morph => {
+        var event;
+        if (morph.selector) {
+            if (
+                contains(['doBroadcast', 'doBroadcastAndWait'], morph.selector)
```
jmoenig, line 5797: also here, I think it's probably a good idea to also include the new 'doSend' selector

src/gui.js
```diff
@@ -8502,6 +8502,35 @@ SpriteIconMorph.prototype.copySound = function (sound) {
     this.object.addSound(dup.audio, dup.name);
 };
 
+// SpriteIconMorph flashing
+
+SpriteIconMorph.prototype.flash = function () {
+    var previousOutlineColor = this.outlineColor.copy(),
```
jmoenig, line 8508: hmm... do we really need to copy the color? Can't we just remember it, or is it mutated anywhere in between here and switching back?

src/gui.js
```diff
@@ -8502,6 +8502,35 @@ SpriteIconMorph.prototype.copySound = function (sound) {
...
+        previousOutline = this.outline,
+        previousState = this.userState;
+
+    this.outlineColor = SpriteMorph.prototype.highlightColor;
+    this.outline = 2;
+    this.userState = 'pressed';
+
+    this.rerender();
+
+    world.animations.push(new Animation(
```
jmoenig, line 8518: I'd be careful about directly referencing the "world" variable, just in case there's a page with several worlds on it at some point in the future. How about instead assigning a local var "world"?

src/gui.js
```diff
@@ -8502,6 +8502,35 @@ SpriteIconMorph.prototype.copySound = function (sound) {
     this.object.addSound(dup.audio, dup.name);
 };
 
+// SpriteIconMorph flashing
+
+SpriteIconMorph.prototype.flash = function () {
```
jmoenig, line 8507: Nice! I like how this is modeled after the SpriteMorph>>flash method

src/objects.js
```diff
@@ -5788,6 +5788,24 @@ SpriteMorph.prototype.allMessageNames = function () {
...
+            ) {
+                event = morph.inputs()[0].evaluate();
```
jmoenig, line 5799: Not sure if *evaluating* the first input always works, e.g. if it's an embedded reporter. In this case something might go awry. Probably best to check whether the first input is an InputSlotMorph, before evaluating it.

src/blocks.js
```diff
@@ -3115,6 +3126,26 @@ BlockMorph.prototype.userMenu = function () {
     return menu;
 };
 
+BlockMorph.prototype.showMessageUsers = function () {
+    var ide = this.parentThatIsA(IDE_Morph),
+        corral = ide.corral,
+        getter = (this.selector === 'receiveMessage') ?
```
jmoenig, line 3132: Oooh, now you're being fancy! Love it, though :-)

src/blocks.js
jmoenig, line 3096: Also, we should probably check for fancy senders of '__shout__go__' and '__clone__init__', right? So we might want to include the green-flag and "When-I-start-as-clone" hats also.

bromagosa: Totally. Adding it now.

src/objects.js
bromagosa: Also adding it now.

src/gui.js
bromagosa: We don't need to copy it indeed. I don't know why I thought we di.

src/gui.js
bromagosa: Oh yes. I actually forgot to assign it, but that was my intention :)

src/objects.js
bromagosa: Thanks, done too!

src/blocks.js
```diff
@@ -3093,12 +3093,13 @@ BlockMorph.prototype.userMenu = function () {
...
     if (contains(
-        ['doBroadcast', 'doBroadcastAndWait', 'receiveMessage'],
+        ['doBroadcast', 'doSend', 'doBroadcastAndWait', 'receiveMessage',
+            'receiveOnClone', 'receiveGo'],
         this.selector
     )) {
         menu.addLine();
         menu.addItem(
-            (this.selector === 'receiveMessage' ?
+            (this.selector.indexOf('receive') === 0 ?
```
jmoenig, line 3102: Haha! I guess we need to remember this naming convention now :-)

src/blocks.js
```diff
@@ -3129,15 +3130,31 @@ BlockMorph.prototype.userMenu = function () {
...
+    } else if (this.selector === 'receiveOnClone') {
+        message = '__clone__init__';
+    } else if (inputs[0] instanceof InputSlotMorph) {
+        message = inputs[0].evaluate();
+    }
+
+    if (((this.selector === 'doSend') && inputs[1] instanceof InputSlotMorph)) {
+        receiverName = this.inputs()[1].evaluate();
+    } else if (this.selector.indexOf('receive') === 0) {
+        receiverName = this.scriptTarget().name;
```
jmoenig, line 3149: clever!

PR #2678, opened 2020-08-13
Fix minor issue with remaining time

brollb:
I was assuming that you meant "is enabled for three days" to be referring to the remaining time rather than the total time and then noticed that it will read a little funny if there is only one "days" left :)


bromagosa:
This makes sense to me :+1:

jmoenig:
thanks, @brollb and @bromagosa ! Now that Bernat has blessed this I'm gonna pull it :-)

PR #2681, opened 2020-08-21
Update parallel_module.xml

brianharvey:
Fix bug about empty input slot reported by ego-lay_atmanbay in forum.


jmoenig:
hmm... I'm not convinced about the added type-checks. The only benefit they add is supporting empty C-slots. On the flip side they introduce more complexity. I think I'd rather stick to the current version.

PR #2684, opened 2020-08-31
Updating Catalan translation to current 6.2.0dev

jguille2:
Hi Jens!
Just updating Catalan translation to current state (new Libraries menu and "receivers and senders").
Joan


jmoenig:
Thank you so much, @jguille2!

PR #2690, opened 2020-09-14
Start at 1 in updateDefinition (similar to duplication)

brollb:


jmoenig:
good catch, thank you @brollb !

brollb:
No problem - I am always happy to contribute back! (Even if they can sometimes be pretty small changes :) )

jmoenig:
I love little improvements such as this one!

PR #2691, opened 2020-09-18
fixes Load failed: TypeError: Cannot read property ‘width’ of null

bromagosa:
Just checking whether `!src` before accessing any of its properties. The guard was already there, but just two lines below where we first accessed `src`.


jmoenig:
Excellent, thank you Bernat!

PR #2693, opened 2020-09-19
Update localstorage_module.xml

brianharvey:
Add reporter that finds a specified key.


jmoenig:
No, I specifically don't want that. We have `find` as primitive.

jmoenig:
Hmm... I take it back. Maybe we should have that block after all.

PR #2696, opened 2020-09-24
Add "add comment" option to the block context menu

robertf224:
block-add-comment

Just one data point but I think I literally didn't realize you could leave comments for the duration of my TEALS class last year.  I think there's an argument for converting the script area context menu into some sort of toolbar so its actions are more discoverable, but in the meantime I think we can at least make commenting more discoverable by adding it as an option on the context menu for blocks (which I open all the time) vs. just an option on context menu for the script area (which I never really think to open).  Hopefully will lead students to leave more comments 😄 

Couldn't figure out a way to plug into the undo/redo history than just by simulating a drag-drop, happy to change if there's a better way.  Tested out this method and it has the same behavior as adding a comment through the script area context menu
block-add-comment-undo-redo

Also couldn't make sense of the talk of "changesets" in the contributing guidelines and didn't see other PRs doing anything like that but happy to submit in a different format if necessary.

Rob

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -3024,6 +3024,35 @@ BlockMorph.prototype.userMenu = function () {
         "delete",
         'userDestroy'
     );
+    if (isNil(this.comment)) {
+        menu.addItem(
+            "add comment",
+            () => {
+                var comment = new CommentMorph();
+                this.comment = comment;
+                comment.block = this;
+                comment.layoutChanged();
+
+                // Simulate drag/drop for better undo/redo behavior
+                var scripts = this.parentThatIsA(ScriptsMorph),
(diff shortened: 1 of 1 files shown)


jmoenig:
Hi @robertf224 , good idea and great observation!
thanks for the PR. I'm going to pull it now and check it out. I like your solution for adding the comment to undrop/redrop stack. There are probably a bunch of special cases we need to cover, such as to disable comments for certain blocks (e.g. in the palette) and to make sure this feature also supports custom blocks. But I'll look into that myself.

jmoenig:
Okay, just glanced over your code and did some quick checks, you seem to have thought about everything. This is GREAT, thank you for a *perfect* PR!

PR #2699, opened 2020-10-02
Improve Greek Translation

HM100:
This PR will translate various strings on Snap! into greek, including some of the blocks and menus that weren't been in Greek for a long time


jmoenig:
Awesome, thank you very much! Would you also like to add your name to the translators list?

HM100:
Yes! You can add me as HM100 as I barely expose my real name inside the internet, unless you need it

jmoenig:
Great, will do. Thanks!

PR #2714, opened 2020-11-15
Fixing (bitmap)costume fitting

jguille2:
Hi!
## Issue:
- Bitmap costumes larger than stage are fitted to the stage dimensions.
- This is running Ok (the image is fine), but _costume extent_ info is wrong. It shows stage extents. Then width or height is wrong (depending on the fitting).
- And this causes different problems."width/height" of costume reporter is wrong,  and also the "pixel of costume" is not fine.

## Cause:
- `shrinkToFit` function uses `thumbnail` execution
- `thumbnail` returns a fine fitted image, but in a Canvas with the dimensions of their container (where the costume is fitted)
- This is fine for `thumbnail` because images are centered (for example inside the Wardrobe (`CostumeIconMorph`) or for `SpeechBubbleMorph`)

## Proposal
This PR writes its own code for `shrinkToFit` to scale (fit) images in a proportional canvas.

Note: SVGCostumes are not fitted. I'll submit a new PR with a proposal to do it (to get the same behavior)

Thanks,
Joan

### DIFF
--- a/src/objects.js
+++ b/src/objects.js
@@ -9879,8 +9879,27 @@ Costume.prototype.editRotationPointOnly = function (aWorld) {
 // Costume thumbnail
 
 Costume.prototype.shrinkToFit = function (extentPoint) {
+    var scale,
+        fittedExtentPoint,
+        fittedCanvas,
+        ctx;
     if (extentPoint.x < this.width() || (extentPoint.y < this.height())) {
-        this.contents = this.thumbnail(extentPoint);
+        scale = Math.min(
+            (extentPoint.x / this.width()),
+            (extentPoint.y / this.height())
+        );
+        fittedExtentPoint = new Point(this.width() * scale, this.height() * scale);
+        fittedCanvas = newCanvas(fittedExtentPoint, true);
+        ctx = fittedCanvas.getContext('2d');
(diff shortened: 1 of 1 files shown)


jmoenig:
Thank, Joan! Great analysis and solution proposal. Funny, just this past Friday @jadga-h was pointing out this very issue to me, and we discussed whether we should prioritize fixing it. And here you are with a solution. Great minds... :-)

Will pull in the next few days for next week's patch.

jguille2:
Fine!!
Last night I've worked on other proposal to extend the same "fitting" to svgCostumes. It's a little more _tricky_ because I need to decode-change-encode the "svg-string" to implement it. But I think it's fine, and also it can help us with a lot of _svg_s that are not running ok in Snap! because it has no width/height info in their original data.

That code is just finished (and it works!), but then I see another _issue_ related with the costume "refresh" after some manipulation. It is a separate issue, but it's weird if we decide to do that fitting. I open another issue to see this.

Joan

jmoenig:
Excellent! Sounds great that you're looking into SVGs as well, we are, indeed, encountering a lot of issues with them. Thank you so much!

PR #2716, opened 2020-11-17
Updating Spanish translation

jguille2:
Hi Jens,

Tomorrow I have the Snap! plugin presentation in MoodleMootSpain 2020.

It will be a Snap! presentation for the Spanish (Spain and America) moodler community. So I've updated the Snap! Spanish translation. Now, all the blocks (and dropddown options) are translated. Pending to check menu options and help messages (oh, but I need time to prepare the presentation!! :) )

There is no rush here. I'll use a local distro.

Thanks,
Joan


jmoenig:
Thanks, Joan! I will pull this later today, after I'm done with a few other things I'm working on right now. Then we can either go ahead and release the current changes right away or deploy them to dev.

PR #2722, opened 2020-11-22
Updating Greek Translation

HM100:
Blocks translation should be almost done


jmoenig:
Excellent, thank you very much!

PR #2725, opened 2020-11-28
Fixing svg loading

jguille2:
Hi!

This can be the first step to improve svg management (next steps would be "fitting large costumes" and a "svg stretching" feature).

## Current issue
- Many _svg_ images are badly uploaded in Snap! The reason is they (I see this in many free svg images) have not _width_ and _height_ attributes. Image viewers and browsers load and show them properly, with a default size, but our management of the HTMLImageElement (for the SVG_Costumes) has problems.
- In Firefox, this results in "empty" costumes loaded.
- In Chrome this seems to go better (not empty) but the result is not fine. We get cropped images (depends on the proportions), with smaller thumbnails and, if we play with the "size" property,  we will see a totally wrong behavior.

## Testing this issue

[svgTests.zip](https://github.com/jmoenig/Snap/files/5609967/svgTests.zip)

Two svg images to test.

- The original one (it has not _width_ and _height_ attributes)
- The sized sample (is the original, adding these attributes with a text editor)

## This PR proposal
- It checks svg images (from importing or dropping svg files).
- _viewBox_ attribute is mandatory. Without it, the image is not loaded (no errors)
- If _width_ and _height_ attributes are missing, it writes them, using the "viewBox" values.
- Yes, if "viewBox" has large values, that image will have a large size. But if this PR is accepted, "fitting behavior" will be added (like bitmapCostumes does) later.
- Note that this _checker_ is made in two functions with an asynchronous process (an _onload_ event). This is required to draw the costume thumbnail in that creation moment (it would be updated later without problems, but not in the first moment). The reason is the _drawImage_ function (to draw that thumbnail) needs a _loaded_ ImageElement. And if is not completely loaded, it draws an empty image.

Joan

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -2139,7 +2139,39 @@ IDE_Morph.prototype.droppedImage = function (aCanvas, name) {
 };
 
 IDE_Morph.prototype.droppedSVG = function (anImage, name) {
-    var costume = new SVG_Costume(anImage, name.split('.')[0]);
+    var myself,
+        viewBox, w, h,
+        svgNormalized,
+        headerLenght = anImage.src.search('base64') + 7, // usually 26 from "data:image/svg+xml;base64,"
+        svgStrEncoded = anImage.src.substring(headerLenght),
+        svgObj = new DOMParser().parseFromString(atob(svgStrEncoded), "image/svg+xml").firstElementChild;
+    name = name.split('.')[0];
+
+    // check svg content and size it if necessary before loading
+
+    if (svgObj.attributes.getNamedItem("viewBox")) { // viewBox attribute is mandatory
+        // width and height are required
+        if (!svgObj.attributes.getNamedItem("width") ||
+                !svgObj.attributes.getNamedItem("height")) {
+            viewBox = svgObj.attributes.getNamedItem('viewBox').value;
+            w = Math.ceil(viewBox.split(' ')[2]);
+            h = Math.ceil(viewBox.split(' ')[3]);
+            svgNormalized = new Image(w, h);
+            svgObj.setAttribute('width', w);
+            svgObj.setAttribute('height', h);
+            svgNormalized.src = 'data:image/svg+xml;base64,' +
+                btoa(new XMLSerializer().serializeToString(svgObj));
+            myself = this;
+            svgNormalized.onload = function () { myself.loadSVG(svgNormalized, name); }
+        } else {
+            this.loadSVG(anImage, name);
+        }
+    }
+};
+
+IDE_Morph.prototype.loadSVG = function (anImage, name) {
+    var costume = new SVG_Costume(anImage, name);
+
     this.currentSprite.addCostume(costume);
     this.currentSprite.wearCostume(costume);
     this.spriteBar.tabBar.tabTo('costumes');
--- a/src/gui.js
+++ b/src/gui.js
@@ -2145,27 +2145,34 @@ IDE_Morph.prototype.droppedSVG = function (anImage, name) {
         headerLenght = anImage.src.search('base64') + 7, // usually 26 from "data:image/svg+xml;base64,"
         svgStrEncoded = anImage.src.substring(headerLenght),
         svgObj = new DOMParser().parseFromString(atob(svgStrEncoded), "image/svg+xml").firstElementChild;
+
     name = name.split('.')[0];
 
-    // check svg content and size it if necessary before loading
+    // check svg 'width' and 'height' attributes and set them if needed
 
-    if (svgObj.attributes.getNamedItem("viewBox")) { // viewBox attribute is mandatory
-        // width and height are required
-        if (!svgObj.attributes.getNamedItem("width") ||
-                !svgObj.attributes.getNamedItem("height")) {
+    if (svgObj.attributes.getNamedItem("width") &&
+            svgObj.attributes.getNamedItem("height")) {
+        this.loadSVG(anImage, name);
+    } else {
+        // setting HTMLImageElement default values
+        w = '300';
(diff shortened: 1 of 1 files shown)


brianharvey:
I have a couple of questions.

1.  Is there a reason for not giving an error message for a malformed vector image?  All else being equal, silent errors are confusing.

2. I don't know anything about SVG format; is the lack of a view box something that will happen in practice?  If the file has width and height but no view box, could the latter be derived from the former?

jguille2:
Hi Brian,

First of all. I didn't know anything about SVGs :) I've been working and I think I've took care of everything (or at least I've tried)... but I’m sure more glances will help us not forget anything.

- About error messages. The result (nothing) is the same if I drag and drop other not supported files. For example, if I drag and drop a _pdf_ file, nothing happens.
But if you want, it's really easy to implement that "error message".
Tell me if you want, and I'll add this single line.

- About _viewBox_. No, its value can't be derived from _width_ and _height_. These values (width and height) are about the visualization (the "out side", related to where is being drawn) and _viewBox_ is about the "inner" coordinate system and about which elements will be displayed. So, to know something about  the viewBox possible values, we need to see all the internal elements.
So, I think it must be required.

Joan

brianharvey:
Well, if it were me, I'd give an error message for PDFs too ("I don't know how to import a ____"), and for a bad SVG "Malformed SVG file."  But I guess it's the way it is because Jens wrote it following the Scratch philosophy of never giving an error message, so I'll defer to his decision on this.

jguille2:
Ok,

I wait your decisions. If finally the "error message" is added, maybe that message could be "Unsupported SVG format", because I don't have enough knowledge to say "malformed". Maybe that file is right for other software, and I can't see a "mandatory" viewBox attribute in the SVG spec.

About other comments, maybe your generic proposal (or maybe something like "This file can't be imported") would be fine. The main reason for this is to avoid wasting time in "multiple tests", because people don't know if there is something wrong or what's happen...

jmoenig:
This looks great, Joan, thank you!
Quick question, if the viewbox attribute is missing, can't we just add it to the svg before loading it? Wouldn't that take care of the error message problem?

jguille2:
Hi!
I want to make two changes here... I wait for your opinions...

1. ViewBox format

All _svg_ files that I've seen have the _viewBox_ values separated by _whitespace_. But I see _specs_ allow _comma separated values_. So I have to change my "reader". I would do it with regExp (to do it shorter)... but maybe you don't like it. Tell me what you want and I will update this.
With regExp it would be changing this:
```javascript
Math.ceil(viewBox.split(' ')
```
for this code:
```javascript
Math.ceil(viewBox.split(/[ ,]/).filter(item => item)
```
Just tell me if you agree with this code or you like a larger form.

2. Make viewBox not mandatory

I've been watching a lot of free _svg_ images and testing...

What I see:
  - "Unsized" svg files (without _width_ and _height_) are "a disaster". Firefox makes it worse (no image). And Chrome seems better (you get an image, sometimes is cropped) but thumbnail and resizing action are really bad.
  - _ViewBox_ cannot derived from a size, because the inner coordinates (and not only its size)
  - But "sized" images without _viewBox_ are fine. If their missing _viewBox_ had other info, their behavior will be different... but it is still a "consistent" behavior (thumbails, resizing...)
  - When images has no "size" and no "viewBox", following htmlElements rules, they are assigned to a "300x150" size. It can be a little disaster (cropped images or smaller images with larger bounds). But we can get an image (in Chrome) ... although it will have the same problems related to "unsized" images.

And then, my "new" proposal:
  - If the image is not sized, we have to size it!
    - If there is a _viewBox_, we use its _width_ and _height_
    - If there is not a _viewBox_, we use `width=300` and `height=150`
  - And if the image is sized, we do nothing else, just load this image (although there was no _viewBox_)

Why? With this change, we keep only the improvements without having any loss. Firefox problems will disappear. And the inconsistencies with the thumbnails and resizing will also disappear. Of course, we won't fix some "broken" images... but we cannot invent what is not there.

That's all!
Tell me what do you think and I'll update this PR.

Joan

brianharvey:
Sounds great!  Thanks for all the detailed research, Joan.

jmoenig:
Yeah, this sounds alright to me, Joan.

Although, the more I think about this the more I'm wondering whether we're following the right approach here. We shouldn't be worrying about about splitting attribute strings at all, and we really shouldn't have to putz around with parsing at the XML-level, this feels so unnecessarily meta. Can't we simply take the SVG source and let the browser create an SVGElement, and *then* we simply interact with *that* programmatically? 

https://stackoverflow.com/questions/10085123/manipulate-svg-viewbox-with-javascript-no-libraries

Then we also wouldn't have to worry about reading the SVG *again* after we've injected some XML nodes? See, when I said I'd like to talk to SVGs as objects I wasn't talking about XML nodes, but about SVGs as DOM elements.

Does that make any sense?

jguille2:
Hi Jens,

I made the changes that we have commented.
I think now is fine, improving current behavior for all cases, and without losses.

About the possibility of changing our current implementation, using directly a SVGElement...I'm not sure. Maybe I have to make a prototype and evaluate pros and cons. My first thought is that maybe it is not a good idea. Why? Using SVGElements would improve svg manipulation: creation (fitting) and stretching basically (few operations) and I think this would worsen the other (much more common) operations: loading and saving projects, because I think we would have to encode-decode those SVGElements.
Maybe I'm wrong... It is my first thought.

Joan

jmoenig:
Great, thank you, Joan! Let's pull this and play with it. You may be right about avoiding actual SVG elements, as they might also contain JS themselves, and thus get rejected by browsers.

PR #2737, opened 2020-12-05
Added Area Library 

Barthdry:
A library that is capable of doing fast calculation of general areas of Square, Rectangle, Triangle, Trapezoid, Circle and Oval with pi in higher accuracy state


brianharvey:
I'm sorry, but I don't think you're ready to contribute libraries.  There are some things you have to learn first.

To start with the two obvious ones:

1.  A _predicate_ (hexagonal shape block) is a function that always reports True or False.  Your blocks report numbers, so they should be reporter (oval) shaped.

2.  You say you use a "higher accuracy" value of pi, but it's only six digits.  Standard computer floating point gives you 15 or 16 digits of precision, so the reported value of pi should be 3.141592653589793.  It's possible to get even more precision but it takes a lot of extra effort and this is good enough for most purposes.

But a more subtle problem is that even if you fixed the above bugs, this just isn't so hard for a user to write as to justify a library.  We have libraries for things that are really complicated, things that can't be done without using JavaScript (note, that's not the same as "things that can be done using JavaScript"!), or things that you have to know a lot about arcane details of Snap<i>!</i> to be able to write.  Your blocks are things that any user should be able to write easily and quickly if they know the geometry.

Learn more, then think about contributing a library.  Thanks.

Barthdry:
Ok i'll try

PR #2739, opened 2020-12-06
Fitting new large svgCostumes to the stage dimensions

jguille2:
Hi Jens,

This PR
- Add to the SVG loader a "viewBox" extra checking, to set it was missing with the "new sized" or default values. This is required to fit the images (withour any rasterizing) or for a future (soon) svg stretching.
- Fit large svg images to the stage dimensions.

Note: I know 'Costume' and 'SVG_Costume' objects have the 'shrinkToFit' method defined. It may seem better to implement this fitting there. But it is only needed for the (new) image loading, and it would be "heavy" to do it in an extra step (decoding, changing and coding again... a bit heavy for large (memory) images).

Tell me if it is not right or if you think this must be implemented into the 'shrinkToFit' method.

Thanks,

Joan

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -2140,44 +2140,71 @@ IDE_Morph.prototype.droppedImage = function (aCanvas, name) {
 
 IDE_Morph.prototype.droppedSVG = function (anImage, name) {
     var myself,
-        viewBox, w, h,
+        viewBox,
+        w = 300, h = 150, // setting HTMLImageElement default values
+        scale = 1,
         svgNormalized,
         headerLenght = anImage.src.search('base64') + 7,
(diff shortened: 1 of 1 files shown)


jmoenig:
thank you, Joan, this all makes sense to me and sounds very good. I like your idea to change the SVG when we first import it, because that's the entry point to everything else. I'm not even sure whether, if and when we want to introduce special cases for stretching SVG_Costumes at all. Once we start going down that avenue we might just open Pandora's box about SVGs in general. I kinda like our current approach to keep the SVGs around and rasterize them just-in-time for most operations. I'll pull this later, today, thank you!

PR #2740, opened 2020-12-06
Added Hindi Language Translation and Tamil. With Area Library Improved 

Barthdry:
I had updated the tamil language translation. Added Hindi Translation Beta. Also improved Area library and added more function.


jguille2:
Hi Barthdry,

My suggestion is to separate your proposals in different PRs, to allow their discussion (if needed) and to enable different acceptances (if they can be).

Extra notes:
- I see you change the "Stage" word. Have you commented this with vinayakumar R? He did another translation (and a lot of strings more) and maybe you can talk to him about this.
And, other comment...Tamil file has a lot of German translations. I know you didn't this. But I wonder, is it better for Tamil people? If not, it is better to delete all the no-translated strings, to see them in English.

- The same for your new Hindi translation. Why Estonian? You have about 25 translated strings in a file of ~1500 lines. It has no sense... and I guess Hindi people know more English than Estonian, don't they?

- About your library... I see you have changed some problems that Brian commented in your other PR. But note his last comment
>But a more subtle problem is that even if you fixed the above bugs, this just isn't so hard for a user to write as to justify a library. We have libraries for things that are really complicated, things that can't be done without using JavaScript (note, that's not the same as "things that can be done using JavaScript"!), or things that you have to know a lot about arcane details of Snap! to be able to write. Your blocks are things that any user should be able to write easily and quickly if they know the geometry.

- And, for the next (separated) PR, take care of not push extra changes (like comments in other files).

I'm sorry for all my critical comments... but we all will find the ways to make Snap! better. Thanks!

Joan

Barthdry:
Yes i will improvise them

PR #2741, opened 2020-12-07
Added Javascript and Area Library. Also improved the Tamil Translation

Barthdry:
Javascript library has alert, prompt and site open. Also added area library. Improved tamil translation highly


jmoenig:
Thanks for the improvements of the Tamil translations, that's really great!

Can you please decouple the PR for the translation from the contribution of the two libraries? In general we're trying to not introduce new libraries unless they address a very specific curricular purpose. So it would help us decide if you made a seperate PR for each of them with some explanations what's in them.

Thank you!

PR #2742, opened 2020-12-07
compiled multimap

brianharvey:
This PR makes the compiled version of MAP variadic.  The expected use case is that the function input's ring does not have formal parameters, and  corresponding items of the input lists fill the empty slots in the ringed expression.

For compatibility, if the ring is given formal parameters, then the first one ("value") is bound to a list of the corresponding items of the input lists.

Google Chrome001

PS I'm hoping this will speed up the APL library quite a lot.


jmoenig:
Looks cool to me, thanks! Pulling now...

PR #2744, opened 2020-12-07
Added Tamil Translation

Barthdry:
Most words that remain in english has been cleared and made into tamil. The Language translation was improved.

### DIFF
--- a/locale/lang-ta.js
+++ b/locale/lang-ta.js
@@ -185,14 +185,14 @@ SnapTranslator.dict.ta = {
     'translator_e-mail':
         'vnkmr7620@gmail.com', // optional
     'last_changed':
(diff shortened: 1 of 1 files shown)


jmoenig:
hmm... the current Tamil translation is indeed very incomplete, but so is this one. There are quite a few "mixed" strings that don't look like they are complete or intentional. Is this still a work in progress or do you actually want us to pull this now? Thank you!

Barthdry:
You can pull this

Barthdry:
Yes you can merge and the thing is that maybe there is no word for absolute and some maths thing.

jmoenig:
Okay, I'm going to pull this one and reject the other one. Thank you!

PR #2746, opened 2020-12-10
New Library: Javascript Library 

Barthdry:
I have added a simple library of javascript for those who are children, teen or non devoloper who uses snap. If they are not able to use the default javascript function. They will be able to use this library to exeucute simple javascript.
**Issues Solved:-**
Made Non Js Coder and non coder use simple js


jmoenig:
Thanks, but we're not interested in teaching JavaScript.

PR #2759, opened 2020-12-30
Update list-utilities.xml

brianharvey:
Speedup via JS primitives and via compiling hofs.  New notation: ☠︎in front of title means this block is for internal use, not for users.


jmoenig:
Thanks! Instead of - or in addition to - the pirate-skull symbol on the "private" helper blocks how would you feel about adding another option to the custom block definition offering to "hide in palette"?

brianharvey:
Sure, that's fine.

brianharvey:
So, will that be in 6.5.1?

jmoenig:
I think we can add the option to hide a custom block in the palette quickly for the next release. Since it would be a new feature we'd probably give it a new minor release number, 6.6.0. Alternatively we could make it a "hidden" shift-right-click feature first, I dunno. It seems like being able to hide helper functions might just be something we end up using a lot in many places and other libraries as well.

OTOH this might be opening pandora's box to all kinds of "hiding blocks" out of other motives, such as obfuscating how something works. And we might even want to think some more about this, e.g. whether and how we might wish to include a visual indication to such "helper blocks", because not seeing them in the palette might be pretty confusing to readers trying to understand a project. I'm also a little bit afraid that once this feature gets out it becomes all the rage and everybody starts using it with their JS function :-)

PR #2765, opened 2021-01-11
Updated 170 keywords in Tamil Language

### LINKED ISSUE #953: tamil language use throws error
If I use snap to switch language to "Tamil". I get the following error:

Load failed:TypeError: Cannot read property 'add' of undefined.

image

Thanks for the great software!

Barthdry:
This may have fixed #953
## Change Log

- I have translated more keywords in tamil
- Also i have made some non translated or non translatable words to english

## Files Modified
Tamil Language file in locale

### DIFF
--- a/locale/lang-ta.js
+++ b/locale/lang-ta.js
@@ -1,10 +1,10 @@
 /*
 
-    lang-de.js
+    lang-ta.js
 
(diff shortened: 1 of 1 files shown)


jmoenig:
Thank you! Will pull for the next patch.

PR #2771, opened 2021-02-04
Trim usernames on signup

cycomachead:
This only applies to signups. We should also do this on the backend too.


jmoenig:
let me know if @bromagosa approves, then I'll be glad to pull is asap for v6.6. Thanks!

PR #2787, opened 2021-02-25
Final 6.6 manual and APL lib

brianharvey:
The library could still probably be improved, but it uses the new primitives so it's way faster already.

I think I'm finished working on 6.6, release when ready.


jmoenig:
yay, awesome! Thanks, will pull and release asap.

PR #2799, opened 2021-03-08
fix broken PIPE block

brianharvey:
Oops, see https://forum.snap.berkeley.edu/t/6-7-0-released/5730/3


jmoenig:
Ugh. Can you just take out the line-feed and let the PIPE block handle its layout itself? Some formulas are so short they fit into a single line. I would much rather prefer it that way.

brianharvey:
Umm, I suppose, but to me it looks really weird when all but one expression line up.

Anyway, the point is, in making this edit I apparently broke the block, because apparently input-list inputs disappear  (in the recursive call) when you change the appearance of a block.

jmoenig:
Yes, this needs to be fixed for sure, I agree. But can we also agree to take out the line break? I don't understand what you mean that "all but one expression" line up? In fact, *all* the expressions now line up perfectly and vertically, there isn't any single one behaving differently. But the *whole* variadic input can be moved to a fill the whole horizontal space in some situations, where it is getting wider than the standard block with. In *that* case it'll insert a line-break itself.

jmoenig:
you can just take the block from the frequency distribution library that's still intact

PR #2805, opened 2021-03-14
Catalan translation for 6.7.1

jguille2:
Hi Jens,

Late for 6.7.x because I see master is already building version 7...
But this updates translation file with current *master* state.

Thanks,

Joan

### DIFF
--- a/locale/lang-ca.js
+++ b/locale/lang-ca.js
@@ -185,7 +185,7 @@ SnapTranslator.dict.ca = {
     'translator_e-mail':
         'bernat@snap4arduino.rocks, jguille2@xtec.cat', // optional
     'last_changed':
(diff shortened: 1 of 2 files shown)


jmoenig:
thank you, Joan!
I will revert to 6.7.1 tomorrow and put the 7 dev version into a separate branch, because we need to push a patch with another fix anyway, so  this PR is timed exactly right!

jguille2:
Thanks Jens,

It will be great to have a separate "dev" branch. I guess version 7 is not scheduled for these next days and then, we can continue pulling to master for the "ordinary maintenance".

Now I'm updating Snap4Arduino and our 4 Moodle plugins, to build them with a more recent Snap! (with v6.7.1) And I was doing "hacky things" to add my Catalan patch in those versions :smile: 

Joan

PR #2806, opened 2021-03-14
fix strideTranspose

brianharvey:


jmoenig:
excellent, thank you! Will pull later today.

PR #2808, opened 2021-03-16
Update list-utilities.xml

brianharvey:
Remove WARP and also (sniff) remove empty slot substitution from the library # FOR EACH block.


jmoenig:
Thanks, will pull later this week for a new patch. I understand taking out WARP, but I don't think the library version also needs to remove empty slot substitution, especially not if you're already using it in the curriculum. It's perfectly fine for different blocks to offer different behavior, such as the library version also exposes the index, which is really useful. Up to you!

PR #2810, opened 2021-03-17
check if variable in repeat is number, fixes #2809

### LINKED ISSUE #2809: No typechecking for variables used in loops
Hi there,
if we use a string as an input in a `repeat ()` statement, we will get an infinite loop. I prepared a project to illustrate this here:
https://snap.berkeley.edu/snap/snap.html#present:Username=seegerer&ProjectName=wiederhole%20string%20mal&editMode=true

While I know that this not how a `repeat ()` statement is supposed to be used, but it might confuse learners, who might experience this while learning to code with Snap!

I suggest to add a check to the `repeat ()` statement that checks whether the literal can be parsed as an integer.

manzanillo:
This PR fixes the isse 2809 (https://github.com/jmoenig/Snap/issues/2809).
It just checks if the variable used in repeat is a number.

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -61,431 +61,420 @@ StageMorph, SpriteMorph, StagePrompterMorph, Note, modules, isString, copy, Map,
 isNil, WatcherMorph, List, ListWatcherMorph, alert, console, TableMorph, BLACK,
 TableFrameMorph, ColorSlotMorph, isSnapObject, newCanvas, Symbol, SVG_Costume*/
 
-modules.threads = '2021-March-15';
+modules.threads = '2021-March-15'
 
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks, Stefan, but ugh, why did you reformat the whole file instead of just the one line in question? That's probably a mistake, right? Can you maybe re-issue the PR with just the one change instead of changing every line in the file? That would be great. In case there's an issue with your IDE feel free to just point out your suggested change here in the comment section, then we can apply it together. Thanks!

manzanillo:
argh. Yeah, prettier autoformat on save feature. Wasn't aware of that. New PR incoming x)

PR #2811, opened 2021-03-17
add typecheck to doRepeat command, fixes #2809

### LINKED ISSUE #2809: No typechecking for variables used in loops

manzanillo:
This PR fixes the isse 2809 (#2809).
It just checks if the variable used in repeat is a number.

This time without having prettier "prettify" the whole file. Sorry, Jens :-)

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -2535,7 +2535,8 @@ Process.prototype.doRepeat = function (counter, body) {
         outer = this.context.outerContext, // for tail call elimination
         isCustomBlock = this.context.isCustomBlock;
 
-    if (counter < 1) { // was '=== 0', which caused infinite loops on non-ints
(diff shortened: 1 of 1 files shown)


jmoenig:
excellent, thank you, Stefan!

PR #2814, opened 2021-03-22
Update apl.xml

brianharvey:
fix deal, outer product


jmoenig:
thanks! Will pull asap when I get the chance to, but it will be later this week for a next little patch to be released towards the weekend.

PR #2830, opened 2021-05-13
Fixing 'lang' flag and anchor launcher

jguille2:
First fix. Lang launcher.

'lang anchor' and 'lang flag' were implemented after the main Snap! launcher.
This was not optimal... but it wasn't the real problem. The second asynchronous start killed the first one and the other flags were not applied and, the main problem, loaded projects were gone.

With this PR, we first look at these options (lang flag or anchor) and then, first (and only) start using that non-persistent lang option. Flag "wins" anchor to have a more consistent behavior (we can apply flags to all our anchors, changing their default behaviors).

If you agree with this, I will submit another PR, fixing '#run:' anchor start process... Continue...

Joan

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -364,9 +364,6 @@ IDE_Morph.prototype.openIn = function (world) {
         if (dict.noExitWarning) {
             window.onbeforeunload = nop;
         }
-        if (dict.lang) {
-            myself.setLanguage(dict.lang, null, true); // don't persist
-        }
 
         // only force my world to get focus if I'm not in embed mode
         // to prevent the iFrame from involuntarily scrolling into view
@@ -606,16 +603,33 @@ IDE_Morph.prototype.openIn = function (world) {
                 this.cloudError()
             );
         } else if (location.hash.substr(0, 6) === '#lang:') {
-            urlLanguage = location.hash.substr(6);
-            this.setLanguage(urlLanguage, null, true); // don't persist
-            this.loadNewProject = true;
+            dict = myself.cloud.parseDict(location.hash.substr(6));
+            applyFlags(dict);
         } else if (location.hash.substr(0, 7) === '#signup') {
             this.createCloudAccount();
         }
         this.loadNewProject = false;
     }
 
-    if (this.userLanguage) {
+    function launcherLangSetting() {
+        var langSetting = null;
+        if (location.hash.substr(0, 6) === '#lang:') {
+            if (location.hash.charAt(8) === '_') {
+                langSetting = location.hash.slice(6,11);
+            } else {
+                langSetting = location.hash.slice(6,8);
+            }
+        }
+        // lang-flag wins lang-anchor setting
+        langSetting = myself.cloud.parseDict(location.hash).lang || langSetting;
+        return langSetting;
+    }
+
+    if (launcherLangSetting()) {
+        // launch with this non-persisten lang setting
+        this.loadNewProject = true;
+        this.setLanguage(launcherLangSetting(), interpretUrlAnchors, true);
+    } else if (this.userLanguage) {
         this.loadNewProject = true;
         this.setLanguage(this.userLanguage, interpretUrlAnchors);
     } else {
--- a/src/gui.js
+++ b/src/gui.js
@@ -457,26 +457,38 @@ IDE_Morph.prototype.openIn = function (world) {
                     }
                 );
             }
-        } else if (location.hash.substr(0, 5) === '#run:') {
+       } else if (location.hash.substr(0, 5) === '#run:') {
+            dict = '';
             hash = location.hash.substr(5);
-            idx = hash.indexOf("&");
-            if (idx > 0) {
-                hash = hash.slice(0, idx);
-            }
+
+            //decoding if hash is an encoded URI
             if (hash.charAt(0) === '%'
                     || hash.search(/\%(?:[0-9a-f]{2})/i) > -1) {
                 hash = decodeURIComponent(hash);
             }
-            if (hash.substr(0, 8) === '<project>') {
-                this.rawOpenProjectString(hash);
-                applyFlags(myself.cloud.parseDict(location.hash.substr(5)));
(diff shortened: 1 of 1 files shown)


jguille2:
Hi Jens,

To be able to continue until you decide on this PR without code conflicts... I continue here fixing the "#run:" anchor behavior.

If you don't agree with this, I will revert these changes...

jmoenig:
Hi Joan,
can you remind me again what issue this addresses? Do you want to open a project in a specified language without making the language setting persistant, is that it?
If memory serves I made it so you can open Snap! in a given language (and not make that language be persistent), but you cannot combine it with opening or running a specific project. Is that what you're changing here?
Thanks!

jguille2:
Sorry Jens,

I've already push the other fix...

I want to answer first your question about the first fix, and I will explain later the "#run:" fixes.

About lang:

- Current behavior.
  + First, Snap! start with the "userLanguage" setting (to "paint" all the IDE with this setting, or "en" by default).
  + After this start, Snap! apply anchors. And then, it loads a project (from cloud, from url, direct with a xml in the url called...)
  + And after this, it apply flags... and then, maybe it wants to repaint the IDE with other "non-persistent setting" language.
-Problems.
  Those asynchronous processes have conflicts. Other flags are not applied (because the last "building" delete their action). Ant the big problem... The loaded project disapears!!

- This lang fix
  If there are lang flags or anchors, we can make our first "Snap! building" with that language (the same we do with the userLanguage setting). After this, we can apply achors and flags without any asynchronous conflict.

That's all.
If it is no clear, I can prepare some examples of these conflicts.

Joan

jguille2:
I wait to finish this discussion and then I will explain the "#run: anchor" fixes.

If something is wrong, I can delete all and start again... No problem.

jmoenig:
Thanks, Joan, for the explanation. I'm wondering whether these race-conditions are really caused by asynchronous processes. That doesn't seem very likely, because the only asynchronous task is reading a project file, isn't it? I'm suspecting something else to maybe be the culprit here: the ```Morph.nextSteps``` property. If more than a single function set it, previous quasi-processes get cancelled. Could that be the case?

jguille2:
I think that the asynchronous processes is part of the problem.

- I test this:
https://snap.berkeley.edu/snap/snap.html#run:/snap/Examples/vee.xml&lang=es&noCloud
and I see 'es' language (ok), 'noCloud' also applied (ok) but the IDE is en 'editMode' (and you know 'run' default is in "presentMode")

- and
https://snap.berkeley.edu/snap/snap.html#run:/snap/Examples/vee.xml&lang=en&noCloud
(yes, the same but in English)
again language and 'noCloud' ok, again 'editMode' (ko) and 'run' is not applied (green flag is not pressed)
And testing this (last url) in different browsers, give me some surprises (in Firefox the costume of Sprite(2) is lost)

- A lot of differences testing with other anchors (see later) but also different problems with different projects, because "times" are different. Doing this inside a Moodle is worse, because javascript is doing other things... and I see a lot of surprises (unexpected behaviors).

- Why async. problems? You know 'setLanguage' has a 'callback'. It is 'interpretUrlAnchors'. And this has (or it can has) 'getURL' to load the project, also async. And with the current behavior, lang "anchor or flag" is calling a new 'setLanguage'...

cycomachead:
I think this is a step forward and does clean up a couple issues, though at some point we should consider refactoring this code.

jguille2:
Hi!
First of all, I want to say again that I can close this PR and restart again with your orientations if these solutions are not fine. No problem. But I comment here my second commit (fixing '#run:' anchor bugs) to show the whole subject... and consider a this "partial" solution for our launcher.

- `#run:`anchor bugs
We have different problems here... 
  + _hash_ is sliced till the first `&`, trying to take off _flags_. But our _xml projects_ have a lot of `&` and so, passing the xml content by URL is not running.
  + `hash.substr(0, 8) === '<project>'` is not right. It only needs to delete the `>' character.
  + But the end of the project must be `</project>`, not `&`
  + And then, the _dict_content (flags) are not well calculated.
  + Different cases (xml content, relative paths, absolute paths...) are reacting differently.
 
- This fix
I think the code is clear. I try to ensure that all the possible project values ('xml source' and 'xml paths' (absolute, relative to site and relative to page)) are running ok and apply flags when required.
This commit fix all the `#run:` behavior except 'lang flags'. The other (first) commit does it.

- A general solution
Yes, I think we can recode our _openIn_ function with a clearer code (where some actions like 'decodeURIComponent', 'find anchor and flags', 'setLanguage'... appears only once). And maybe I could do this because I think I know what we want ('full functionality' of our current options, a "full backward compatibility", and maybe a new more general anchor with all the possibilities integrated (cloud, web and local projects, all behaviors (related to flags) available...
But to do this I need to know more clearly Jens opinions... or maybe Jens want to do this because it is pretty critical (our launcher!)

- But the powerful of this fix...
_&editMode_ and _&noRun_ flags are not the default _#run:` behavior. And then, with only this anchor, we can control all the possibilities of our launcher.
So, after this fix, I will use this anchor (run) in a new version of the Moodle plugin to launch Snap! inside the classroom environment in the best way. We can customize the Snap! window depending on its use (working, grading, showing collections...), also depending on user (Moodle user) preferences (like lang) and other administration decisions (like Cloud options).
Current Moodle plugin already does some of these options, executing some JS at the beginning... but there I have a lot of troubles  managing asynchronous processes of that starting page.

Thanks!

Joan

jmoenig:
Thanks, Joan! I hope to come up with some time to have a closer look and conduct some tests myself, but I like the overall idea of this, let's get it in!

jguille2:
Thanks Jens!
Now I can start creating the new Snap! Moodle plugin :)
The main goal is that it will work with our official online distros (using postMessage API feature)... but I was waiting because I wanted to have a "cleaner" launcher without being triggering JS scripts and without those asynchronous problems.

Joan

jmoenig:
thank *you*, Joan, for this very welcome contribution and great solution to a longstanding issue!

PR #2832, opened 2021-05-16
CursorMorph renders with same color as its target

towerofnix:
Simple fix to make the text cursor render with the same color as the text on its target element. (In current Snap, it always renders black.)

I encountered this when trying out the "fade blocks" feature; I made them dark enough that the text in input slots turned white (to personal taste, 80%). Then I typed in an input, but found the cursor was nearly illegible against the relatively dark background:

A simple script with the cursor inside a "say" block's text field. It's a small black line, hard to see against the dim purple background.

With this fix, the cursor is much brighter and easier to read, just like intended for the actual text:

The same script with the cursor in the sample; now, it's much easier to see and locate.

Since this uses the same color as the text (which [draws using the same method](https://github.com/jmoenig/Snap/blob/21a681fc3cf89194164ddf470a4f958e9d57d656/src/morphic.js#L8588)), this should always produce an equally visible cursor—e.g, in other text inputs which remain black on white, the cursor respects this:

The cursor inside a sprite's name field. Like the text, the cursor is black against a white background.

(PS: though offtopic, I love that Snap! uses the font I've customized in the browser as my default sans-serif!)

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -5513,6 +5513,7 @@ CursorMorph.prototype.init = function (aStringOrTextMorph, aTextarea) {
     // override inherited defaults
     ls = fontHeight(this.target.fontSize);
     this.setExtent(new Point(Math.max(Math.floor(ls / 20), 1), ls));
+    this.color = this.target.color;
     
     if (this.target instanceof TextMorph &&
             (this.target.alignment !== 'left')) {
@@ -5523,6 +5524,12 @@ CursorMorph.prototype.init = function (aStringOrTextMorph, aTextarea) {
     this.gotoSlot(this.slot);
     this.updateTextAreaPosition();
     this.syncTextareaSelectionWith(this.target);
+};
+
+// CursorMorph drawing:
+
+CursorMorph.prototype.getRenderColor = function () {
+    return this.target.getRenderColor();
 };
 
 // CursorMorph event handling


jmoenig:
Thanks, but I'm not interested in changing the cursor color.

towerofnix:
@jmoenig That's fair enough and I don't mean to bother you if you don't have the time, though if you do, could you explain why you're not interested in that? I figured it's only pulling a really small improvement, and lends significant usability to the "fade blocks" feature, where otherwise entering text is considerably less accessible/convenient than in the default look.

cycomachead:
For what it's worth, I think this makes a lot of sense. I don't use fade blocks really, but the text cursor probably should match the text color. 🙂

jmoenig:
No text editing tool I've ever used matches the cursor color to the text color. Some settings are more legible than others, and I'm trying to minimize the ranges with really low contrast. What really could improve things is always the same strategy: Let everything be configurable, including the cursor color, the selection color, fonts, etc. etc. But that comes with other downsides, including maintenance and running into yet other edge-cases requiring further customization.

RubenVerg:
> No text editing tool I've ever used matches the cursor color to the text color.

Matching the text color is just an implementation detail, since the text will always be black or white. Most editors have a contrasting cursor color with the background, and since text already does that, the cursor can follow.

cycomachead:
> No text editing tool I've ever used matches the cursor color to the text color.

Jens, you use your mac in dark mode. :) The default text cursor is white when the text is white and black otherwise.

jmoenig:
Pages, bright mode, black font with dark background

pages

RubenVerg:
> pages

You opted in to lack of contrast, what about white text on a dark background?

jmoenig:
dark bg

RubenVerg:
yuck is there a place to file Pages bugs?

RubenVerg:
Although you should compare Snap text editing to plain text editors, not those with formatting options.

cycomachead:
I mean not editors but like standard macOS text inputs -- we are not talking about formatted text here. (like the Safari URL bar or finder search)

I suppose the cursor color could be more accurately described as a function of the background than text color, though I still think this is an improvement in legibility.

jmoenig:
Things can definitely get weird in certain ranges of "fade blocks", but general solutions might not help that much. Here's a 80% fade setting with a lighter background, where the black cursor actually fits better.

RubenVerg:
Does it fit better though? It just sticks out

towerofnix:
@jmoenig I'm in agreement with @cycomachead, i.e. that in UI elements which automatically set their text to white to contrast with a darker background, adjusting the text color to follow. This is distinct from opting into low contrast via inline formatting options (which yes, is a somewhat more difficult issue to solve, because dynamically switching cursor color *within one input* is a somewhat odd behavior).

I think it's inevitable that some options of "fade blocks" will look better than others, but if the *best* options exist only on one side of the scale—i.e, when "fade blocks" isn't used at all—that indicates a flaw in the UI which altogether undermines the feature, and gives the impression that it's not something you're "supposed" to use. Totally customizable UI would let users tweak to how they like, but IMO, *even if* an option were added to set a "cursor color mode", it would still be best to provide sensible defaults! Making the cursor follow the text color would definitely help significantly on the other end of "fade blocks", making it appear as a user preference given attention to work well like any other feature.

(I'm also in agreement with @RubenVerg that the white cursor would look better in that screenshot, though I know that's primarily a matter of taste.)

PR #2835, opened 2021-05-19
Hindi Translation Completed for Now

Barthdry:
I have completed translating hindi's most visible elements while opening but still only 50% of the element is completed to be fair but i hope this can be pulled for now as i am busy for now and also it is hard to type on onscreen keyboard of hindi and also not an expert. Maybe i may continue this later so i am giving a pull request for now.


jmoenig:
Awesome, thank you very much!
I'll pull this for the next patch / minor release.

PR #2837, opened 2021-05-24
Consider empty rings empty slots

brollb:
This is the fix I alluded to in https://github.com/jmoenig/Snap/issues/2836. If empty rings are not considered empty slots by design, then this can be rejected, of course :) I was going to wait to open the PR but thought it might actually help to explain my thinking in the comments in the issue!

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -6678,6 +6678,10 @@ RingMorph.prototype.dataType = function () {
     }
 };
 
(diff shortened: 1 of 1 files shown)


jmoenig:
I remember changing this for v7, now empty rings inside HOFs like MAP etc. should be considered empty slots. Geez, I *don't* remember how I did that, though. But I'm hoping that your suggestion now is the official behavior ;-)

PR #2856, opened 2021-07-06
Hide all primitives works in Single Palette mode.

cycomachead:
This is a small fix to make the 'hide primitives' function work as before.

* also, calling "flushBlocksCache" in unified/single mode clears the whole cache

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -5797,7 +5797,7 @@ IDE_Morph.prototype.switchToDevMode = function () {
 
 IDE_Morph.prototype.flushBlocksCache = function (category) {
     // if no category is specified, the whole cache gets flushed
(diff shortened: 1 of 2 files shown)


cycomachead:
Oh I think I need to fix `show primitives` too... Also, dan found another bug with hide primitives in the non-single palette mode to investigate.

jmoenig:
This is nice and I'm going to pull it now. Then we can continue working on it.
I think we need to introduce another "mode" where we show all hidden blocks without changing their visibility state. Also, I want to make it so there is a common UI for both primitives and custom blocks to edit their palette visibility state. Once we have that "show all blocks" mode, we can then decide to also suppress hidden blocks in the search results.

PR #2857, opened 2021-07-07
Display block with error

cycomachead:
This replaces #1781 -- I've remade the commits since the history was kind of a mess and the original commit in that PR was quite old. I think this is a nicer implementation anyway. :) 

This doesn't do much to always try to find the best thing to show, but it does include a case for when a single variable errors, it shows the caller. Presumably there's other cases where we could find improvements....

Screen Shot 2021-07-06 at 7 03 24 PM

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -1176,9 +1176,9 @@ Process.prototype.errorBubble = function (error, element) {
     // above the text of error.
     var errorMorph = new AlignmentMorph('column', 5),
         errorIsNested = isNil(element.world()),
-        errorPrefix = errorIsNested ? localize('Inside a custom block') + ':\n' : '',
+        errorPrefix = errorIsNested ? `${localize('Inside a custom block')}:\n`: '',
         errorMessage = new TextMorph(
-            errorPrefix + localize(error.name) + ':\n' + error.message,
+            `${errorPrefix}${localize(error.name)}:\n${localize(error.message)}`,
             SyntaxElementMorph.prototype.fontSize
         ),
         img, blockImage;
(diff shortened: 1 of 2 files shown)


src/threads.js
```diff
@@ -1174,6 +1171,32 @@ Process.prototype.errorObsolete = function () {
     throw new Error('a custom block definition is missing');
 };
 
+Process.prototype.errorBubble = function (error, element) {
+    // Return a morph containing an image of the elment causing the error
+    // above the text of error.
+    var errorMorph = new AlignmentMorph('column', 5),
+        errorIsNested = isNil(element.world()),
+        errorPrefix = errorIsNested ? `${localize('Inside a custom block')}:\n`: '',
```
jmoenig, line 1179: hmm... I don't understand those funny characters ``` `${` ``` is that some wild new ES12 syntax that I need to catch up with?

jmoenig: Oh! It's what they call a "template string". Look at that!! I've just learned something new 😂

src/threads.js
```diff
@@ -1174,6 +1171,32 @@ Process.prototype.errorObsolete = function () {
     throw new Error('a custom block definition is missing');
 };
 
+Process.prototype.errorBubble = function (error, element) {
+    // Return a morph containing an image of the elment causing the error
+    // above the text of error.
+    var errorMorph = new AlignmentMorph('column', 5),
```
jmoenig, line 1177: I love this design, Michael!

src/threads.js
cycomachead: I LOVE template strings! In my own use, they've pretty much replaced all `a + ' ' + b` with ``${a} ${b}`` in my own use. It's also nice because you don't need to escape quotes. :) 

(Though, of course, you do need to escape \`, but that's pretty rare!)

src/threads.js
cycomachead: thanks! 

Also, how do you feel about making the error speech bubbles "clickable"? I've been working with a couple folks, and one thing is that they disappear a little quickly.

jmoenig: I'm open to suggestions and experiments!

src/threads.js
jmoenig: how could I not know about template strings until today?! Of course they're incredibly useful. Thanks for showing me!

brianharvey: I never heard of them before, either, but as soon as I saw that they're delimited with backquotes it was clear what they are!  This is a longstanding Lisp feature, not for strings, but for Lisp expressions generally.  Lisp uses comma, though, instead of dollarsign; I guess the JS people thought people would be more likely to put a comma in by accident.

The question is, can you nest them?  I.e., can you put one inside the ${...}?   In Lisp, if you do that, you can say, e.g., ",,(...)" to evaluate the expression two environments out.

cycomachead: you can indeed nest them! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#nesting_templates

The "tags" are kind of neat too, though I don't personally use that much.

PR #2858, opened 2021-07-12
Update lang-zh_CN.js

Simon-Mong:
Update the translation of new blocks and delete some unnecessary translations of the names of Math functions.


jmoenig:
Thank you, @Simon-Mong ! Pulling now....

PR #2860, opened 2021-07-16
follow up fixes for single palette

cycomachead:
* fix hiding/showing primitives when using the variables palette (in not single palette mode)
* in single palette mode, do _not_ show labels if there are no blocks beneath the label.

The second fix is useful because in current master, you get overlapping labels when hiding all blocks. But also for some of the microworld-y things, we don't want to call extra attention to the fact that there could be some other block.

### DIFF
--- a/src/objects.js
+++ b/src/objects.js
@@ -3118,7 +3118,7 @@ SpriteMorph.prototype.freshPalette = function (category) {
     if (category === 'unified') {
         // In a Unified Palette custom blocks appear following each category,
         // but there is only 1 make a block button (at the end).
-        // arrange the blocks in the unified palette column-wise:er
+        // arrange the blocks in the unified palette column-wise:
         let cat1 = this.categories.slice(0, 8),
             cat2 = this.categories.slice(8);
 
--- a/src/objects.js
+++ b/src/objects.js
@@ -2847,10 +2847,7 @@ SpriteMorph.prototype.customBlockTemplatesForCategory = function (category) {
         isInherited = false, block, inheritedBlocks;
 
     function addCustomBlock(definition) {
-        if (!definition.isHelper &&
-            (definition.category === category ||
-            (Array.isArray(category) && category.includes(definition.category)))
-        ) {
+        if (!definition.isHelper && definition.category === category) {
             block = definition.templateInstance();
             if (isInherited) {block.ghost(); }
             blocks.push(block);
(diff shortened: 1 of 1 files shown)


src/objects.js
```diff
@@ -2847,10 +2847,7 @@ SpriteMorph.prototype.customBlockTemplatesForCategory = function (category) {
         isInherited = false, block, inheritedBlocks;
 
     function addCustomBlock(definition) {
-        if (!definition.isHelper &&
-            (definition.category === category ||
-            (Array.isArray(category) && category.includes(definition.category)))
-        ) {
+        if (!definition.isHelper && definition.category === category) {
```
cycomachead, line 2850: This was kind of a nasty one to debug. IMO, it's best if this function doesn't ever have to think about categories being an array...

src/objects.js
```diff
@@ -3041,6 +3038,11 @@ SpriteMorph.prototype.freshPalette = function (category) {
                     ]
             };
 
+        if (category === 'unified') {
```
cycomachead, line 3041: We need this up here to cover the 'show primitives' function too.

src/objects.js
```diff
@@ -3120,7 +3118,7 @@ SpriteMorph.prototype.freshPalette = function (category) {
     if (category === 'unified') {
         // In a Unified Palette custom blocks appear following each category,
         // but there is only 1 make a block button (at the end).
-        // arrange the blocks in the unified palette column-wise:
+        // arrange the blocks in the unified palette column-wise:er
```
cycomachead, line 3121:
```suggestion
        // arrange the blocks in the unified palette column-wise:
```

src/objects.js
```diff
@@ -3129,36 +3127,35 @@ SpriteMorph.prototype.freshPalette = function (category) {
...
+        }, []);
     } else {
         // ensure we do not modify the cached array
         blocks = this.getPrimitiveTemplates(category).slice();
     }
     blocks.push('=');
     blocks.push(this.makeBlockButton(category));
 
-    if (category === 'variables') {
-        category = ['variables', 'lists', 'other'];
```
cycomachead, line 3154: Well, this was my fault trying to be concise, but is the real problem, because this is what confused the hide/show primitives.

src/objects.js
```diff
@@ -3041,6 +3038,11 @@ SpriteMorph.prototype.freshPalette = function (category) {
...
             };
 
+        if (category === 'unified') {
+            more.unified = Object.values(more).reduce((x, y) =>
+                x.concat(y));
+        }
+
         function hasHiddenPrimitives() {
             var defs = SpriteMorph.prototype.blocks,
                 hiddens = StageMorph.prototype.hiddenPrimitives;
```
cycomachead, line 3048: Probably not worth re-writing, but I wonder if it would be more clear to iterate over blocksForCategory rather than all blocks...

jmoenig:
thanks, Michael! Pulling now...

jmoenig:
Uh-oh, now there's a new problem of overlapping categories when I hide one in between:

overlapping categories

jmoenig:
Ah, it only happens with operators, because that's in front of the special "variables" category...

cycomachead:
Argh! I saw that happen once, then I thought I'd fixed it after refreshing. :/

jmoenig:
Aha, this isn't even a new bug, it's been there all along. Okay, that actually makes it easier for me to find it :-)

jmoenig:
it's the horizontal alignment of the Rings in the operator's category that's causing it once you hide all of them.

jmoenig:
fixed! it was all my fault 🤦‍♂️

cycomachead:
ahhh. thanks! :)

Issue #2863, opened 2021-07-17
very minor misfeature in RESHAPE


brianharvey:
a RESHAPE to no dimensions should report the first item as a scalar, not the empty list.  (The shape of an empty list is [0], not [].  One dimension, length zero.)  I checked and real APL reports a scalar.

RubenVerg:
~So Snap! is fake APL?~

PR #2864, opened 2021-07-17
reshape to no dimensions

brianharvey:
I hope you don't mind me sneaking an executable line in between var declarations.  Makes it not have to do a redundant test. :)

Fixes issue #2863 that I just filed...

### DIFF
--- a/src/lists.js
+++ b/src/lists.js
@@ -687,9 +687,9 @@ List.prototype.reshape = function (dimensions) {
     // answer a new list formatted to fit the given dimensions.
     // truncate excess elements, if any.
     // pad with (repetitions of) existing elements
-    var src = this.ravel().itemsArray(),
-        size = dimensions.isEmpty() ? 0
-            : dimensions.itemsArray().reduce((a, b) => a * b),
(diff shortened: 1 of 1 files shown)


src/lists.js
```diff
@@ -687,9 +687,9 @@ List.prototype.reshape = function (dimensions) {
     // answer a new list formatted to fit the given dimensions.
     // truncate excess elements, if any.
     // pad with (repetitions of) existing elements
-    var src = this.ravel().itemsArray(),
-        size = dimensions.isEmpty() ? 0
-            : dimensions.itemsArray().reduce((a, b) => a * b),
+    var src = this.ravel().itemsArray();
```
cycomachead, line 690: Jens’ style would be to do this:
```suggestion
    var size, src = this.ravel().itemsArray();
```

then remove the next var

brianharvey: OK, done, except there were two more variables to declare so I moved those up too.

Sorry for all the confusing reverts; somehow Github Desktop seemed to be saying that I'd pushed this change right into the master branch.  But all is well.

PR #2867, opened 2021-07-20
Block user HTTP requests to backend API

bromagosa:


RubenVerg:
do those API links work because the request is sent with credentials? maybe it's better to just disable those and let the request return 403 or whatever

- RubenVerg ❤️


________________________________

bromagosa:
We can't disable these from the backend. They're needed by both the frontend and the community site, and there's no way to distinguish between a legitimate call (say, from the cloud menu in the editor) and a call issued by a user using the URL block of the XHR extension.

RubenVerg:
> do those API links work because the request is sent with credentials?

bromagosa:
Some need to work *without* credentials, of course. How can I ask whether someone is logged in unless it's without credentials, for example?

We've been going over several solutions for years... this one works now that the URL is received by the primitive already fully formed.

joker314:
**SECURITY**
The RegEx generated by this pull request appears to be case-sensitive, meaning that a URL of the following kind is likely to bypass the filter
```
https://sNaP.bErKeLeY.eDu/logout
```

To make the regular expression case-insensitive, consider
```diff
- url.match(`snap\.berkeley\.edu.*${which}`)
+ url.match(new RegExp(`snap\.berkeley\.edu.*${which}`, "i"))
```
or alternatively, convert the URL being tested to lowercase first - before it's fed into the regular expression

RubenVerg:
also it completely ignores mirrors/forks that support some form of login, maybe it should match for `location.host`?

jguille2:
Hi,

For me, current code (pointing to 'snap.berkeley.edu') is fine.

But if you want to debate this, for me the best is not 'berkeley' or 'location.host'... because the key is the 'clour url', that can be changed if somebody build another cloud.
Then `Cloud.url` is the key, because the "problem" is prevent "doing things" when our users are logged, and they are logged (if they are) in that cloud.

Joan

PR #2868, opened 2021-07-23
Better handling of blocked cloud URLs

bromagosa:
Check URL against all known cloud domains, and make the regex case insensitive.


jmoenig:
I'm fine with these changes. @bromagosa and @cycomachead are you okay if I pull these before SnapCon? Tnx

cycomachead:
Yeah. That’s good. 

-- 
Michael Ball
From my iPhone
michaelball.co

PR #2869, opened 2021-07-25
Show Hidden Primitives in Find, and allow Showing them

cycomachead:
Right now, it's pretty hard to build a project that uses hidden primitives, because it's
a lot of work when you realize you need something that's hidden.

This adapts the Find function to show hidden primitives (hidden custom blocks are findable).
I also realized we have no individual `show` function, so that's been added. Before there was no way to get to it. :)

_Long Term_: I think find is kind of add odd place to reveal hidden primitives... However, for the purposes of writing and updating projects for students it's a good solution. Ultimately some alternative view that lists all blocks an their status would be welcome, but that's for later.


cycomachead:
Screen Shot 2021-07-25 at 12 45 37 PM

Online: 
https://snap.berkeley.edu/versions/michael-dev/snap.html

src/blocks.js
```diff
@@ -2814,10 +2814,17 @@ BlockMorph.prototype.userMenu = function () {
                     );
                 }
             } else if (this.selector !== 'evaluateCustomBlock') {
-                menu.addItem(
-                    "hide",
-                    'hidePrimitive'
-                );
+                if (this.isHidden()) {
+                    menu.addItem(
+                        'show',
```
cycomachead, line 2819: This text is a little bit vague (given that you can see the block), but I think it makes enough sense in the context, and given that it's the opposite of hide, I think it's fine.

src/blocks.js
```diff
@@ -3239,23 +3246,44 @@ BlockMorph.prototype.developersMenu = function () {
...
         doDeclareVariables: 'variables'
     };
-    cat = dict[this.selector] || this.category;
-    if (cat === 'lists') {cat = 'variables'; }
-    ide.flushBlocksCache(cat);
-    ide.refreshPalette();
+
+    palette = dict[this.selector] || this.category;
+    if (palette === 'lists' || palette === 'other') {palette = 'variables'; }
+    return palette;
```
cycomachead, line 3286: I called this the `palette` and not the `category` b/c really that's where the block 'lives'. 

Also not sure it was worth updating, but similar code exists in `Process.prototype.doChangePrimitiveVisibility`. This method could probably be used there. (And all of this speaks to the fact that I think it might be time for a `PaletteMorph` object that collects all this, but eh....)

src/blocks.js
pamelafox: Would un-hide be clearer?

jmoenig:
hmm... I'm not going to pull this before SnapCon. But I'm also working on a different system for hiding and showing blocks, one that works the exact same way for both primitives and custom blocks, also doesn't show hidden blocks in the search results, lets you hide any block programmatically (using the `hide var` mechanism we already have in place) and has a "view all hidden blocks" mode. But this is going to be for after SnapCon...

cycomachead:
Ah, that's cool! 

I think it's fine to wait, but we really do at least need to be able to show individual blocks to be able to make progress on building curriculum projects. So if it's going to be a long time, then I think this would be a useful short term solution. :)

cycomachead:
I am going to close this, since the recent change to allow blocks to be dragged out of the Hide Blocks palette should address the same need.

PR #2875, opened 2021-08-06
added blocksZoom URL param to set blocks scale for the current session

bromagosa:


cycomachead:
don't think a URL is a problem, but I wonder if we should consider a project setting?

jmoenig:
apparently this was something urgent for a course of the OpenUniversity Barcelona starting in September. I agree that we want to come up with an understanding of which settings can be saved where. For this particular use case it was best to make sure the setting does *not* get saved anywhere at all.

bromagosa:
Just pointing out that this was urgent for the EDC Math+C project, not the UOC course :)

PR #2876, opened 2021-08-12
Allow Redirection to /dev when newer version of Snap! is needed

cycomachead:
This solves the issue of people who have used newer versions of Snap! but share their projects on the Snap!Cloud. 

It pops a warning up at the time of parsing, and does *not* automatically redirect.

I defined `IDE_Morph.prototype.developmentPath` such that an extensions could overwrite this path.
Setting the value to simply `snap.html` effectively disables the redirect message. 

If you are already on a dev version and somehow are using an even newer version of Snap! then you are not offered a redirect.

Screen Shot 2021-08-11 at 5 05 49 PM

@jmoenig I know this means a 6.9.1 but I think it would be helpful. I can update history, cut 6.9.1 and then merge into v7.


cycomachead:
I took care off all the usual tasks to prepare a release, so this is basically ready.

HISTORY.md
```diff
@@ -2,10 +2,17 @@
...
+This is a small patch to support projects which have begun to use v7 development.
+
+* **Notable Changes:**
+  * Snap! now offers to redirect you to the development version if you load a file with a newer serializer version.
+  * Extension authors who maintain development versions may want to set `IDE_Morph.prototype.developmentPath`
+  * Extension authors who want to disable this behavior should set `IDE_Morph.prototype.developmentPath = 'snap.html'`
 ## 6.9.0
 * **Notable Changes:**
     * JS-functions are now disabled by default until switched on in the settings menu per session
-    * error messages in presentation mode are now shown as pop-up messages onstage
```
cycomachead, line 8: Kinda annoying, but all the changes below in this file are just trimming whitespace.

src/gui.js
```diff
@@ -5392,7 +5414,7 @@ IDE_Morph.prototype.openProject = function (name) {
         this.setProjectName(name);
         str = localStorage['-snap-project-' + name];
         this.openProjectString(str);
-        this.setURL('#open:' + str);
+        this.setURL('open:' + str);
```
cycomachead, line 5417: Slightly unnecessary, but so is `#`.
We should consider removing this altogether.

src/gui.js
```diff
@@ -7576,13 +7598,9 @@ ProjectDialogMorph.prototype.rawOpenCloudProject = function (proj, delta) {
             this.ide.nextSteps([
                 () => this.ide.openCloudDataString(clouddata)
             ]);
-			location.hash = '';
-            if (proj.ispublic) {
```
cycomachead, line 7580: All projects have a URL that they can open form.

src/gui.js
```diff
@@ -7614,6 +7632,9 @@ ProjectDialogMorph.prototype.saveProject = function () {
                 );
             } else {
                 this.ide.setProjectName(name);
+                location.hash =
```
cycomachead, line 7635: Side effect: If you open someone else's project and save to your account, the URL now updates to your project! 🎉

src/gui.js
```diff
@@ -7702,16 +7723,6 @@ ProjectDialogMorph.prototype.shareProject = function () {
                         this.buttons.fixLayout();
                         this.rerender();
                         this.ide.showMessage('shared.', 2);
-
-                        // Set the Shared URL if the project is currently open
-                        if (proj.projectname === ide.projectName) {
```
cycomachead, line 7707: This is removed b/c by definition, the project URL should already be present. (And thus there's no need to see if you are sharing a different project than the open one.)

src/gui.js
```diff
@@ -7748,9 +7759,6 @@ ProjectDialogMorph.prototype.unshareProject = function () {
                         this.buttons.fixLayout();
                         this.rerender();
                         this.ide.showMessage('unshared.', 2);
-                        if (proj.projectname === ide.projectName) {
```
cycomachead, line 7751: Same deal as before, the URL is still a functional URL! :)

cycomachead:
Pre Jens' approval, I am going to merge this in 6.9. We'll officially deploy in 24-48 hours. :)

PR #2881, opened 2021-08-18
Port Serializer Redirection, Cloud Project URLs to v7

cycomachead:
Using another branch so I can resolve the conflicts.


cycomachead:
@jmoenig This is the PR to back port (forward port?) the 6.9.1 changes to v7. Should be a clean merge now. :)

src/gui.js
```diff
@@ -8240,10 +8251,7 @@ ProjectDialogMorph.prototype.unshareProject = function () {
                         entry.label.rerender();
                         this.buttons.fixLayout();
                         this.rerender();
-                        this.ide.showMessage('unshared.', 2);
-                        if (proj.projectname === ide.getProjectName()) {
-                            location.hash = '';
-                        }
+                        this.ide.showMessage('unshared.', 2
```
cycomachead, line 8254:
```suggestion
                        this.ide.showMessage('unshared.', 2);
```

Ugh, this was a bad merge

libraries/bignumbers.xml
```diff
@@ -1 +1 @@
...
...ors"><comment x="0" y="0" w="234.66666666666666" collapsed="false">The identity function: reports its input.&#xD;It&apos;s useful to get things like 3/4 or 5-2i into&#xD;numeric input slots.</comment><header></header><code></code><translations></translations><inputs><input type="%s"></input></inputs><script><block s="doReport"><block var="x"/></block></script></block-definition><block-definition s="Scheme number %&apos;function&apos; of %&apos;number&apos;" type="reporter" category="operators"><comment x="0" y="0" w="300" collapsed="true">Provides Scheme arithmetic functions not in JavaScript</comment><header></header><code></code><translations>pt:_ de _&#xD;</translations><inputs><input type="%s" readonly="true"><options>number?&#xD;complex?&#xD;real?&#xD;rational?&#xD;integer?&#xD;exact?&#xD;inexact?&#xD;exact&#xD;inexact&#xD;finite?&#xD;infinite?&#xD;nan?&#xD;numerator&#xD;denominator&#xD;real-part&#xD;imag-part&#xD;magnitude&#xD;angle</options></input><input type="%s"></input></inputs><script><block s="doReport"><block s="reportApplyExtension"><l>big_scheme(fn, num)</l><list><block var="function"/><block var="number"/></list></block></block></script></block-definition></blocks>
```
jmoenig, line 1: is this the "old" v6 pre-extension prims version? If so we'll have to be careful not to overwrite the v7 changes....

cycomachead: The difference is only in the newline. (Not really sure why this one has one and the other didn't, but it's fine.)

Also we probably should update these to say Snap! 7 at some point. :)

jmoenig: > Also we probably should update these to say Snap! 7 at some point. :)
Whoa, you're right! Thanks for catching this

cycomachead: Probably just after we swap to just 7. :) (Since, this won't break anything, and it's mostly a nicety in this case.)

cycomachead:
@jmoenig One last thing to merge before v7.

PR #2905, opened 2021-11-01
Rename `reportListLength` to `reportListAttribute`

WarpedWartWars:
So you get *something* when you click `help...` on ([length v] of E]), instead of nothing, even if it isn't quite correct.


jmoenig:
good catch, thank you!

WarpedWartWars:
You're welcome, Jens.

PR #2910, opened 2021-11-12
Adding Help Image To SCRIPT VARIABLES Block

joecooldoo:
This PR is for the addition of the Script Variables help message.


brianharvey:
Three things.  First, what does your edit to snap.html have to do with help screens?

Second, we already have a help screen for SCRIPT VARIABLES.  That's not one of the blocks that need attention.

Third, a moot point because of the second, but just so you know, it's not quite true that the variable disappears when the script ends.  If the script reports a procedure (made with a gray ring) that uses the variable, the variable lasts as long as something points to that procedure.  This is how we do local state in OOP.

joecooldoo:
> Three things. First, what does your edit to snap.html have to do with help screens?
> 
> Second, we already have a help screen for SCRIPT VARIABLES. That's not one of the blocks that need attention.
> 
> Third, a moot point because of the second, but just so you know, it's not quite true that the variable disappears when the script ends. If the script reports a procedure (made with a gray ring) that uses the variable, the variable lasts as long as something points to that procedure. This is how we do local state in OOP.

1. I didn't edit it, I accidentally clicked commit instead.
2. When I right-click and press help, nothing appears.
3. OK

brianharvey:
That's weird; it works for me:
Screen Shot 2021-11-12 at 3 43 57 PM

PR #2911, opened 2021-11-12
Adding help image for INDEX OF E]

joecooldoo:
This adds my help image for the INDEX OF block


brianharvey:
This is better -- what it says is correct.

I hope you won't feel picked on, but I have some suggestions for changes:

1.  The format (full-width blue bar, then full-width gray bar) doesn't match the other help files.  That's why I suggested starting with the two blank help screens I sent before; that's how I make them, by copying one of the blanks and then filling it in.

(By the way, simple help screens, like this one, don't have a blue rounded rectangle at the bottom; they just fit in the gray rounded rectangle.  The blue one is for details, special cases, etc.)

2.  The comment you put on the SAY block should just say "SAY 4" -- as it is, you don't show the user what INDEX reports.

joecooldoo:
> This is better -- what it says is correct.
> 
> I hope you won't feel picked on, but I have some suggestions for changes:
> 
> 1. The format (full-width blue bar, then full-width gray bar) doesn't match the other help files.  That's why I suggested starting with the two blank help screens I sent before; that's how I make them, by copying one of the blanks and then filling it in.
> 
> (By the way, simple help screens, like this one, don't have a blue rounded rectangle at the bottom; they just fit in the gray rounded rectangle. The blue one is for details, special cases, etc.)
> 
> 1. The comment you put on the SAY block should just say "SAY 4" -- as it is, you don't show the user what INDEX reports.

Fixed.

joecooldoo:
BTW, I don't feel like I'm being picked on. Your just a really nice guy. :)

brianharvey:
Wonderful, thanks!

joecooldoo:
NP!

PR #2913, opened 2021-11-15
Help page for APPEND

joecooldoo:
I created the help page for APPEND.


brianharvey:
Your text suggests that the first input is _modified_ to include the items of the second list.  It should say "creates a new list containing..."

Your huge font doesn't leave much room for refinements.  This is a case where I'd use the two rounded rectanges, gray and blue; in the second one I'd compare the result of APPEND [a b c] [1 2 3] with the result of LIST [a b c] [1 2 3].

joecooldoo:
> Your text suggests that the first input is modified to include the items of the second list. It should say "creates a new list containing..."

It says 

> any items in the second list will be added
to the end of the first list.

I don't see why this isn't clear enough.

> Your huge font doesn't leave much room for refinements.

I do have to agree: It is big. I will resize it.

> This is a case where I'd use the two rounded rectanges, gray and blue; in the second one I'd compare the result of APPEND [a b c] [1 2 3] with the result of LIST [a b c] [1 2 3].

I will add this to the help screen.

brianharvey:
"Added to the end" suggests that after doing the APPEND, the first list has changed to include the items of the second.  To prevent that misunderstanding, you have to say "a new list" or equivalent.

I was going to show you how the other list reporter help screens do that, but it turns out only KEEP does it right; I should fix all the others.

joecooldoo:
I have a problem: result pic for the comparison takes up a lot of space and is hard to read when it is small. What will I do?

joecooldoo:
How is this:

>"Append" means to "add (something) as an attachment or 
supplement". This means a new list with the first list copied to it
will be reported with the items in the second list joined to it.

joecooldoo:
OK, I made changes. How is it now?

brianharvey:
Oh, I didn't make myself clear.  What I want to compare are these:
untitled script pic

untitled script pic (1)
The idea is to show how APPEND is different from LIST, even though both can combine multiple lists.  In a tutorial rather than a help screen, I'd also include
untitled script pic (2)
Those three are the primitive list combiners.

I don't know why you're so tied to using asymmetrical language to describe the role played by the two inputs.  Why not just "APPEND reports a new list containing the items from the input lists, in order"?  Especially because it can have more than two inputs. :)  I guess the problem is that you looked it up in the dictionary. :)

maryfries:
I don't contribute to this repo, so this is just a helpful suggestion, but I recommend uploading an editable original file along with the snap-ready image file *for all help messages* so that they can be tweaked in the future without needing to be re-written from scratch. :)

joecooldoo:
> this is just a helpful suggestion, but I recommend uploading an editable original file along with the snap-ready image file for all help messages 

I have them all saved to my computer, ready to be edited at any time...

joecooldoo:
I know APPEND can take more than 2 inputs, I don't know why I described it like that. I use your suggestions and recreate the help screen.

joecooldoo:
@brianharvey I have updated the help screens once again with your suggestions.

joecooldoo:
@brianharvey?

brianharvey:
Getting there, just a couple of details.

I would remove the short line to the right of the number "1." in the comparison section.

And I'd change "1" and "2" to "append" and "list" respectively.

Thanks.

brianharvey:
> I have them all saved to my computer, ready to be edited at any time...

I think Mary's point is that someone other than you might want to edit them at some future time. :)

joecooldoo:
@brianharvey If you want, I can send a Scratch project with the SVG files in them if you want.

joecooldoo:
Oh and I updated it again.

brianharvey (approved):
This is great, thanks!  I'll ask Jens to pull it.

joecooldoo:
Your welcome! Glad to be a help! :)

PR #2921, opened 2021-12-04
UCB ACE Lab Encoded Script Pics Edits

jesus-villalobos:
Changes to `BlockMorph.prototype.userMenu` definition are provided in `UCBChangesToScriptPics.js` and changes to the `readImage()` method of `HandMorph.prototype.processDrop` are provided in `UCBChangesToMorphic.js`. New definitions are credited to Snap! Forum user @ DarDoro. 

Potential security issues: Processing arbitrary XML that is presumed to be Block definitions.


jmoenig:
Thank you so much, Jesus! This is really, really nice. I'm looking at it now...

jmoenig:
I see that @DarDoro has been involved, that's awesome!

UCBChangesToScriptPics.js
```diff
@@ -0,0 +1,486 @@
...
+					return int2BStr(data.length) + res + int2BStr(crc32(res));
+				};
+				function SaveData(dataURL, name) { 
+					return Object.assign(document.createElement("a"), 
+					{download: name, href: dataURL}).click(); 
+				};
+					
+				var ide = this.parentThatIsA(IDE_Morph),
+					top = this.definition || this.topBlock();
+				let xml = unescape(encodeURIComponent(ide.serializer.serialize(top))),
```
jmoenig, line 91: what happens to dependencies referenced in the script, i.e. custom blocks? This is a loose end on my side, so I'm afraid that it'll cause red "undefined" blocks when folks file it in.

UCBChangesToMorphic.js
```diff
@@ -0,0 +1,296 @@
...
+            file = event.dataTransfer?.files?.[0],
+            menu = new MenuMorph(this, "Process Dropped Object");
+
+        pic.onload = () => {
+            canvas = newCanvas(new Point(pic.width, pic.height), true);
+            canvas.getContext('2d').drawImage(pic, 0, 0);
+
+            (async () => {
+                if (!file && url) {
+                    file = await fetch("https://api.allorigins.win/raw?url="+url)
```
jmoenig, line 83: I'm little nervous about hard-coding the url of someone else's server into the function that handles loading graphics. What if the server is down? Would it make sense to add a fallback option in that case, or - forgive my ignorance - is that already handled by the async nature of this?

UCBChangesToScriptPics.js
```diff
@@ -0,0 +1,486 @@
...
+    );
+
+    /* UCB Script Pic Edit:
+     *  - Incorporated new method to userMenu to integrate scriptPic's and XML code.
+     *  - Code credited to Snap! Forum user Dardoro.
+     */
+    if (this.isTemplate || !this.definition) {
+		menu.addLine();
+		menu.addItem("script pic and xml...", 
+			() => {
```
jmoenig, line 62: this is really nice! I would be more comfortable if this could be factored out into its own method rather than inlining it all into the code that represents the menu. We might even wish to reuse it for other image exports, right? I mean, we could export a sprite *as* picture and include all its code (and even other costumes, hehe!)

UCBChangesToScriptPics.js
```diff
@@ -0,0 +1,486 @@
...
+    menu.addItem(
+        "help...",
+        'showHelp'
+    );
+
+    /* UCB Script Pic Edit:
+     *  - Incorporated new method to userMenu to integrate scriptPic's and XML code.
+     *  - Code credited to Snap! Forum user Dardoro.
+     */
+    if (this.isTemplate || !this.definition) {
```
jmoenig, line 59: what's this condition about? You want this option only to be available for (top) blocks that are templates, i.e. sitting in the palette or primitives. I don't understand the first constraint. The second one (only primitives) seems to aim for preventing custom block without their definition, and that's a good idea, but here you're only looking at the script's very top block. There still could be custom blocks further down inside the script, right? Would it make sense to scan the whole script for the existence of a custom block?

DarDoro: If I may ... 
Latest version was 
`if( !this.isTemplate || this.definition){//for scripts & custom blocks only`

Intentions explained in the forum [post](https://forum.snap.berkeley.edu/t/enable-image-webp-format-behind-shift-property-to-store-costumes/8083/27).

> Despite its image, the custom block should be exported from the palette, NOT from the script editor.
> During a block editing session, editor content is wrongly serialized as a script:
> 
> `<script>...`
> 
> Custom block exported from the palette has a proper form.
> 
> `<blocks app="Snap! 6, https://snap.berkeley.edu" version="1">...`

jmoenig: Ah, thanks, that makes sense, yes. So, we probably should be exporting not the serialized script but the / a custom block definition from the block editor. Good point.

UCBChangesToMorphic.js
cycomachead: AFAIK, this is a way of getting around CORS limitations. I would think it's OK to just remove this and raise an error if a file cannot be read.

DarDoro: It was primarily intended for drag&drop images from Discourse posts, served by CDN. So depending on the expected use cases may be not necessary (or CORS headers can be sent to allow snap's domains).

UCBChangesToScriptPics.js
jesus-villalobos: I will definitely look into making this its own method so we can take care of any image exports! Which Morph do you think it would make the most sense to make this a prototype of since the method would have to work for multiple?

UCBChangesToScriptPics.js
jesus-villalobos: Thanks @DarDoro for offering the clarification :)

p.s. sorry for not tagging you in the original PR, I wasn't sure if it was you :')

UCBChangesToScriptPics.js
brianharvey: Is there any reason not to include the xml in _all_ script pics?  That would reduce the cognitive load on users, and afaik doesn't make the png unreadable in other software.

jesus-villalobos: Yes I agree and that's definitely a possibility. I had it like that at first, but I thought I'd submit this PR consistent with @DarDoro's base project, though. If there are no objections, I also think that completely integrating it will be the best for user experience.

jmoenig: I'm not so sure about this. There are - and will be - many screenshots and pictures of Snap! scripts in all kinds of web sites and print materials that don't offer embedded sources. Therefore I like that @DarDoro visually indicates the presence of sources in the generated image (but I'd also like to reflect on whether it should be a technical term like <xml> or perhaps some other pictogram / icon, perhaps a lambda or a stylized block, and we'll probably also place it somewhere where it doesn't overlap with any of the blocks, i.e. we'd want to add transparent space ontop).

cycomachead: The original user case for this (i.e. curriculum authorship) sort of requires that we don't put a special label on images. We want them to look normal, since the use case is for authors, not students, to use the resulting scripts. 

Whether or not it should be everything is an interesting question. I am of two minds: On the one hand, including most scripts doesn't seem to hurt, but on the other hand we are going around turning a punch of static images into something semi-executable, at least inside Snap<em>!</em>. That's probably an OK trade off, but it's kind of like the dangers of allowing anyone to embed VB inside Word/Excel docs.

UCBChangesToScriptPics.js
cycomachead: I think it would be more appropriate to allow this to be done from either location. Though, there are complexities here. In an editor instance, we can look up the actual serialized definition and attach it to the script pic. We do need to make sure the block has been saved.

UCBChangesToScriptPics.js
```diff
@@ -0,0 +1,486 @@
...
+					return arr.reduce((res, byte) => res + String.fromCharCode(byte), '');
+				};
+				function int2BStr(val) { 
+					return arr2Str(Array.from(new Uint8Array(new Uint32Array( [val] ).buffer)).reverse());
+				};
+				function buildChunk(type , data) { 
+					let res = type + data; 
+					return int2BStr(data.length) + res + int2BStr(crc32(res));
+				};
+				function SaveData(dataURL, name) {
```
cycomachead, line 84: This should probably just be the `saveFileAs` method in IDE_Morph?

UCBChangesToScriptPics.js
brianharvey: The analog to malicious VB is a script pic with XML that doesn't match the picture.  I guess we _should_ worry about that, but once we _allow_ the PNG/XML combination, we've opened that door.  _Our_ script pics aren't going to be malicious in that way, so it really doesn't matter from a security standpoint if we _always_ make hybrid pictures or only _sometimes_ make them.  Bad guys will presumably take a PNG and add their own XML by hand.

But for the use case we have in mind, it's not necessary that we allow the XML to be run.  We have to allow it to be loaded, translated to another language, and exported as another script pic.  But we could invent the category of non-runnable script if that's what it takes to thwart bad guys.

cycomachead: Yeah, that's true -- i mean really what we should just have is a command line script that converts images and we could delay the option to load scripts in.

UCBChangesToScriptPics.js
jesus-villalobos: Yes, I hadn't thought about this, but referencing a custom block within a script does bring back an "undefined" block. To your point below, we will most likely have to scan for the existence of a custom block before serializing. I'll look into this more in depth.

UCBChangesToMorphic.js
```diff
@@ -0,0 +1,306 @@
...
+                    menu.addItem(
+                        "Interpret script pic as blocks.", 
+                        () => {
+                            if (blocks.includes("Custom\tBlock\tDefinitions")) {
+                                let customsSplit = blocks.split("Custom\tBlock\tDefinitions");
+                                await ide.droppedText(customsSplit[0], file.name, ""); // drop definitions on first
+                                await ide.droppedText(customsSplit[1], file.name, ""); // drop scripts on afterwards
+                            } else {
+                                ide.droppedText(blocks, file.name, "");
+                            }
```
jesus-villalobos, line 107: To be able to properly decode the XML inside the images, I tried two things. At first, I thought that simply concatenating the custom block definitions with other scripts would properly bring everything back. However, there were messages about missing block definitions. 

To try to get around that, I separated the definitions and scripts by a delimiter. The thought process behind this was that I'd be able to drop the custom block definitions onto the `ide` then drop the scripts afterwards so there would be no errors. However, I found that only the second drop is successful and the other is disregarded. Even after adding `async/await` calls to the function calls, only the second call succeeded.

Any thoughts on what to do?

jesus-villalobos: For example, the following would be embedded in the PNG below:
`<blocks app="Snap! 7dev, https://snap.berkeley.edu" version="2"><block-definition s="bar" type="command" category="other"><header></header><code></code><translations></translations><inputs></inputs></block-definition><block-definition s="foo" type="command" category="other"><header></header><code></code><translations></translations><inputs></inputs></block-definition></blocks>Custom	Block	Definitions<script><block s="setHeading"><l>90</l></block><custom-block s="bar"></custom-block><block s="changeXPosition"><l>10</l></block><custom-block s="foo"></custom-block></script>`

I left the definitions empty for simplicity, but the same applies with blocks within them.

setHeading (19)

DarDoro: I've got mediocre success with this direct synchronous import
```
ide.rawOpenBlocksString( customsSplit[0], file.name, true); // import custom blocks
ide.openScriptString( customsSplit[1], file.name, false); // drop scripts on afterwards

```
Custom blocks are imported OK,  but `"<script>"` import seems to be flawed, in this version, regardless of the source - XML or PNG.

jesus-villalobos: Thank you for the functions! I just added them in a new commit. I don't have an idea of how to get around the `Undefined` blocks, but @jmoenig @brianharvey @cycomachead if you have any ideas they'd be greatly appreciated :)

DarDoro: Makeshift workaround, till import of the script will be fixed in Snap! 
```
delete ide.serializer.scene.stage;//workaround, force loading script in "ide.stage" context
ide.openScriptString(customsSplit[1], file.name, false);
```

jesus-villalobos:
Hello everyone! I added the new changes that @DarDoro suggested, and it seems now that custom blocks and scripts are saving and returning as we expect. Any comments, feedback or next steps are appreciated! Thanks :)

jmoenig:
Thanks, @jesus-villalobos ! Hey, how about we schedule a video-chat to talk this trough? Can you perhaps let me / us know when would be a good time for you? Also feel free to just send an email (doesn't have to be in here).

jesus-villalobos:
Hey there @jmoenig! I know we're all probably in different time zones, but I'm free after 5:30pm PST on Wednesday and Thursday this week, so anytime that works for you all in there is good with me :)

jesus-villalobos:
Hello everyone! Just a quick note from a conversation with Dan: upon adding inputs to custom blocks, they are no longer properly exported ... he is good at finding bugs! haha! We found that choosing the `script pic and xml` option will export the script pic with the parameters, but the XML will not include them. Here's the script pic:

setHeading (3)

But, when the script pic was dropped back on the script window, the custom block was not loaded and the script produced this:

setHeading (4)

If any of you have an idea of what might fix this, any help or ideas are appreciated :)

jmoenig:
Good  catch. FYI I've started working on a general export/import with dependencies mechanism and format. It's - as you've already found out - a little tricky, but we'll work it out. When that's done we should be able to just use it for this script-pics feature.

jesus-villalobos:
Oh ok, yes a general import/export format will definitely make everything a lot smoother. 

On another note, Dan wanted me to explore exporting custom blocks from the script window since that's the only place where a single custom block cannot be exported as of now ... it must be connected to another block to export from the script window. I'll begin looking into it and seeing if I can generalize the current method for both exports. Nevertheless, Dan was definitely very excited when I presented the work to him earlier, and he's ready to use it for BJC!

jmoenig:
Right. I think once we have the generic export/import feature for *any* block it'll be trivial to export ... any ... block not just from the palette but also from elsewhere.

jesus-villalobos:
Perfect. That definitely sounds like it's the best way to go. 
Let me know if I can help along the way and I look forward to meeting with you to get things ready to go!

jmoenig:
yep, will do. I hoped to get this ready for v7.3, but it turned out to be trickier than I at first expected, but it's next up, and I'm working on this now. As I've got some other urgent thinks to take care of I can't yet say when I'll be done with this format, but let's just assume we'll have it shortly.

THEN we can can think about - and might want to already experiment with that - not just script pics containing  code, but also sprites that have not just one costume, but potentially more *and* code inside them. You'd see a PNG but you'd *get* a whole sprite including many costumes, sounds, (custom) blocks and scripts, which you could easily share. What do you think?

brianharvey:
Do you mean exporting a sprite into metadata of a png?  I can imagine that could be confusing, if someone thinks they're _importing_ just a costume and it turns out to do all this other stuff.  So maybe the png should be visually distinguishable somehow, like, I dunno, an Alonzo watermark?

jmoenig:
Oh sure, I'm all for indicating that, yes!
But yes, the idea to explore where else we could use this code-embedding trick outside of script pics. For example, wouldn't it be neat if we could have little animations, GIFs or even movies, and they could include the whole project?

jesus-villalobos:
That definitely sounds like a great thing that we might want to look into. I believe that you all (Jens, Brian, Dan and Michael) had talked about exporting sprites with all their metadata, costumes and whatnot, so I think that adding blocks and scripts wouldn't be too far off. I'll begin looking into it.

Brian, I think that we could do something similar to what we did for the script pics with embedded XML ... upon their drop and data extraction, we can give users the option to choose what parts of the embedded data they want to use. 

In either case, I definitely agree with Jens that the whole embedding has many potential pros.

brianharvey:
Do gifs allow for arbitrary metadata the way pngs do?

jesus-villalobos:
According to the [W3 Spec for GIFs](https://www.w3.org/Graphics/GIF/spec-gif89a.txt), the underlying block format of GIFs allows for "special purpose" plain text blocks that do not affect the rendering of the GIF itself: 

> Special Purpose blocks such as the Comment Extension and the Application Extension are neither used to control the process of the Data Stream nor do they contain information or data used to render a graphic on the display device.

As for how large these blocks can be, it states that there can be multiple blocks of size 0-255 bytes:

> any number of [comment extension blocks] may appear in the Data Stream ... [each block is a] sequence of sub-blocks, each of size at most 255 bytes and at least 1 byte, with the size in a byte preceding the data.

This is what I gathered from the documentation, so though there is a way, I'm not sure how deterministic the final positioning of the blocks is. @DarDoro might be better off answering this question. Any thoughts @DarDoro?

jmoenig:
there are also animated PNGs (APNG)

DarDoro:
PNG and WEBP are natively supported by the canvas/blob api. PNG is a loosely coupled stream of self-contained data chunks.
EXIF with XML data is literally injected between other chunks with a few lines of code.

Encoding GIF requires extra libraries. There was already a project in Snap! to record the screen as an animated GIF. So it's doable, no doubt. 

As for the size constraint, I think it's implementation details, not formally declared.

The idea of the whole sprite in GIF/PNG seems appealing at first, but...
The sprite has costumes. So we will have an image containing XML, containing BASE64 encoded PNG's...
Something like subresource URI to cloud stored object, seems to be better suited for very complex scenarios.

DarDoro:
> If any of you have an idea of what might fix this, any help or ideas are appreciated :)
> 
> @ jmoenig @ DarDoro

While Jens is working on more generic solution this may help.
```
if (block.blockSpec() == custom) {
    customBlocksString += ide.serializer.serialize( block);
} 
```
The surprising benefit is that block body is exported 🤣

jmoenig:
Okay, I've implemented a more general way to encode scripts into shareable XML that includes all the custom block definitions used in a script along with any custom palettes. With a call to

`BlockMorph >> toXMLString()`

you should be getting XML that you can stick into a script pid, and loading that XML into another project will give you exactly that same script again, in that target sprite. It'll include all global and local custom blocks (and overload existing same-spec custom blocks).

Why don't you give it a try, and let me know what you find. Then let's schedule a video chat to talk about how we're actually going to integrate this feature.

Now, onto doing the same for sprites....

jesus-villalobos:
This is amazing @jmoenig!! I just met with @ddgarcia about the new progress and the new changes that you provided us with and he is excited! I will refactor the code, and if I run into any problems I will let you know.

Thank you so much!!

jmoenig:
Great! I've worked some more on this, and have generalized the `toXMLString()` api some more, extending it to arbitrary scripts (so can also programmatically serialize scripts inside sprites that are not currently edited in the IDE) and also to sprites.

For scripts:

1) Determine the script's `topBlock()` (otherwise you'll serialize whichever block you call this on, including the ones attached to its bottom)
2) On the `topBlock()` call `toXMLString(receiver)`, where the `receiver` parameter is the sprite that this script is part of. The `receiver` input can be left blank if the sprite is currently edited in the IDE, which is usually the case, because otherwise the user can't right-click on the script in the first place. So you'll usually not need to pass the receiver into the `toXMLString()` function call. However, at some point in the future we might also want to support saving / exporting scripts directly from speech bubbles and variable watchers, and *then* their receiver sprite will not be the active one. For those cases the parameter will do the trick.

for sprites:

Simply call `sprite.toXMLString()` to get a full xml that you can export, stick into a PNG, and re-import into (another) Snap. the new sprite export format likewise now contains all dependencies (i.e. all parts attached to the sprite in the case of nested sprites, and all global custom blocks referenced by any of the included sprites, as well as all custom palette categories)

I'm excited what y'all will make with this!

jesus-villalobos:
@jmoenig sorry for the delayed response, but thank you so much for the new methods!! I got rid of the recursive finder for custom block definitions, and it is definitely much cleaner now. I also experimented with a new condition to differentiate between custom blocks, after noticing that one could not request `script pic and xml...` from the IDE when `top` is a custom block. 

If this isn't something that you're all interested in pursuing, please feel free to let me know, but if you think it's worth it I will need some help to get it moving. Since I'm unable to use your new `toXMLString()` on a custom block, and `ide.serializer.serialize` doesn't return the proper XML, I'm at a standstill. 

Let me know what you think!!

jmoenig:
hi @jesus-villalobos I don't understand your issue with scripts whose top-block is a custom block, they export and import just fine for me. Could you share an example with me? Alternatively / additionally I'm ready to chat with y'all on Zoom or wherever you like. Thanks!

jesus-villalobos:
hey @jmoenig! Sorry, I realized that I had the wrong condition for custom blocks, but it's all fixed now :) I know Dan wanted me to include an option for users to be able to get a `definition pic and xml...` for custom blocks, both from the IDE and the selection panel on the left, so I included that as well. We thought that it was good because, it allows for users to either get a pic of the custom block itself, or a pic of the definition. For example:

foo (6)
and
bar (4)

or

foo (4)
and
bar (5)

If you think this is unnecessary, feel free to let me know, and we can modify or remove it altogether.

jesus-villalobos:
I think the only thing left to do is make `injectScriptPic()` function a method of a class. Do you think that it would make the most sense to make it a method of the `ScriptMorph` class?

Also, other than that, I'm also ready to meet whenever! I know the time difference from California -> Germany is weird, but as you may have noticed by now, I work in the wee hours of the night 😅. So feel free to let me know when works best!

jmoenig:
Excellent, great progress @jesus-villalobos ! Ugh, I realize that exporting a "definition script" will require some tweaking that I didn't think of before, was that the issue? Do you need me to add functionality for this?
If you're *really* up late we could chat my early morning. I can usually make it at around 7 AM my time, which would be 10 PM yours, if that's okay with you. Alternatively we I could also make it in the late afternoon my time, around 6 PM, which would be early morning (8 AM) yours. Does any of that work for you, and if so, which day would be best for you?

jesus-villalobos:
Thanks @jmoenig! And no worries ... it took me a while to figure out how to make it clean, but in the end, it's just a check for `this.isCustomBlock` that determines whether or not the block can supply a `definition pic`. I like the current format because it allows users to export both `script pics` and `definition pics` from _either_ the script window or the selection panel on the left. I don't think that additional functionality is necessary, but if you think that you can do something that eliminates that additional check and maintains the same behavior then that would be nice. 

And, yes, as a night-owl a late meeting for me, early for you, would be ideal. I can make any day this week work, so that would be your Thursday or Friday morning, so if one is better for you, let me know and I can send a zoom link.

jmoenig:
Great! How about my Friday morning / your Thursday night?

jesus-villalobos:
Sounds great! I just sent an invite right now, so let me know if you got it! :)

jmoenig:
got it, thanks! Can we shift it by one hour, though @jesus-villalobos? I need to drive my kid to school at 7 AM my time, but I'll be back at 8 AM. Could that still work for you? Thanks!

jesus-villalobos:
@jmoenig I pulled the changes, and it looks amazing! I'm excited to share the results with Dan!! I know there still might be bugs, so I'll try to see if I can find any as well as trying to get the block definition pics. Let me know if there's anything else that I can help with!! We definitely want to get this ready to go!

jesus-villalobos:
@jmoenig I'm definitely happy to see this feature begin it's path into production! And, definitely, a very big thank you and lots of words of appreciation to @DarDoro! His base project really kickstarted this whole thing!

Let me know if there's anything that I can do!

jmoenig:
Hey, btw, did you see the "label of block" feature I've added? This should totally help you with the autograding project. Lemme know if you need pointers...

jesus-villalobos:
@jmoenig I actually did not notice that yet. Can you give me pointers to what/where this feature is?

jmoenig:
BTW, I can't get over how cool it is that you can now right-click on these pictures, download them to your computer and drag them into the new Snap! to play with the actual blocks!!!! Thank you!!

jesus-villalobos:
@jmoenig this is really cool! I downloaded your scripts, and EASILY dropped them onto the window and began interacting right away 😄!!! This will definitely help us make autograding a reality soon!! Thanks so much!

PR #2927, opened 2021-12-10
Updating Catalan translation for v7

jguille2:
Just this!
Updating Catalan translation before v7 release!

Thanks,
Joan

### DIFF
--- a/locale/lang-ca.js
+++ b/locale/lang-ca.js
@@ -641,7 +641,7 @@ SnapTranslator.dict.ca = {
         'definició',
     'custom?':
         'personalitzat?',
-    'global':
+    'global?':
         'global?',
 
     // operators:
--- a/locale/lang-ca.js
+++ b/locale/lang-ca.js
@@ -185,7 +185,7 @@ SnapTranslator.dict.ca = {
     'translator_e-mail':
         'bernat@snap4arduino.rocks, jguille2@xtec.cat', // optional
     'last_changed':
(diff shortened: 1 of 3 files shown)


WarpedWartWars:
https://github.com/jguille2/Snap--Build-Your-Own-Blocks/blob/130a74fd24d9730820d76d1837159d2f5ca6b18a/locale/lang-ca.js#L644-L645

Shouldn't "global" there have a question mark?

jguille2:
Yes, you're right! Thanks.

I copied that definition from German translation
(https://github.com/jmoenig/Snap/commit/54d87b52251208bc905d0aa5b4c7de0caa621913#diff-1d798b72c72c550db90220b4fff057628e94940569e6093d3c73bd40e16972adR734)

Interestingly, in both German and Catalan, the word is the same as in English .... so the error remains only in the code :)

I will correct it now.

Joan

jguille2:
Jens, I've fixed that 'global' ->'global?'

(https://github.com/jmoenig/Snap/blob/ecbefe7e4275d946ce551c35b139e97b3b88f8df/src/blocks.js#L804)

also on German translation (revert this if you don't agree... but I think it was good... I haven't changed 'last_changed' info... because it was a 'typo'... but maybe I'm wrong)

jmoenig:
Ah yes, thank you, that was another "last-minute" afterthought :D

jmoenig:
will pull later today, tnx

Issue #2928, opened 2021-12-10
Library names translations


jguille2:
Hi Jens,

Importing libraries... library names are not translated since 
https://github.com/jmoenig/Snap/commit/d33d4c0162cab7e3ae1727106fcbd8c3147a91ce#diff-6dcd1cb9e0e8e3ac976d0efe29c2c79a8b80d391c6bd96af7c2100f618544960R10855

The sequence is:
```
(gui.js)
new LibraryImportDialogMorph
LibraryImportDialogMorph.prototype.buildContents
LibraryImportDialogMorph.prototype.installLibrariesList
new ListMorph
(morphic.js)
ListMorph.prototype.init
ListMorph.prototype.buildListContents
this.listContents.addItem(
    label, // label string
    element, // action
    null, // hint
    color,
    bold,
    italic,
    this.doubleClickAction,
    null, // shortcut
    true // verbatim - don't translate
);
```
The "problem" is
- *listContents* has *verbatim* parameter
- but *ListMorph.prototype.ini* has not a *verbatim* property
- and then, *installLibrariesList* can't use that parameter

My suggestion, to avoid conflicts with other functions...
- Add that *verbatim* param to ListMorph (default *true*)
- And use it in *buildListContents*
- Then we can use "verbatim=false" in *installLibrariesList* keeping current behavior for all other calls.

Jens, I can type this and submit a PR... but maybe you prefer other ways... or maybe you don't want to do it now (I know we are in  a RC state).

Tell me what you want...

Joan

jmoenig:
Hi Joan, thank you! Sure, go ahead and issue a PR, this sounds reasonable. Are you sure it doesn't affect other parts of the UI?

jguille2:
Ok Jens, I'll do it later...

About other affectations... It's the reason I suggest to call that "verbatim=false" only into *installLibrariesList*

jmoenig:
sounds good! Thank you.

PR #2929, opened 2021-12-10
Fixing translation issue of library names

jguille2:
Fixing #2928
-  Adding *verbatim* param to ListMorph (default *true*)
- *ListMorph.buildListContents* function use now this *verbatim* param to create the list (in *listContents.addItem* action)
- And then, *installLibrariesList* use it with *verbatim = false" to allow translations


jmoenig:
thank you, Joan, will pull later today!

PR #2951, opened 2021-12-29
Api fix

zakkolar:
Makes API method `IDE_Morph.prototype.broadcast()` work with new `message` upvar in `when I receive [any message]` block (bug described [here](https://forum.snap.berkeley.edu/t/api-broadcast-any-bug/9632)).

May need extra testing to make sure `ide.broadcast()` with a callback works as expected.


jmoenig:
thanks, Zak, great catch and fix!

PR #2952, opened 2021-12-29
Make SymbolMorphs match height of input in blocks.

opencoca:
This makes blocks containing BlockSymbolMorph parts match the height of blocks with input(s) . 

TODO increase height of blocks containing only BlockLabelMorphs to match.


jmoenig:
can you elaborate a little bit what this PR addresses, perhaps with a screenshot before and after? seems like the only change is an adjustment factor....

opencoca:
Yes it's a very minor tweak.  It standardizes the height of blocks.  

Please disregard the silly grey block name in the gif it's unimportant.

Block_morph_height

As you can see in the animation this pull tweaks things so that the blocks with SymbolMorphs and the blocks with inputs are of the same height.  I'm still trying to decide how to best standardize the size of blocks containing only BlockLabelMorphs.

jmoenig:
Thank you, but the block height is intentionally variable, e.g. if you nest reporters in input slots, that will als increase the height. Thanks for the code, but I'll close this, because ... the current behavior is exactly how I designed it.

PR #2959, opened 2022-01-05
Update strings.xml

brianharvey:
substring of: from: to:
didn't handle negative indices as advertised.


jmoenig:
excellent, thank you!
Will pull for the next patch

PR #2960, opened 2022-01-05
Catalan libraries translations

jguille2:
Hi!
Just translating libraries to Catalan

Joan


jmoenig:
Thanks, Joan!
Seems like @brianharvey's fix to the strings library got in the way of this huge translation effort. Would you like to apply your translation to Brian's new version (afaik only the `substring` block is affected), or do you want me to do it?

jguille2:
Thanks Jens,

I hadn't seen that last change.

I've updated strings.xml with that last change and I think this is ok now.

Joan

PR #2970, opened 2022-01-24
Update textCostumes_module.xml

brianharvey:
Fix bug from forum about numbers as names (yeah I know...), save and restore costume (not number) and pen size, make the name of a button saying foo to "foo button".


jmoenig:
thanks, I'll pull it. But I'm not okay with the name change to "... button", because "button" is English, and *not* what the user specified, so I'll revert that change.

To be pulled later this week for this week's release patch

PR #2976, opened 2022-02-03
Fix continuations only running the last block of their parent contexts

### LINKED ISSUE #2969: Continuations act in mysterious ways
really broken continuations

If Snap!'s continuations were undelimited, as I originally assumed, this script would say 1, and then it would say 2, 3, adds b, b2 and c to the list on loop, as the continuation is invoked, returning execution to a previous point over and over.

However, this is not the case. What actually happens is that they are... "sort of" delimited. The script says 1, 2, 3, 2, 3... adds b, b2, c - then, on the second go-around, it decides to _skip over_ adding b, only adding b2, and then adding d as if the continuation returned like a procedure.

I would get if the script just said 2 and 3 on the go-around without adding b2 again. That would mean these continuations are delimited by the end of the procedure, and I suppose that makes sense. But this behaviour of continuations seems completely unintuitive to me, and I can't seem to make sense of it! Is this behaviour intentional?

rdococ:
Should fix #2969

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -7080,11 +7080,11 @@ Context.prototype.copyForContinuation = function () {
             isString(this.expression));
     if (isReporter) {
         cur.prepareContinuationForBinding();
-        while (cur.parentContext) {
-            cur.parentContext = copy(cur.parentContext);
-            cur = cur.parentContext;
-            cur.inputs = [];
-        }
+    }
+    while (cur.parentContext) {
+        cur.parentContext = copy(cur.parentContext);
+        cur = cur.parentContext;
+        cur.inputs = [];
     }
     return cpy;
 };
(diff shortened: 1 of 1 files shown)


jmoenig:
I'm sorry, but I won't pull this, because it breaks the way continuations are supposed to work in Snap!.

Here's a test case which is currently not handled gracefully by Snap's continuations, which your PR addresses and "fixes":

Test1

However, removing all input state from continuations results in other tests no longer validating, especially this one:

Test2

Here, re-evaluating the continuation should count up in steps of 5, which it currently does. Applying your suggestion of removing all input state no longer correctly captures the iteration inside a loop and results in counting up in steps of 10 instead.

Snap's continuations are a limited, experimental feature, basically the accepted and agree-upon test cases can be found in this project:

https://snap.berkeley.edu/snap/snap.html#present:Username=jens&ProjectName=continuations&editMode

PR #2990, opened 2022-02-27
JSCompiler stuff

xBZZZZ:
1. use `new Map()` instead of `{}` for gensyms 
2. use `arguments[arguments.length - 1].selector(arg0, arg1, arg2)` instead of `arguments[arguments.length - 1].selector.apply(arguments[arguments.length - 1], [arg0, arg1, arg2])` (that was compiled example, not actual code)
3. escape in `value instanceof Array`
4. `// fix for blocks like (x position)`
5. new `escape` function that makes sure `string` is a string and escapes weird unicode chars but doesn't escape `'` because compiler always puts escaped string in double quotes (`"`)

Also, how do you get `Array` object in `Snap!`?

Tell me what I did wrong if I did something wrong.

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -7728,7 +7728,7 @@ JSCompiler.prototype.escape = function(string) {
     while (len > i) {
         char = string.charAt(i++);
         if (" abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!#$%&'()*+,-./:;<=>?@[]^_`{|}~".indexOf(char) === -1) {
-            escaped += "\\u" + (char.charCodeAt(0) | 0x10000).toString(16).substring(1);
+            escaped += '\\u' + (char.charCodeAt(0) | 0x10000).toString(16).substring(1);
         } else {
             escaped += char;
         }
--- a/src/threads.js
+++ b/src/threads.js
@@ -7657,7 +7657,7 @@ JSCompiler.prototype.compileInputs = function (array) {
 };
 
 JSCompiler.prototype.compileInput = function (inp) {
-     var value, type;
+    var value, type;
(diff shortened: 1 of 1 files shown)


brianharvey:
To answer your question:   You use LIST.  See https://forum.snap.berkeley.edu/t/i-copied-a-list-to-another-variable-and-then-when-i-change-the-copy-it-changes-the-original-too/896

jmoenig:
Ao, Brian, that wasn't his question. The JIT-Compiler has special provisions for handling JS Array objects with strings inside them, because those need to be escaped, and he's asking where in Snap those Arrays are created. It's a technical question. The answer is: When a dropdown menu has a translatable string, that string is handled as an Array around it. this also means, that users cannot make their own ones, which in turn means they can't use such Arrays to trick the JIT compiler into evaluating arbitrary Javascript. So, *that* part is not a security risk, I think.

xBZZZZ:
> this also means, that users cannot make their own ones

But they can:
untitled script pic

jmoenig:
Ah, you're right, of course, using the new meta-programming features. Gosh, meta-programming is the source of all evil, I knew it. Thanks!

xBZZZZ:
> Ah, you're right, of course, using the new meta-programming features. Gosh, meta-programming is the source of all evil, I knew it. Thanks!

You also can just edit XML project file.

jmoenig:
Yep, that's also true.
Your PR looks all good to me, I didn't yet test it, but from what I've seen it appears you totally know what you're doing here, thank you!
I'll pull this for the next patch / minor release. Currently working on some other stuff that I want to go into that next release, so It may take another couple of days for me to get around to it. Please don't think I'm forgetting this, and again: Thank your for your contributions!.

PR #2996, opened 2022-03-02
Change browser tab title when project title changes.

gigamonkey:
(I looked at the contributing guidelines and saw the thing about changesets but also looked at the recent pull requests against your repo and it looks like at least some of them are normal PRs that make small changes to existing files. Happy to try to do this some other way if you want me to but this seems simplest to me.)

Anyway, I got tired of not being able to find the right Snap tab in my browser because they're all titled the same thing, so I cooked up this change to set the `document.title` when the project changes (i.e. on saves and opens).

I've tested this in local development but I assume I can't save to the cloud safely so I didn't test that. But if I've found the right code path to do this in it should change whenever the title in the UI changes.

Hoping you'll find this useful as it will make my Snap! programming that much more pleasant.

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -1258,6 +1258,9 @@ IDE_Morph.prototype.createControlBar = function () {
         scene = myself.scenes.at(1) !== myself.scene ?
                 ' (' + myself.scene.name + ')' : '';
         name = (myself.getProjectName() || localize('untitled'));
+        if (name !== null && name !== 'untitled') {
+            document.title = "Snap! " + name;
+        }
         txt = new StringMorph(
             prefix + name +  scene + suffix,
             14,


jmoenig:
Wonderful, will pull later today for the next patch, thank you! I agree it's super useful, I'm going to love this myself.

Oh, geez, did I not remove the part about changesets yet? Haha, you can apply some software-archeology to it and detect my Smalltalk roots 😄 

The next patch in which this change is going to become active might take some more days, or even weeks this time (I don't hope so, but it's possible). For the next minor release we're discussing changing some basic associative arithmetic primitives to be variadic instead of dyadic. As you might see in the recent commits the logic and code is mostly there, but it's the design we need to discuss and agree on, especially in the light of existing curriculum, courses and other materials. So, if it seems that not much will be happening in this repo, it's because we're busy talking to each other about the recent ones.

cycomachead:
Very nice, I've been meaning to do this for a while. 👍

PR #2997, opened 2022-03-03
Update animation_module.xml

brianharvey:
Fix GLIDE not to find the destination coordinates with the pen down.  Per forum.


jmoenig:
Thanks!
Alternatively we could probably determine the target coordinates using math, if only I knew how to, haha!
Will pull for next patch.

PR #2999, opened 2022-03-07
Add a working tts library for Chrome OS

TipsyTheCat:
Simple JS tts api implementation that actually works on chrome browsers.
Done by using the built-in tts api and adding `window.speechSynthesis.cancel()` before speaking the message, which fixes tts not working.


jmoenig:
Umm, can you tell us what this is about, in particular how it's different from the existing "text to speech" library?

TipsyTheCat:
Ever since [this](https://chromestatus.com/feature/5687444770914304), most (if not all) chrome users have been unable to invoke JavaScripts tts api using 
```
var msg = new SpeechSynthesisUtterance();
msg.text = "Hello world";
window.speechSynthesis.speak(msg);
```
But, by canceling using `window.speechSynthesis.cancel()` allows speech to properly start.
example:
```
var msg = new SpeechSynthesisUtterance();
msg.text = "Hello world";
window.speechSynthesis.cancel()
window.speechSynthesis.speak(msg);
```

jmoenig:
Hmm... I've just tried the existing library in Chrome 99 and it works just fine...

jmoenig:
no, in Chrome on a Mac and also in Windows.
Okay, I've read up on what you pointed out, and it seems that some user interaction with a page is required before tts will work. But that's literally the case when you click the green flag button to start a project, isn't it? I don't see how calling `cancel()` would circumvent the autoplay deprecation. Can you tell me more about this? How did you get the idea to call `cancel()`.

TipsyTheCat:
Personally, I have not idea why it doesn't work or why doing this _does_ work... [This](https://1drv.ms/v/s!AvAOlKcZY4gtiOspT8NyYkd_kikaMw?e=nR2s0G) is a screen recording of using tts with and without `cancel()`

TipsyTheCat:
Also, pardon my unclearness...

TipsyTheCat:
Ok, so I noticed that on first run of the block it doesn't speak. I'm going to convert this PR into a draft to figure out how to fix this...

PR #3003, opened 2022-03-10
Hide Hidden Blocks, Show Categories In libraries Viewer

cycomachead:
This brings the Libraries viewer into 7.0-land, and shows the category labels in the preview window, like would be shown in the single palette. It also hides the blocks which are hidden in the library when importing.


src/gui.js
```diff
@@ -8995,30 +8995,31 @@ LibraryImportDialogMorph.prototype.installLibrariesList = function () {
...
+            this.displayBlocks(fileName);
         } else {
-            this.showMessage(localize('Loading') + '\n' + localize(item.name));
+            this.showMessage(`${localize('Loading')}\n${localize(name)}`);
             this.ide.getURL(
-                this.ide.resourceURL('libraries', item.fileName),
+                this.ide.resourceURL('libraries', fileName),
                 libraryXML => {
+                    let serializer = this.ide.serializer,
+                        palette = serializer.parse(libraryXML).childNamed('palette');
```
cycomachead, line 9013: this suggests to me the serializer could use a `loadXMLModel` function which we could then pass to the load blocks, but I don't think it's a big deal...

jmoenig:
nice! thank you, I need to play with this, yay!!

src/gui.js
```diff
@@ -9147,41 +9148,42 @@ LibraryImportDialogMorph.prototype.hasCached = function (key) {
...
 LibraryImportDialogMorph.prototype.cachedLibrary = function (key) {
-    return this.libraryCache[key];
+    return this.libraryCache.get(key).blocks;
 };
 
+LibraryImportDialogMorph.prototype.cachedPalette = function (key) {
+    return this.libraryCache.get(key).palette;
+};
+
+
```
cycomachead, line 9163:
```suggestion
```

src/gui.js
```diff
@@ -9147,41 +9148,42 @@ LibraryImportDialogMorph.prototype.hasCached = function (key) {
...
+    var ide = this.ide,
         selectedLibrary = this.listField.selected.fileName,
         libraryName = this.listField.selected.name;
 
     // restore captured user-blocks categories
     SpriteMorph.prototype.customCategories = this.originalCategories;
 
-    /*
     if (this.hasCached(selectedLibrary)) {
+        console.log('Loading from cache: ', this.libraryCache)
```
cycomachead, line 9175:
```suggestion
```

cycomachead:
I was also thinking as I wrote this...

* We should have a `CheckboxPalette` object which we can use for importing and exporting, and hiding and showing blocks.
* In a library, now that we have a dependency resolved we could allow you to import  subset of the blocks. 


Nothing to stop the these fixes, just a thought. :)

jmoenig:
I'm sorry @cycomachead , I needed to fix the library preview because I changed the format of the blocks list to be an object with a "local" and a "global" key, because in the next release we can also export sprite-local custom blocks and individual scripts using them. If you find some time (no rush!) could you glance over your changes and see if they work with those changes? Thanks!

cycomachead:
@jmoenig OK this works with your changes now. :)

jmoenig:
Yay, thank you, @cycomachead !

src/gui.js
jmoenig: wait! I think I've added just that in the meantime, as I was revising libraries anyway...

jmoenig:
Oooh, this is really, really nice! Great work, thank you, Michael!!

PR #3009, opened 2022-03-15
JSCompiler stuff #2

xBZZZZ:
note: I was testing using secret cb block

1. untitled script pic now works
2. parameters now default to empty string
3. script variables now work
sv
note: JSCompiler declares variable when it compiles svv block but snap declares variable when svv block is executed:
e
4. se and c now work with script variables and parameters
note: compiled c block that changes script variable or parameter doesn't error when changing by not number
5. bypass calling `Process.prototype.reportNewList` and `Process.prototype.reportBoolean`

offtopic: why is `threads.js` formatted using both tabs and spaces?

### DIFF
--- a/src/threads.js
+++ b/src/threads.js
@@ -7504,9 +7504,10 @@ VariableFrame.prototype.allNames = function (upTo, includeHidden) {
 function JSCompiler(aProcess) {
 	this.process = aProcess;
 	this.source = null; // a context
- 	this.gensyms = null; // temp dictionary for parameter substitutions
+ 	this.gensyms = new Map(); // temp dictionary for parameter substitutions
   	this.implicitParams = null;
    	this.paramCount = null;
+    this.scriptVarCounter = null;
 }
 
 JSCompiler.prototype.toString = function () {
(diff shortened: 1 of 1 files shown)


jmoenig:
thank you, this all looks great to me, I'll pull it shortly!
Hehe, many people contribute to the repo, probably using different whitespace setting in their editors. Even on my different computers and OSes there are probably different settings for that.
Thank you for this really nice and very thoughtful work! The hidden "compile" block's purpose is exactly for this, to use it for debugging the JIT compiler. I appreciate your help.

jmoenig:
I'm sorry, I merged this and then was getting a bunch of errors, hence I've reverted it for now. Need to investigate into those errors. Hmm... can I reopen this somehow?

xBZZZZ:
> I merged this and then was getting a bunch of errors

Send screenshot of script than gives errors.

jmoenig:
okay, I'm adding your changes now, and then I'll try to edit some issues. Thanks!

jmoenig:
I've pulled it and added some slight changes, mostly to appeal to JSHint's quibbles. I'd appreciate if you could look over my changes and let me know if they break anything (I hope not). Thanks!

PR #3013, opened 2022-03-18
add embed mode button

9072997:
This adds a button to the IDE that enters embedded mode without any UI. This is sort of a fuller than full screen option. I care about this because it makes Snap programs more usable as interactive lessons on large touch displays. There is currently no way to exit this mode, since all the UI is hidden.

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -828,6 +828,7 @@ IDE_Morph.prototype.createControlBar = function () {
         settingsButton,
         stageSizeButton,
         appModeButton,
+        embedModeButton,
         steppingButton,
         cloudButton,
         x,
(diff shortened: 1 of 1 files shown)


jmoenig:
interesting, thank you!
Can you tell me a little bit more about what exactly you like about this mode? Is it that the stage is even bigger, or that no other buttons are shown? Not being able to switch back to editing the project - and to save it! - seems like it would scare a lot students (and teachers)...

9072997:
It is both that that stage is larger and that there are no other buttons. I'm looking for a more full-featured (and ideally open-source) alternative to Hitachi's StarBoard software for creating interactive lessons. Here is an example:

I'm a second grade teacher who has been tasked with creating an interactive whiteboard lesson about money. I create a new Snap! project and change the stage size to something 16:9, to match the aspect ratio of my board. I create a $1-bill sprite and a sprite for each type of coin (quarter, dime, nickel, penny). Clicking on each sprite splits it up into smaller coins
- bill -> 4 quarters
- quarter -> 2 dimes + 1 nickel
- dime -> 2 nickels
- nickel -> 5 pennies
I use the "when I am dropped" trigger to detect any other sprites that I am touching and if their value adds up to a different coin/bill I combine them into that coin/bill. So, you can click a bill and it becomes 4 quarters. You can drag 4 quarters together and they become a bill.

Now I want to actually run this lesson. I want to put it in a full screen format where people clicking on the board won't accidently exit the program. This is where I see the super-full-screen with no buttons being useful.

No, the button overlap is not intentional. If that is your only issue with this, I can try to fix that.

brianharvey:
We have a longstanding policy against unbreakable restricted modes.  There doesn't have to be a visible button, but there has to be a way out -- you click three times fast, or something like that.  You click/touch in the lower left corner.  Something.

PR #3015, opened 2022-03-21
Add Meghan's turtle costumes

brianharvey:
His name is "Tad."

Thanks, Meghan Taylor!


jmoenig:
Uh, I was just going to pull this, then I realized that these files are at around 3 MB each. That's waaaay tooo big for our projects. You import one and your project is already pushing the limit of what you can save. Our other projects are all well below 100 *KB*. Can someone help shrink these?. Also, didn't we agree on SVGs?

brianharvey:
SVGs, unless they're just a wrapper around the bitmap, lose the beautiful watercolor look (done with actual paint and brushes!).  I'll see how much smaller they can get without looking horrible.

jmoenig:
Yay, thanks!

PR #3030, opened 2022-04-07
When bignums on, make numeric slots accept +-*/ and i

WarpedWartWars:
duplicate because I reforked `jmoenig/Snap`


jmoenig:
I think I've already mentioned that I'm not interested in this

brianharvey:
Jens, don't be like that.   You don't like the library, so we're not allowed to improve it?

WarpedWartWars:
I've also already said that if I know what bugs there are, I might be able to fix them.

PR #3033, opened 2022-04-08
Add RESHAPE help

WarpedWartWars:
@brianharvey can you look at this and see if it's ok?


brianharvey:
Yes, it's fine, especially considering how much space each example takes.  But if I were writing it, I'd call out special cases like these:

untitled script pic

untitled script pic(1)
Note that the second one reports a scalar (a number, in this case), not a list.

It might also be worth showing this amazing example:

untitled script pic(2)

untitled script pic(3)

although maybe we should limit the size of the help.  :)

WarpedWartWars:
I'll try to fit the first two, then see if I can fit IDENTITY MATRIX.

WarpedWartWars:
@brianharvey I added the first two, but was unable to fit IDENTITY MATRIX due to size:
reportReshape

brianharvey:
This is the sort of case for which having two background colors is helpful; the two special cases should be on grey.  

I think instead of "This isn't limited to lists" I'd make it affirmative:  "You can use a  simple scalar input (number or text string)."  (Partly because using JOIN with one input is a technique not everyone knows how to do.)

WarpedWartWars:
Grey as in the grey that "Reshapes a list to the provided dimensions" is on?

I'll change that.

WarpedWartWars:
@brianharvey I made some changes, including what you said, but mostly in the .pdn file I have, so I can change it easily, because .pdn supports layers but .png doesn't. I'm still going to use the .png here, though:
reportReshape

brianharvey:
Oh, oops, I misremembered which bg color is where.  Look at the FOR help, for example.  Also I thought I had the empty helps in the repo but they're not there.  Here:
BLANK
BLANKWIDE

I'm really sorry for letting you jump into this without a proper introduction.  It's totally okay if you don't want to change the one you've done (which is just what I wanted, by the way).  But since you have a source file with layers, maybe you can semi-easily change the background.

Thanks again for taking this on!

WarpedWartWars:
It would be much easier if I could figure out how to do a 9-patch in paint.net.

WarpedWartWars:
@brianharvey Is this better?:
reportReshape

brianharvey:
Perfect!  Thanks.

brianharvey (approved):
@jmoenig  Jens, this is ready to pull.

WarpedWartWars:
Not yet! I have to add the file!

WarpedWartWars:
Now this can be pulled.

brianharvey:
Oh...  I thought you'd already done that.  Sorry.  <small>Not my day.</small>

jmoenig:
Thanks!

PR #3036, opened 2022-04-11
add Jahrd to costumes

brianharvey:
Five pngs and edited COSTUMES

(Yes, small files.)


WarpedWartWars:
1.09 MB is not that small...it's a little over a tenth of the project size limit.

brianharvey:
Okay, I made that one smaller.  The rest were already small enough I think.  There's a limit to how small they can be and still look like watercolors (which is what they are, originally).

WarpedWartWars:
Jahrd 2 and Jahrd 6 are also a bit big at 611 and 712 KB, respectively.

Jahrd 1, 3, 5, and now 4 are almost certainly small enough.

brianharvey:
Oh pooh.  It's only a factor of two.

jmoenig:
I'm happy to pull those (beautiful!!) costumes, but I'd be even happier if they were significantly smaller in size...

PR #3039, opened 2022-04-13
Fix #3038

### LINKED ISSUE #3038: is identical too help has a green `false`
The help for `IS IDENTICAL TO` shows the false case with a green false:

image

Seems like it should show it like this:

image

WarpedWartWars:
@brianharvey is this one good?


brianharvey (changes requested):
It's great, except that in the bottom part the text "but it's still the same list" is there twice.  :)

WarpedWartWars:
I fixed that bit.

brianharvey (approved):
Ready to pull, thanks!

jmoenig:
thanks!

PR #3047, opened 2022-05-10
added primitives to get, set and optimize standard Snap Key Maps

himanss:
first contribution, notify me if I messed up in any way.

this commit adds primitives for interacting with Snap!'s standard Key Maps
Example: ``[["key1","val1"],["key2","val2"]]``

these primitives can be created out of Snap! blocks but is not "out of the box" and primitives likely are faster


cymplecy:
This isn't my repository but if I was to wanting to suggest changes and wanted them incorporating into Snap!, I would provide much more information as to the perceived issue (with examples) if I wanted my PR to be looked at :)

It can be a good idea to discuss and propose an issue initially on the Snap! forums

PR #3074, opened 2022-07-22
Update TuneScope library

ego-lay-atman-bay:
I've modified the TuneScope library to add more compatability. I've made it so you don't have to use a note name. You can either user a midi note number, or a note name (C4). Additionally, if you just use a note without an octave, it'll just put it in octave 4 (e.g. `C` will become `C4`). I did make a javascript function to convert note name to note number, but I didn't use that in the `(convert note [] to MIDI)` block, it's done purely in snap (cause I made that first, the javascript function is pretty much directly translated from the snap block).

I did also allow you to use note durations as numbers, or text. I did also update the text durations to make it much more customizable, (now you can do `dotted dotted quarter triplet`). There is a javascript function to also do this, but the snap block is done purely in snap (the javascript function was only added to keep compatability).

I have improved other blocks, making them use less helper functions, scale block now generates the scale instead of reading it from an array (which allows for more types of scales), generalized some blocks, updated the asoc block to be the one in the list utilities library, and more.

These are the blocks I added 

<image src="https://user-images.githubusercontent.com/70973578/180567839-00679d5d-5357-46d0-a132-64f43f9b26bd.png" width="25%" height="auto"></image>

(the hz blocks were only added because there's a play frequency block)

I have also edited the javascript functions to allow the use of midi numbers into the `playNote()` function. The changes in `TuneScope.js` are just changing some dictionaries to functions, and a small fix.

There is an issue I did notice, now this code 
<image src="https://user-images.githubusercontent.com/70973578/180571563-4529252a-5d25-4256-97fd-10ebc981348f.png" width="50%" height="auto"></image>

it plays the wrong octave, and an additional note. If I run this in the current dev version, it just doesn't do anything. I don't know what's causing it.

You can test it out [here](https://ego-lay-atman-bay.github.io/Snap/).


ego-lay-atman-bay:
Hold on, I just realized my dotted note calculations are wrong.

ego-lay-atman-bay:
there, fixed.

watts-j:
Thanks.  Taking a look at this now.  I'll get back ASAP.

ego-lay-atman-bay:
I do want to clean it up a bit (remove an unused function, put in semicolons that I forgot to put in (although, I think it looks better without)). The block library itself is much much more polished than the javascript. I can only do it on Saturday, because I'm busy all week.

ego-lay-atman-bay:
Replying to @watts-j https://github.com/jmoenig/Snap/issues/3072#issuecomment-1197149000

Thanks for the feedback.I can see where you're coming from. Here's why I made most of my decisions.

TL;DR Most of them, were made in the eyes of a programmer, not a music student. I do understand (now) why you made your decisions, and agree with some of them. 

The reason I put in default values, is because, the average joe is not going to know what type of values go into the slot. Having a default value allows the user to just pull out the block, and run it, quickly understanding how the input effects the output.

I do realize now that your note menu is better, mine just looks nice, but it's not exactly that good.

**Play note:**
I'm not exactly sure what you mean by

> replicate your results in Snap. Wherever possible, we'd like to keep the code as transparent as possible and minimize the use of javascript.

because you can already replicate them in snap, by using the note to midi block, if the value is not a number. I only added the javascript functions to get the current implementation of the play track function to work.

**Position of note:**
All you need to do to remove the extra scales, is to remove them in the menu, the code is literally just `item (#) of ((scale type) scale (note))`.

**Interval between notes:**
I'm a little confused on what you said, because my block reports practically the same thing as yours, with the benefit of it not running a say block if the note is not in the scale (it's just really annoying for a reporter to run a say block, if it were a command block, I wouldn't be annoyed). One thing I did differently is, if the pitches are different, it's not considered to be in the scale. One thing I wish I did differently is, the value it reports if a note is not in the scale, should be an empty string, not 0 (to be more in line with other snap blocks).

About the "MIDI 60 major scale", you can put the midi note number in the scale slot, but you can also put in the note name, just like every other note input, you know, for consistency. I agree, MIDI 60 major scale does not sound right, but it makes more sense in programming.

**Play chord:**
The only thing I changed with this block is that it now accepts either a list of notes, or a note name (which is used to get a major chord).

**Diminished scale/chord:**
I may have to agree with you. I couldn't figure out how diminished scales worked, which is why I didn't completely rewrite the code, I just improved it.

**List of major scales:**
I guess you could remove it, although I did remove the need for it entirely, I just left it in because it might be useful to show someone them.

**Major/minor chord:**
I can see that. Understanding how hyperblocks work, can be confusing to someone without an explanation. I'd leave it in as an opportunity  to tech the kids hyperblocks. If you still don't want it, then I'd suggest you set a script variable to the scale, then get the items from the variable, as running the same block, to get the same result 3 times, may not be the best decision (in programming).

**Major/minor chord position:**
In terms of programming, a regular number is better, but I guess in music theory, Roman numerals are better. If you want to use Roman numerals, then at least allow the use of regular numbers. I can hlp with that. All you would need to do is check if the value is a text (not a number), then convert it from Roman numerals to regular numbers. I can try and create a better script to translate Roman numerals to numbers.

**Drum loop:**
Here's a comparison of my block and yours (and a huge difference in using for `each in list`, rather than using `for i = 0 to length of list`)

mine:

8f92ce79-91a9-42e8-bc82-772d8dbd53b4

yours:

1cd0310e-6d0d-4f40-8905-e87ad57107e8

It's also not why my tracks stopped working correctly, that was just on the javascript side of coding. One big reason is, I wasn't even using the drum loop block (because I couldn't figure out how to use it).

**Note from HZ:**
I'm fine you remove it, because it's in the audio comp library. Also, I'm now tinking, it should go in the sounds category. It'll e out of the way there. 

**HZ from note:**
Same as above.

Another thing, the biggest difference between the audio comp version and this version, is that I added the piano keyboard dropdown.

**Major scale:**
You can remove the more complicated scales just by removing them from the list in the block, and the dropdown menu. 

I have also come up with a teaching opportunity for the way I made my script.

The way I did it, it uses the actual step numbers for the scales. For example, a major scale is `2212221` or `whole whole half whole whole whole half` steps. You can use that to teach kids how scales work. Of course, I did make it in a way that is very expansive, so the kids could also create their own scales, and see how it works.

Now, of course, if you want the kids to be looking in the code, you could replace the chromatic scale list to not be a reshape block with a `split [111111111111] by letter` (12 1s) , or a list of 12 1s.

**Beats in time signature:**
To be clear, I was only trying to replicate the results you had in the table, not trying to get it accurate. If I were going for accuracy, it would actually be as simple as using the first number, no math needed. I guess you could rename the block to `quarter notes in time signature` (`beats in time signature` doesn't really make sense, I just don't want the name of the block to be super long, like `quarter notes in measure with time signature [4/4]`).

Every other block you mentioned, that I didn't, I just didn't have anything to say about it. But seriously, thanks for the feedback. I'll see if I can make these changes to the library tomorrow (friday), or Saturday.

ego-lay-atman-bay:
Here's the changes I made.

* I removed the non-standard scales from the scale block.
* Changed the note menus back to the original one. Some menus are just the note names, no octaves attached.
* The scale block now detects what accidental you want, such as, Cb will return the Cb scale with all accidentals flat. Or if you put in B#, it'll return all accidentals in sharp.
* If you put in a list of numbers into the scale type input in the scale block, it'll use that as the scale pattern. I made sure to add a comment on the script that gets the pattern. I'm ok if you remove this, it can be complicated.
* I hid the diminished scale/chord blocks.
* Added "Digital Keyboard" to the `Play MIDI Controller` block, and made it the default (keeping all the other options in the menu, don't want to remove any). I made it the default, because the name seems like a name that multiple midi keyboards would use.
* Renamed `beats in time signature` to `quarter notes in time signature` as that's what it is. I don't know if you want to hide this block.
* I did fix the `interval between` block to work with getting the interval between the start and end of the scale, e.g. C4 to C5 in the C4 scale. I also made that block return nothing if an input note is not in the scale. (well, this was in an earlier commit, but I decided to put it here).
* I hid the `note [] ()` block (the one I added, not the one that was already in the library).
* I made the play note block convert a note name to a note number and put that into the function, all within pure snap.

There's probably more fixes that I can't think of right now, but this is the majority of it.

I do have a suggestion, can you try and figure out a much better way to do tracks, as the current way is not exactly very discoverable, and if it's being added to snap, it should be easy to understand without having to read something, or watch a video. I still don't know how to use drum loops, so that might be

ego-lay-atman-bay:
Don't pull yet, I accidentally merged my snap mod into my tunescope branch. I need to figure out how to reverse it.

ego-lay-atman-bay:
Ok, I've reverted all the changes. I honestly don't know how that happened.

jmoenig:
I'm not going to pull this - at least not anytime soon. I feel that TuneScope is @watts-j 's and their team's project and that I shouldn't mess with any of their code and certainly not with their design. While I enjoy a good design critique and discussion I'm often rather annoyed by unsolicited code-contributions on Github myself. And I can tell from talking to other maintainers of FOSS projects that overly-eager and self-confident code contributors are regarded as more of a problem than an asset in the open-source community. Now I see that @watts-j has been responding in this thread, and therefore I'm going to leave it open for a while. But for the future I strongly suggest to file an issue (as you did, thank you!) and discuss possible ways to address it with the maintainers *there* rather than producing lots of code right away. I hope y'all understand. Thank you!

ego-lay-atman-bay:
Now that I think about it, I see where you're coming at. Thanks for leaving it open for discussion.

ego-lay-atman-bay:
I decided to close this, because I think I should create a pull request on the tunescope repository instead

PR #3078, opened 2022-07-27
Updating Greek Translation

HM100:
It was 2 years since I last added new strings for the Greek language. In 2022, several more have been added, to make it more complete. I tested them before making the PR

All current blocks existed as well as several UI options have been translated into Greek


jmoenig:
How wonderful, thank you so much, @HM100 !

PR #3090, opened 2022-08-01
JSCompiler: cache, runtime implicit parameters...

xBZZZZ:
## runtime implicit parameters
* `implicitParamCount` is now unused
* can remove  part of  block
* work according to [these rules](https://forum.snap.berkeley.edu/t/why-are-ring-parameters-weird/11553#post_3) but they are not same as [snap's weird behaviour](https://forum.snap.berkeley.edu/t/why-are-ring-parameters-weird/11553#post_1)
## very basic function cache
* it's `JSCompiler.prototype.functionCache` array
* is cleared before pushing new function if length &ge; 300
* `JSCompiler.prototype.compileFunction` code makes more sense than my explanation
## :zap: blocks now work with compiled rings (not with any other functions)
* 
* :zap: blocks will error if can't compile instead of synchronously running not compiled rings<br>
* compiled rings have `inputs` property: array of explicit parameter names of original (not compiled ring)
## changed how , , ,  are compiled


jmoenig:
This is very nice, thank you so much!
I love the changes you made to make the implicit parameter count obsolete, and also the introduction of a simple cache. But I think we do need the fallback to simply run a function that is *not* (yet) compilable in the "flash" blocks (instead of throwing a compile-error), because some of our existing libraries occasionally rely on this. Can you make that possible again? Thank you!

xBZZZZ:
> But I think we do need the fallback to simply run a function that is _not_ (yet) compilable in the "flash" blocks (instead of throwing a compile-error), because some of our existing libraries occasionally rely on this.

Maybe make it fallback to behave like non-:zap: blocks instead of synchronously running snap rings.

xBZZZZ:
## added `lst_sort(list, fn) without JS compiling` and `dta_group(list, fn) without JS compiling` extensions
## made :zap: blocks behave like non-:zap: blocks if can't compile.

xBZZZZ:
## about commit https://github.com/jmoenig/Snap/pull/3090/commits/cad3a6e8c8ca3e0523ffbc477d011f866262ce6e
made it return null because  returns `null`

it might look like empty string but

xBZZZZ:
is anyone reading this?
## about commit https://github.com/jmoenig/Snap/pull/3090/commits/ef1032f0cc468b91e19a0b174eae9af22ace8491
 now reports `4` instead of `1`

WarpedWartWars:
> is anyone reading this?

Yes.

xBZZZZ:
## about commits https://github.com/jmoenig/Snap/pull/3090/commits/2de90b3f7054a0595733e54e7663a94b9ac5c206 and https://github.com/jmoenig/Snap/pull/3090/commits/a3116e7593052da6ee0e508a4d3db145b3fc1dc3
this now works:

xBZZZZ:
cache is broken because [ (forum link)](https://forum.snap.berkeley.edu/t/empty-ring-empty-ring-with-parameter/12151/4)

PR #3094, opened 2022-08-04
Improve lookups in sw.js, more clear extensions docs

cycomachead:
* In sw.js this always tries to fetch, then falls back to the cache.
* The Snap! version number is now included in the cache name

* Add docs/ folder to collect the various documentation files
* Add Extensions.md which is largely the comment content of extensions.js


jmoenig:
Thanks, Michael, this is very useful, I'll pull it asap!

PR #3101, opened 2022-08-08
fix dev mode right click `pic...` not working

xBZZZZ:
`window.open` with `data:` URIs doesn't work in modern browsers: https://www.bleepingcomputer.com/news/security/firefox-will-block-navigational-data-uris-as-part-of-an-anti-phishing-feature/

made `pic...` popup look like this:

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -4510,7 +4510,27 @@ Morph.prototype.developersMenu = function () {
     );
     menu.addItem(
         "pic...",
-        () => window.open(this.fullImage().toDataURL()),
+        () => {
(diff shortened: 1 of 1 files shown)


ego-lay-atman-bay:
I'm just hoping this'll finally work in chrome, so I can actually use it.

xBZZZZ:
> I'm just hoping this'll finally work in chrome, so I can actually use it.

it doesn't work in chrome?

xBZZZZ:
works in chrome for me in windows 7 virtual machine:

jmoenig:
Very nice, @xBZZZZ, thank you so much! Will pull asap.

PR #3102, opened 2022-08-09
hyperized bitwise operators

bromagosa:
(no hurries!)

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -264,50 +264,58 @@ SnapExtensions.primitives.set(
 
 SnapExtensions.primitives.set(
     'bit_and(a, b)',
-    function (a, b) {
-        return a & b;
+    function (a, b, proc) {
+        return proc.hyperDyadic(((a, b) => a & b), a, b);
     }
 );
 
 SnapExtensions.primitives.set(
     'bit_or(a, b)',
-    function (a, b) {
-        return a | b;
+    function (a, b, proc) {
+        return proc.hyperDyadic(((a, b) => a | b), a, b);
     }
 );
 
 SnapExtensions.primitives.set(
     'bit_xor(a, b)',
-    function (a, b) {
-        return a ^ b;
+    function (a, b, proc) {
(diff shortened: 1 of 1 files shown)


brianharvey:
While we're at it, could we change the names to something comprehensible?  We're a block language; users don't have to type the names.  But they do have to read them.

The traditional Logo names, for what it's worth, are BITAND, BITOR, BITXOR, BITNOT, LSHIFT, ASHIFT.  But I think we should spell them out even further:
BITWISE _ AND _
BITWISE _ OR _
BITWISE _ XOR _
BITWISE NOT _
LOGICAL SHIFT _ BY _ BITS
ARITHMETIC SHIFT _ BY _ BITS

bromagosa:
After thinking more about it, you're right. I'm always complaining about people naming their functions `f` and their variables `v`. I'm applying it to myself this time :)

jmoenig:
nice, thank you! Just remember that with this change you cannot simply import the new library version into projects that already use the old one, and automatically update them. Instead now you need to manually relabel each old block to its new version.

brianharvey:
Could the library maybe include the old names but with "in palette" unchecked?

jmoenig:
sure, but that won't help with renaming them. No, we just need to be aware that renaming custom blocks in a library does not rename custom blocks in a project. We *could* use translations for that, but that would be really ugly and hard to maintain. No, look, I love the name-change, but just want you to be aware that changing a block label makes a new block.

bromagosa:
Yep, I'm okay with that. I don't think too many people were using it :)

brianharvey:
I guess I don't understand the situation you're thinking about.  What I'm imagining is a project that uses the old names, into which you want to import the new library.

jmoenig:
Yes, we do understand the situation. You'll need to manually relabel the old blocks into their new versions (using the context menu "relabel" option), and then remove the unused old blocks afterwards. We cannot do it automatically. Sorry.

jmoenig:
Will pull later today, thanks, Bernat!

PR #3103, opened 2022-08-09
Add payload to broadcast API

zakkolar:
This updates IDE_Morph.prototype.broadcast with the ability to take a payload to be passed as data alongside the message.


jmoenig:
very nice, I love it, thanks Zak! Will pull asap.

PR #3117, opened 2022-09-19
Enabled Panning in TuneScope via balance parameter, edited blocks to remove Initialize TuneScope in block definitions

harshpadhye:
We edited the TuneScope block definitions to remove the Initialize Tunescope block from their definitions. In the future, we would like to run this block when the user loads in the library. We also edited the `Play Tone` block to enable panning by passing the native `balance` parameter to the `settone` primitive.


jmoenig:
excellent, thank you! I'd like to pull this asap for the next patch - probably coming within the nest 2 weeks (or even earlier, if everything goes well).

Is this ready to pull or do you need anything else from me? How does the initialization work now that you've taken out the init block?

jmoenig:
again pinging this to @harshpadhye, can you tell me whether this is ready to be pulled and deployed to production, or does it need further feedback from me? thanks!

harshpadhye:
@jmoenig Hi Jens, sorry for the late reply. This change is ready to be pulled. 

As far as the initialization block goes, the user will need to click the block at the very start of the project (as opposed to each and every block calling it in their respective block definitions). I couldn't figure out a way to have the Initialize TuneScope block appear on the script stage when the library is loaded (since I could only export the blocks themselves and not the project), but as long as the user clicks the block at the start of their project, everything will work as planned.

harshpadhye:
Just added the last bit of code for the PR, if that get approved the PR is complete and ready to merge. Excited to see the updates in the next patch!

libraries/TuneScope/TS_init.js
```diff
@@ -16,6 +16,8 @@ for (var i = 0; i <= 127; i++) {
     tempMidiFreqs[note] = 440 * Math.pow(2, (i - 69)/12)
 }
 
+window.currentNote = ""
```
jmoenig, line 19: Sigh. I'm not really happy with extensions spamming the global namespace, and neither can you. Remember that these changes will stick around even after the user makes a new project that doesn't use TuneScope or opens another project in the same session. Extensions should strive to be ... extensions and at least attempt some kind of modularity. Actually, keeping track of state is the *one* use case that makes sense for the "make variables" library. You could initialize a TuneScope "_current note" global variable (notice the underscore so it wouldn't conflict with anything the user might create - and just use that inside your custom block. But I realize that other extensions also suffer from this anti-pattern of just spamming the global namespace with their variable, hence my "sigh" - so I guess it's okay if you do so, too. I was hoping - perhaps unreasonably so - that you might be smarter than some of the others. I'm still going to pull this, but I'd like you to consider daring to become more of a systems architect than a duck-tape hacker ;)

libraries/TuneScope/TuneScope.js
```diff
@@ -26,6 +26,13 @@ SnapExtensions.primitives.set(
     }
 );
 
+SnapExtensions.primitives.set(
+    'ts_getcurrentnote()',
+    function () {
+        return window.currentNote
```
jmoenig, line 32: I'm puzzled by this. What happens if you play a chord? Or polyphonic music? Which note is going to be the "current" one? Seems like there can only be one *global* "current" note and whichever gets triggered last "wins" regardless of their duration. Frankly, this doesn't make sense to me. In a polyphonic / harmonic setting asking for a "current note" doesn't make sense at all. I would leave it up to the user script to keep track of this. I'm curious why you need a "current note" and what use cases you are envisioning.

jmoenig:
thanks, @harshpadhye . I'm ready to pull this later this week. I just want to make double sure that you've thoroughly tested it and are aware of all the consequences of e.g. changing the initialization routine. Why are you diverging from your previous - successful - lazy initialization pattern? Wasn't that very user-friendly, because TuneScope programmers did *not* have to worry about explicitly making sure to run an initialization block before trying anything else? Did you consider which error message to show in case users forget to first run the init block, and are you sure an error message is the first thing you'll want your users to see? Are you aware that if users update their existing projects with the new library version - which you'll encourage them to do, right? - their projects will no longer work, because the init block needs to be fiddled in at the right place into the right script? These are just questions that spring to my mind, and it's perfectly okay to me if you've discussed that in your team and made those decisions, I just want to make double sure you did actually make those conscientious decisions and are ready to explain them to your users.
Please let me know if I should really pull this and deploy it to production. Thanks!

libraries/TuneScope/TuneScope.js
watts-j: @jmoenig We completely agree about abuse of the global name space.  …  We need to synchronize animation on the Snap_!_ stage with sequences of musical notes. We had a “Current Note” block that provided the status of notes in the first track. This allowed us to trigger events on the stage with the timing of the music. We used the first track as the default for synchronization because that seemed as good a choice as any.

The students have seemed content to use melody tracks rather than chord tracks for this purpose. If there were a use case, this variable could conceivably provide either the note [C4] or the list [ [C4] [E4] [G4] ] in the case of a chord.  

We had some discussions about potential stripping out the list of durations in the first track, and using these to time the durations of the notes asynchronously. Latencies in delivery of the notes could cause this to drift, but we could look into it if this were judged desirable in order to avoid namespace problems.

Possibly there is a more elegant way to determine the timing of the notes and synchronize them with events on the Snap! stage, but so far we have not thought of another way to do this.

jmoenig:
@harshpadhye i've pulled and integrate this now. But I wanted to let you know that I won't be releasing this patch just yet, because I'll be traveling with limited online access for the next 2.5 weeks (vacation in Istanbul, then a teacher PD in Munich and a conference talk in Barcelona, phew). Since the current v8.0.0 release that's in production right now is rock-solid stable I'm reluctant to deploy a patch that might break stuff while I'm not around to fix it right away. I hope that's okay with you and in the interest of your ongoing course. It also leaves you a little more time in case you want to change anything. You can already test the latest version in [the dev url](https://snap.berkeley.edu/versions/dev/snap.html) - note that you might have to flush your browser cache for the changes to take effect.
Cheers!

PR #3129, opened 2022-10-19
Extension buttons

bromagosa:


jmoenig:
thanks, Bernat will pull later today!

brianharvey:
It'd be nice also to be able to add buttons on the taskbar.  Just sayin'.

bromagosa:
Which is exactly what this mechanism is intended for: adding more locations for buttons when we need them :)

brianharvey:
Oh, the comments said "add buttons to the palette" so I'm envisioning that that's like the "Make a block" button, rather than like the green flag button or like the settings menu button.

bromagosa:
That's what it does, yep, but the design takes in account that we may want to add buttons to other places later on.

PR #3140, opened 2022-11-07
Catalan translation updated

jguille2:
Hi Jens!

Catalan translation updated.


jmoenig:
Yay, thank you, Joan!

PR #3147, opened 2022-12-04
Blocks: Implement for each (item) (index) in list block

CodeLikeCrazE:
This commit adds

* the for each (item) (index) in list block
* the code for said block
* the help page for the block

Please note that translations are needed.

This block is useful enough that I believe it should be in the main version of Snap :^)


jmoenig:
There's a version of ´ForEach´ exposing list indices as upvar in the list utilities library.

PR #3159, opened 2023-01-10
faster Costume.prototype.canvasBoundingBox

SArpnt:
previously past the corners of the costume content bounding box, all pixels would be checked twice (once for horizontal, once for vertical)
a fully transparent costume would get scanned 4 times
there was also a lot of overhead from function calls and some minor overhead from unnecessary operations

this pr optimizes it pretty well, although it does use a labelled break and i don't know how good browser support is on that
i also find the code more readable because it doesn't mix row/col with x/y

### DIFF
--- a/src/objects.js
+++ b/src/objects.js
@@ -11089,62 +11089,54 @@ Costume.prototype.shrinkWrap = function () {
 
 Costume.prototype.canvasBoundingBox = function (pic) {
     // answer the rectangle surrounding my contents' non-transparent pixels
-    var row,
-        col,
+    var x,
+        y,
         w = pic.width,
         h = pic.height,
-        ctx = pic.getContext('2d'),
-        dta = ctx.getImageData(0, 0, w, h);
+        dta = pic.getContext('2d').getImageData(0, 0, w, h).data;
 
-    function getAlpha(x, y) {
-        return dta.data[((y * w * 4) + (x * 4)) + 3];
-    }
+    // the stage is usually wider than tall, but costumes usually around square
+    // left/right is done first because it usually reduces the amount of pixels checked twice
+    // (not by a lot, it would be less than half the height of the stage)
 
-    function getLeft() {
-        for (col = 0; col < w; col += 1) {
-            for (row = 0; row < h; row += 1) {
-                if (getAlpha(col, row)) {
-                    return col;
+    var left, top, right, bottom;
+
+    getLeft: {
+        for (x=0; x<w; x+=1)
+            for (y=0; y<h; y+=1)
+                if (dta[(x + y * w) * 4 + 3]) { // get alpha
+                    left = x;
+                    break getLeft;
                 }
-            }
-        }
-        return 0;
+        // all the pixels are transparent since the loop went through all of them, this will create a 1x1 costume
+        return new Rectangle(0, 0, 1, 1);
     }
+    // note that since at least ONE pixel isn't transparent, all of the other loops must finish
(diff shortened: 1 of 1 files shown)


cymplecy:
Just a general comment :)
It's always a good idea to have a discussion about issues (I'm presuming speed of something in this case?) before just pushing an un-solicited PR request to anyone's repository

Apologies if a discussion has taken place somewhere prior to this PR :)

SArpnt:
i didn't have a discussion beforehand but i don't see any reason why a common operation *shouldn't* be made faster
the main thing this affects that i care about is the pen trails block, since this runs every time using it
many projects take pen trails many times per frame so this makes a significant difference

cymplecy:
"but i don't see any reason why a common operation shouldn't be made faster"
Your right, there are very few times where making something go faster is not a good result :)

But, imagine you have a limited amount of time to work on a project and someone just suddenly pushes changes in front of you.

Without, any qualifying information, how are they to know that the solution actually works and doesn't have un-intended consequences?

Should they adjust their priorities by stopping their own development to test out whether the new code is better and worth the risk of integrating into the main code base

Imagine if a lot of people made un-solicited PRs to improve the parts of a project that we wanted improving ahead of other improvements. :)

So, its always good to have a proper discussion, with statistics to back up claims of speed improvements in real-life projects etc etc

And then ask if a PR is wanted

This applies to almost all open software projects and the maintainers of those projects

jmoenig:
Okay, triggered by [your nudge in the Snap! forum](https://forum.snap.berkeley.edu/t/morph-functions-library-patched/17909/56) I've taken some time to mentally work through your propsed PR, and I like it and will pull it for a quick patch later today. Thank you!

I'll probably reformulate it slightly to better match my preferred (code reading) style, but that shouldn't affect the optimization, which - if I understand it correctly - is to exclude regions that have been tested negative in previous scans, and to short-circuit an early-out in case the first scan doesn't hit any pay dirt at all. Did I understand that right?

I'd like to point out that @cymplecy is right with his remarks, and that it would help me much more if you'd care to be a little more explicit in explaining the ideas behind your PR and how exactly you plan to address them. Yes, I can see now that you've tried to do that, albeit perhaps somewhat too concisely, and perhaps also assuming me to be much more "fluent" in the innards of the Canvas API. Don't underestimate how long it takes me to dig through and to understand such a PR, let alone to think about its consequences down the line. I see in your code that you, yourself aren't entirely sure that you've caught every edge case. See, what makes you assume that I'm any better at that? I'll pull it now and do some quick tests and then release it in a quick patch to production later today, hoping for the best.

I have a pretty large code base to maintain (basically alone) and there are a bazillion smart kids and what I assume to be adult hackers suggesting little optimizations (often not on Github but in mails and messages) that most of the time break something down the line which they didn't anticipate, which is why I'm usually *very* reluctant and slow to even dig through unsolicited code contributions. It's a big project with hundreds of thousands of users around the world, and if something breaks while I'm asleep or on vacation, kids can't turn in their assignments for their AP courses (that has happened) and that gets them (not me) into serious trouble. That's why for me stability and maintainability is way more important than special cased optimizations. Therefore, please submit PRs *very* sparingly and, as @cymplecy mentioned, it's often better to open an issue first so we can discuss it beforehand. Again, I'm in this basically alone and I've got a ton of stuff on my plate to develop and maintain in production, and there are a lot of very weird folks out there in the hacker corners of the internet :)

That said, thank you!

PR #3181, opened 2023-02-16
Update mqttExtension.js

cymplecy:
When returning text to Snap!, restore explicitly making payload into a string. 

Also restore cymplecy.uk instead of simplesi.cloud


jmoenig:
Thanks, Simon! Will pull asap (either later today or tomorrow) and push to dev. Next minor release (8.2) will probably happen around Feb 27/28.

Issue #3183, opened 2023-02-19
Unable to report length of very long text


ToonTalk:
![untitled script pic](https://user-images.githubusercontent.com/2138696/219933890-48538f84-5a82-4d31-b7cf-ca36e12ce067.png)

`too long` was created originally as the encoding of 56 costumes but here I just dragged a 224MB file into 8.1.6

cymplecy:
What would like it do instead if the length is too long?

ToonTalk:
How can a string that is the value of a variable be "too long"? 

This works
![untitled script pic (83)](https://user-images.githubusercontent.com/2138696/220083640-3e691099-566b-431e-b5aa-de3a31d2db28.png)

jmoenig:
because you're probably not giving it a string, Ken, but a list of strings, in which case Snap! is trying to hyperize it. Is that possible?

ToonTalk:
I just imported a TXT file

ToonTalk:
![untitled script pic (84)](https://user-images.githubusercontent.com/2138696/220085625-c2007bca-f394-483f-9eb6-8537632ef116.png)

ToonTalk:
I can click on the variable and it displays some of its value:

![untitled script pic (85)](https://user-images.githubusercontent.com/2138696/220086350-97c509df-6d52-41d3-8dc8-fc5af3545d1c.png)

ToonTalk:
![untitled script pic (86)](https://user-images.githubusercontent.com/2138696/220086815-30ae21ef-56e1-449b-8463-f77212210cff.png)

jmoenig:
Ah, now I think I remember: We're not using the `length` property of the string. Instead we're creating an Array from the string and then take *its* length. The reason for this, if memory serves, is to better support non-Western languages with multi-byte characters. That change was introduced by @cycomachead a while ago.
Of course, this can bite you when working with "big" data, because it essentially copies the whole big string temporarily, and - lamentably - because modern browsers, especially Chrome, totally lack any reasonable memory management and instead decide to simply crash or fail in strange ways (man, I want my Squeak back!). As a general remedy I guess working with "big" data requires you to use a data base web server at some time.

jmoenig:
I really don't know how important that change from @cycomachead is, and whether anybody in Asia actually depends on it. I'd be perfectly fine to go back to just using the `String.length` property, if it were just for me. Any thoughts?

ToonTalk:
String.length seems to do the right thing:

[...]

PR #3188, opened 2023-02-25
Fixing letterOf and positionOf to use Arrays as lengthOf does

jguille2:
Hi!

It is related to #3183 

If finally we keep current "Arrays operation" with "length of string" block, this PR fixes "letter of" and "position of" to be coherent with it.

- "Letter of" is simple, because it's only adding that "Array.from(string)" pre-transformation.
- "Position of" (from "strings" library) has more changes... We don't need any "JS primitive", but we have to consider our "blocks cases" to build a coherent operation (if a field is empty...)

There are more blocks affected, from "strings" and "word-sentence" libraries... but all based on "letterOf" and "positionOf" blocks

Because here (github PR) custom block changes are not easily shown, I paste that block here:

positionOf


cycomachead:
Thanks! This LGTM. We should have done this a while ago. :)

aelnosu (approved):

PR #3209, opened 2023-04-05
Greek Translation Update

HM100:
More Updates have been done to the localization since last update in 30 July 2022.


jmoenig:
Thank you so much! I'm pulling this now. We're currently in a somewhat long-ish phase of development, so it might take several weeks - perhaps even months - until the next release. Just so you know. But this will be in dev all along.

PR #3211, opened 2023-04-12
minor fix: use new URL format in Open in Community Site

bromagosa:
Snap! still uses the old cloud project URL, which is still supported by the cloud for backwards compatibility, but we should really be using the new one. This fixes it.

### DIFF
--- a/src/cloud.js
+++ b/src/cloud.js
@@ -1122,7 +1122,7 @@ Cloud.prototype.removeEditorFromCollection = function (
 
 Cloud.prototype.showProjectPath = function (username, projectname) {
     return '/project?' + this.encodeDict({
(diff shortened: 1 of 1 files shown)


brianharvey:
What does the new format look like?  (I probably have to update the manual.)  Tnx.

bromagosa:
It's the community site URLs that have changed, not the editor's.

The new one looks like:

`snap.berkeley.edu/project?username=[username]&projectname=[projectname]`

Whereas the old ones looked like:

`snap.berkeley.edu/project?user=[username]&project=[projectname]`

In any case, the project page accepts both URLs, and will redirect to the new ones automatically.

PR #3214, opened 2023-05-06
Allow #open to accept images

cycomachead:
You can now pass a URL to #open to import an image.
If the image includes embedded metadata, then this is opened as new project would be.

Alterantively, images without metadata should appear as a new costume.

Also, includes some documentation enhamcements. The README file is designed primarily for GitHub, so that visiting the docs/ folder is more clear.

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -468,7 +468,6 @@ IDE_Morph.prototype.openIn = function (world) {
 
         if (location.hash.substr(0, 6) === '#open:') {
             hash = location.hash.substr(6);
-            console.log(hash);
             if (hash.charAt(0) === '%'
                     || hash.search(/\%(?:[0-9a-f]{2})/i) > -1) {
                 hash = decodeURIComponent(hash);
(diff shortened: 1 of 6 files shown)


src/gui.js
```diff
@@ -468,6 +468,7 @@ IDE_Morph.prototype.openIn = function (world) {
 
         if (location.hash.substr(0, 6) === '#open:') {
             hash = location.hash.substr(6);
+            console.log(hash);
```
jmoenig, line 471: looks good to me, Michael! Can you perhaps take out this `console.log` before I pull it? Tnx!

cycomachead: oops! I forgot to push a couple more changes.

src/gui.js
```diff
@@ -6262,11 +6289,17 @@ IDE_Morph.prototype.rawOpenScriptString = function (str) {
...
     this.spriteBar.tabBar.tabTo('scripts');
-    script.pickUp(world);
-    world.hand.grabOrigin = {
-        origin: this.palette,
-        position: this.palette.center()
-    };
+    if (silently) {
+        this.currentSprite.scripts.add(script);
+        this.currentSprite.scripts.cleanUp();
+
```
cycomachead, line 6295:
```suggestion
```

PR #3221, opened 2023-05-16
Show a custom login notice when students log in for the first time

cycomachead:


brianharvey:
I don't see the actual message text.  Is it somewhere?  Or, would you like me to write one?

jmoenig:
it's in the server response

brianharvey:
Oh.  I think that "learner," while generally a better name for people than "student," is confusing in this particular context, and also "someone else" is too broad and paranoia-inducing.  Why do you want to dance around the fact that student accounts are for schools and teachers?  Here's how I would do it:

---

Welcome to Snap!.

Please read all of this message; you'll only see it this once.

<b>You are using a student account.</b>  That's a special kind of account controlled by your teacher rather than yourself.  "Controlled by" means that your teacher can see your private projects, change your password, and even delete projects of yours.  This student account may also be prevented from publishing projects online, commenting on other people's online projects, or participating in the Snap! forum.  Depending on where you live or go to school, these restrictions may be legally required in order for you to be able to use Snap! in school.

For all these reasons, <b>we encourage you to have your own personal Snap! account separate from this student account.</b>  If you don't already have one, you can create one at home or at the public library, using a non-school email address, by visiting https://snap.berkeley.edu .  Your personal account is under your own control and is unrestricted.

[Note that the s.b.e URL shouldn't be an active link, because it does no good for them to visit it from school in this context.  That's why I put a space before the final period of that sentence.]

[next paragraph in smaller type]

Snap! is inspired by Scratch, and we are greatly in debt to their work, but it is a separate language with much more advanced capabilities than Scratch, because they are interested in building a creative community of mostly young kids, but we are interested in teaching serious computer science to teens.  Don't dismiss Snap! as trivial before you even try it, because you think you've outgrown Scratch.

[button text]

I understand and agree to the above.

---

I'm sure this could be edited, and I could live without the Scratch paragraph if the rest of you don't like it.  I said "a non-school email address" rather than "your personal email address" because in some cases the kid might have to use their parent's address.  Of course the bit about the library is there because some kids don't have computers at home.

The point of the button text isn't to try to establish any sort of contract between us and the kid.  It's entirely to discourage kids from just pushing the button without reading the text, which is what many of them would do if the button said "OK."

Even though you haven't yet implemented the restrictions on publishing, using the forum, etc., we're planning to do so eventually, and then it'll be too late for kids who already have student accounts to read a revised paragraph.  The text is worded to accommodate both the current implementation and a later, more restrictive version.

jmoenig:
Sigh. We are going to show this message to learners in 46 languages on all kinds of devices ranging from big TV screens to tiny phones. Please, please, please let's just show a *very* short notice in simple English and not a political essay.

cycomachead:
Let's have this discussion elsewhere...

PR #3229, opened 2023-06-30
add Aleassa costumes

brianharvey:
Unlike the others so far from Meghan, Aleassa is a girl.  :)

Also added her to the credits window in gui.js.


jmoenig:
Thanks, this is super awesome, I *love* Aleassa!

PR #3233, opened 2023-07-18
Polish Translation Update

P1neF0rest935:
Sorry for sending you this file just before the release of the new version of Snap!. I've been working on it for a long time.


jmoenig:
Thank you! I'm sorry I cannot pull it in time for the v9 release anymore. But we'll work on getting your changes in asap. One problem I noticed in your PR is that is still uses the "old" notation for input slots. We'll probably want to address that before integrating these changes.

P1neF0rest935:
Never mind! It's not actually a full version of the Polish translation - as soon as you pull this file to the official Snap! release, I'll download the source code again and complete the locale file (with the "new" notation for input slots) with these missing strings, then send you via PR.

Regarding the new release: only a few blocks are translated into other languages.

PR #3234, opened 2023-07-21
Updating Catalan translation for v9

jguille2:
Just updating translation


jmoenig:
thanks! will pull and include in the next patch, upcoming either later today or later this week ;)

PR #3238, opened 2023-07-26
Do Not cache requests for #open, other URLs

cycomachead:
Recently, we've had reports of BJC projects showing cached versions, despite the server sending headers.
I am not sure whether the PWA-ness affected these requests, but it's been happening more recently.

I think it makes sense for Snap! not to cache requests so the most recent project/image/etc is loaded.

There was a comment about local robot servers, so this should only affect https requests, since robots and local servers
are almost certainly guaranteed to use plain http.

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -8207,9 +8207,13 @@ IDE_Morph.prototype.getURL = function (url, callback, responseType) {
                 }
             };
         }
-        // cache-control, commented out for now
-        // added for Snap4Arduino but has issues with local robot servers
-        // request.setRequestHeader('Cache-Control', 'max-age=0');
+        // do not cache remote requests.
+        // selectively exclude http requests, which likely mean a localhost/robot server.
+        if (url.match(/^https/)) {
+            request.setRequestHeader('Cache-Control', 'no-cache, no-store, max-age=0');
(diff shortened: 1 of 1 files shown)


src/gui.js
```diff
@@ -8207,9 +8207,13 @@ IDE_Morph.prototype.getURL = function (url, callback, responseType) {
                 }
             };
         }
-        // cache-control, commented out for now
```
jmoenig, line 8210: hmm... so reading the comment it appears that we tried this approach before but that it messed with Snap4Arduino and local robot servers. I dunno, @jguille2 can you and @cycomachead  decide whether this is safe to pull? If so I'll be happy to. Thanks!

cycomachead: I don't know what the issues were with robot servers, but the line `if (url.match(/^https/)) {` should present this from being applied to them.

jmoenig: Ah, good point, Michael. Thanks!

jguille2:
Those changes were fine for Snap4Arduino (in fact, it was an issue from a Snap4Arduino user, with problems with the cached responses from servers)...
But you had to go back because some BirdBrain users. Those fixes were not compatible with their server implementation.

jmoenig:
Oh, thanks, Joan! So we probably should check with Tom about these changes before we pull them...

PR #3242, opened 2023-07-31
Mqtt extension update

cymplecy:
add an extra option to MQTT response blocks


jmoenig:
Thanks, Simon! I've just pulled this and patched the installed version with it. You may need to flush your browser cache to get your new library version.

PR #3267, opened 2023-11-03
Translations/es

jedi-force:
Spanish translations in the local file and also added some translations for the word-sentences library.


jmoenig:
Thanks! will pull this later this weekend / early next week

PR #3274, opened 2023-11-18
Hyphen

brianharvey:
Fix hyphen in word-sentence.xml

### DIFF
--- a/libraries/word-sentence.xml
+++ b/libraries/word-sentence.xml
@@ -1,1 +1,0 @@
-word_sentence.xml
--- a/libraries/word-sentence.xml
+++ b/libraries/word-sentence.xml
@@ -1,0 +1,1 @@
+word_sentence.xml


cymplecy:
Jens - it may be me that's mixed up but I think we need to keep the file needs to be named word-sentence.xml and not sentence.xml

Apologies if its me that's got mixed up :)

cymplecy:
Sorry about rubbish post -edited now :)

jmoenig:
thanks, Simon. I was confused about the alias file which @brianharvey added, and decided to instead rename the actual one. Like to keep the files in a simple, understandable format. It *should* be okay, now, right?

cymplecy:
At this moment in time - the LIBARIES file says word-sentence and the file is called word-sentence so I think we are good :)

I think some housekeeping in dev may be in order to standardize other library names to use dash and not underscore

brianharvey:
I did it that way in the hope of not breaking anyone's still-cached old versions, but I guess that isn't really an issue.

jmoenig:
Thanks, Brian, that explains it!

ego-lay-atman-bay:
Now there's an issue where the words sentances library is a project file instead of a library file.

jmoenig:
Ah, rats! Fixing... Thanks!

PR #3276, opened 2023-11-20
Spanish translation update

dhmartin:
Many fixes, added line breaks in many existing hints, symbol names, etc.


jmoenig:
thank you so much! I'm pulling this shortly to be included in the upcoming patch later today.

PR #3277, opened 2023-11-21
modified:   locale/lang-zh_CN.js

moodykeke:
modified locale/lang-zh_CN.js


jmoenig:
thank you so much! I'm pulling this shortly to be included in the upcoming patch later today.

PR #3279, opened 2023-11-23
Copy / paste scripts

ego-lay-atman-bay:
This adds the ability to copy / paste scripts and comments. It works by setting a value to the block xml, which it then loads as if the user dropped a file into the script area. This makes sure that blocks are properly pasted, so if a custom block that doesn't exist in that scene or sprite, it creates the custom block. The only issue with this implementation is that when a user copies a custom block, then changes the custom block and then pastes, the custom block will be replaced by the old definition.

If you right click a block, there will be 3 new options:

- copy all
  - Copy this block and all blocks underneath
- copy block
  - Copy this block only
- cut block
  -  Copy this block, and delete it.

Comments also have new options

- copy comment
  - Copy this comment
- cut comment
  - Cut this comment

When you copy a comment, the width is also copies, so when you paste it, it stays the same width.

When you right click the editor area, there is a new option to paste (it only shows if you have already copied something).

There are also some keyboard shortcuts

- ctrl + c - copy all underneath mouse, or if it's a comment, just copy the comment.
- ctrl + shift + c - copy only this block or comment underneath mouse.
- ctrl + x - cut block or comment underneath the mouse.
- ctrl + v - paste whatever is in the clipboard.

I know this would make editing faster, because you don't have to drag blocks a very long distance now. You can just cut the script, then scroll to where you need the script, and paste (and it helps with accidentally dropping it in a c slot while dragging it up or down).

Again, the only issue with this is that when you copy a custom block, edit it, then paste, the definition will revert back to when you copied the block. I don't know of an elegant way of fixing this, as it would require modifying the `IDE_Morph.prototype.droppedText` function.


cymplecy:
This is a comment that I'd make no matter what the PR was.
Any dev on any project almost NEVER likes unsolicited PRs for many, many reasons too long to go into here 

No matter how good the new feature seems and how well it may or may not have coded it.

ALWAYS, ALWAYS, ALWAYS have a discussion on the forum first :)

I'll open up a forum thread to talk about this

ego-lay-atman-bay:
Ah, thanks. I'm still new to creating pull requests (I haven't made that many). I'm also gonna close this because I forgot that it should be to the dev branch instead of main, and also so we can talk about it more.

brianharvey:
For what it's worth, I think this is a great feature, provided that if there is text selected in a comment, it's still that text that gets cut or copied rather than the whole block.  I haven't reviewed the code.

ego-lay-atman-bay:
I just checked, and apparently it doesn't copy cut or paste blocks while editing text, even ctrl + shift + c, which I had to throw on `event.preventDefault()` to not open the dev tools, doesn't copy a script. I think it's all because when you're editing text, the text area that is actually being focused blocks keyboard events from running in snap. That's just my guess anyway.

PR #3280, opened 2023-11-24
Copy / paste scripts

ego-lay-atman-bay:
This is pretty much the same as my other pull request, but this time it's to the dev branch instead of master, and I did make a couple edits.

This adds the ability to copy / paste scripts and comments. It works by setting a value to the block xml, which it then loads as if the user dropped a file into the script area. This makes sure that blocks are properly pasted, so if a custom block that doesn't exist in that scene or sprite, it creates the custom block. The only issue with this implementation is that when a user copies a custom block, then changes the custom block and then pastes, the custom block will be replaced by the old definition.

If you right click a block, there will be 3 new options:

- copy all
  - Copy this block and all blocks underneath
- copy block
  - Copy this block only
- cut block
  -  Copy this block, and delete it.

Comments also have new options

- copy comment
  - Copy this comment
- cut comment
  - Cut this comment

When you copy a comment, the width is also copies, so when you paste it, it stays the same width.

When you right click the editor area, there is a new option to paste (it only shows if you have already copied something).

There are also some keyboard shortcuts

- ctrl + c - copy all underneath mouse, or if it's a comment, just copy the comment.
- ctrl + shift + c - copy only this block or comment underneath mouse.
- ctrl + x - cut block or comment underneath the mouse.
- ctrl + v - paste whatever is in the clipboard.

I know this would make editing faster, because you don't have to drag blocks a very long distance now. You can just cut the script, then scroll to where you need the script, and paste (and it helps with accidentally dropping it in a c slot while dragging it up or down).

Again, the only issue with this is that when you copy a custom block, edit it, then paste, the definition will revert back to when you copied the block. I don't know of an elegant way of fixing this, as it would require modifying the `IDE_Morph.prototype.droppedText` function.

I am open to talking about this, and modifications. I don't think it would be best to merge this as it is currently, because I'm not sure if the function names should be renamed.


brianharvey:
Good point about naming; "Copy all" should be "Copy below"

Yeah, _really_ it should be called "copy block and below" but I think people will understand it because of what happens if you drag out of a script.

cycomachead:
I would like to play with this but I am not sure we want to offer a “cut” context menu — at least not yet. We still get far too many complaints about accidental deletions. Copy and paste is great though and this will definitely help folks. -- Michael BallFrom my iPhonemichaelball.coOn Nov 24, 2023, at 8:42 PM, Brian Harvey ***@***.***> wrote:﻿
Good point about naming; "Copy all" should be "Copy below"
Yeah, really it should be called "copy block and below" but I think people will understand it because of what happens if you drag out of a script.

ego-lay-atman-bay:
You can play around with it in [my snap mod](ego-lay-atman-bay.github.io/Snap) (as well as other additions I made).

I have also looked at the files that were changed in the pr, and it seems like there are things that I didn't change, but are different from the dev branch. I'm gonna have to go in and clean that up so it's only the copy and paste code.

PR #3304, opened 2024-02-14
3D beetle fixes

bromagosa:
Fix convex shape extrusions, and remove intermediate extrusion faces.

Volumes like this one can now be properly constructed without any geometry errors, and printer slicers will take it:

imatge


jmoenig:
yay!!

PR #3310, opened 2024-02-18
Dropdown library addition

bruh-9000:
Create a clickable dropdown menu which runs a broadcast with the attached data of which option was clicked.


bruh-9000:
Note that this does require a costume in the sprite using it to function, and the size of said costume is how far away from the arrow the dropdown can be opened from.

ego-lay-atman-bay:
I'm gonna say this now, all the libraries you've been creating PRs for, would probably work better if you just posted them on the forum.

bruh-9000:
@ego-lay-atman-bay Yes, I posted them onto the forum last year, but I don't see a reason not not to add them to the Snap! library now that I found out there is a GitHub.

ego-lay-atman-bay:
I can think of a few reasons that jens may not want to add them. The biggest reason is that the libraries are held on a higher level of quality that also need to fit an actual usecase for. While this dropdown library is good, it's kind of redundant, as there is already a menus library.

bruh-9000:
Hm, I was not aware of the menus library. I will close this PR but I think the other PRs have good use cases.

PR #3314, opened 2024-02-22
Remove dependency from global world var in api.js

bromagosa:
Move window.onmessage to IDE_Morph constructor so it doesn't need a global world var

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -346,6 +346,23 @@ IDE_Morph.prototype.init = function (config) {
 IDE_Morph.prototype.openIn = function (world) {
     var hash, myself = this;
 
+    window.onmessage = function (event) {
+        // make the API accessible from outside an iframe
+        var ide = world.children[0];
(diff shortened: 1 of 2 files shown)


src/gui.js
```diff
@@ -346,6 +346,23 @@ IDE_Morph.prototype.init = function (config) {
 IDE_Morph.prototype.openIn = function (world) {
     var hash, myself = this;
 
+    window.onmessage = function (event) {
+        // make the API accessible from outside an iframe
+        var ide = world.children[0];
```
cycomachead, line 351: Maybe you just forgot to push something, but doesn't this still use the global? 🙃 


```suggestion
        var ide = myself;
```

bromagosa: haha, well, this `world` is the parameter you're passing to `openIn`, not a global var, but you're still right that `myself` is the right thing to do :)

PR #3316, opened 2024-02-22
Zak's MicroWorlds cleaned up and conflict-free :)

bromagosa:


brianharvey (commented):
> enable JS extensions by default

Is this right?

ego-lay-atman-bay:
> > enable JS extensions by default
> 
> Is this right?

It looks like that was reverted.

jmoenig:
yep, this PR looks alright, we've been discussing it over a bunch of recent meetings and want to get the EDC stuff into the Snap extensions system, mainly to keep it alive. This is the first batch of extensions. I'm not totally sure whether we'll want to actually expose them - all - to endusers, because then these features will end up in regular projects and will need to be supported by us.

PR #3338, opened 2024-04-15
Spring '24 update Dutch translation

NykkiRusticus:
I'm a student of Joek van Montfort, and as part of my learnings I learned GitHub, while adding hundreds of translations


jmoenig:
Oh wow, this is super awesome and just in time, too! This morning @jadga-h and I were discussing the state of the Dutch translation and whom to ask for updates, and lo and behold: You've already updated all the strings we were talking about, and then even many more. THANK YOU!
I'll pull this asap within the next days and publish it in this week's patch (and let you know here). Yay!!

jmoenig:
Again, thanks, this is now online!

PR #3353, opened 2024-05-11
Improvements to Libraries + Media Browsing for Search

cycomachead:
This isn't a dramatic change, but a meaningful quality of life improvement.

* Adds a search field to the library import dialog (see screenshot)
* replaced the kind of odd tab-delimited file with a JSON file for all media loading
* includes hooks for categories and other search data, mostly for libraries, but could later be used for costumes, etc.

## Libraries Dialog
Screenshot 2024-05-11 at 4 00 30 PM

## Search
Search acts similarly to the project search. It's a simple substring search. 

* library name (translated)
* library description (translated)
* block categories (I think are translated?)
* block specs (not yet translated)

### Block Specs / Updating Script

To get the block specs I ran `ruby utilities/add_blocks_to_libraries.rb`
This parses the blocks from each library's XML file

* It collects the categories
* Skips over 'helper' blocks, since they aren't exposed to users
* Normalizes the block specs, e.g. it replaces `$arrowRight` with "to"
* Script can replace data in the JSON file OR append to it. This may not be granular enough in the future, but that's OK

### Translations for blockspecs
Right now, I haven't addressed this. The vast majority of library blocks have no translations. 
What's tricky is that in the current schema, I assume you are either using the english translation or the current language, but they aren't 'merged'.

If only some block specs are translated, then we should likely 'merge' the English + translated specs. But this was kind of annoying to think about... 

Additionally, the custom translations of symbols are only in English right now.

## Media file schema. 
This is documented in code, but it's like this. Everything other than `name` and `fileName` is optional.

```json
{
  "fileName": "",
  "name": "",
  "description": "",
  "categories": [],
  "searchData": [],
  "translations": {
     "de": { ... },
     "es": { ... },
     ...
  }
}
``` 

This moves the translations from each individual language file to the media file. I've copied over what's currently translated, but not yet removed the original translations. 

IMO, this will make it easier to incrementally translate names and descriptions. If we do update the English name or description, we no longer break the translation. 

I'm not expecting the categories array to be translated, since those are likely to live in the individual translations files already (in the case of blocks) OR could be added for costumes instead of repeating translations many times.

There's no searchData or categories for any of costumes, backgrounds or sounds, but obviously there's been interest in adding that data before...


locale/lang-es.js
```diff
@@ -178,14 +178,14 @@ SnapTranslator.dict.es = {
...
     // translations meta information
-    'language_name':
-        'Espa\u00F1ol', // the name as it should appear in the language menu
-    'language_translator':
-        'V\u00EDctor Manuel Muratalla Morales / Cristi\u00E1n Rizzi Iribarren / Alfonso Ruzafa / David Mart\u00EDn', // your name for the Translators tab
-    'translator_e-mail':
-        'victor.muratalla@yahoo.com / rizzi.cristian@gmail.com', // optional
-    'last_changed':
-        '2023-11-20', // this, too, will appear in the Translators tab
+    "language_name":
```
cycomachead, line 181: I'm not sure these even belong in the current format of the translation file, but using `"` made it easier to parse the file and extract the translations.

ego-lay-atman-bay:
For the block specs, couldn't you do what the search box in the block pallet does?

PR #3354, opened 2024-05-12
Libraries Search: Update sw.js / Examples file

### LINKED ISSUE #3326: [bug] Drop reporter into color input after it is taken out
In the color input type, you can drop a reporter into the color input, but after you take the reporter out, it becomes static. This only happens with a custom block with a color input, primitives with a color input are fully static.
image
image

This happens in both dev and 9.2.10, most notably, the color reporter in the colors and crayons library.

cycomachead:
- updates EXAMPLES to EXAMPLES.json
- caches new JSON files in sw.js


jmoenig:
awesome, thanks, Michael!

jmoenig:
is it actually updating sw.js though?

PR #3361, opened 2024-05-24
dont show warning onbeforeunload if no unsaved changes

2-www:


jmoenig:
thanks for the PR. Could you perhaps issue it against the dev branch, that would make it easier for me to review it.

So, it seems that currently it *is* again possible to do this. In the past this has been notoriously unreliable and not consistent across browsers (https://stackoverflow.com/questions/38879742/is-it-possible-to-display-a-custom-message-in-the-beforeunload-popup) even as recent as a few months ago (https://chriscoyier.net/2023/11/15/careful-about-chrome-119-and-beforeunload-event-listeners/), that's why I'm rather reluctant to add some complicated mechanism to widely used production code, since i'll have to support and maintain it indefinitely. But I'll happily look into it. Again, thanks!

jmoenig:
looks perfect now, thank you!

PR #3363, opened 2024-06-06
Microblocks generic library

bromagosa:
John and I unified our MicroBlocks libraries so now you can connect to a MicroBlocks device either over BLE or serial port and it will work just the same.


jmoenig:
Awesome, thank you! Will pull tomorrow.

PR #3364, opened 2024-06-19
Update colors.xml

brianharvey:
Fix bug reported in forum


jmoenig:
thanks, Brian. This should go into v10 dev, right? Can you change the pull request so it doesn't go into the main branch? Thanks!

brianharvey:
No, this should go in the current (v9) Snap!.  It's a bug fix.

jmoenig:
oh, okay, thanks.
Hmm... given that v10 is rather imminent I'm not sure if we're going to publish another patch vor v9. So you might want to issue this PR for v10 dev anyway to get it into the beta release channel. Otherwise I'll probably wait until I merge dev into main and then pull it for the v10 (rc) release, which might take some more time. Dunno, up to you.

PR #3376, opened 2024-07-12
Update MQTT for V10

cymplecy:
New PR against master not dev
Update mqtt.xml and mqttExtension.js to use lovely new variadic options that are  in V10 

Includes Maximum QoS for subscribe block by Xavier Pi as well


cymplecy:
I think I've not upload properly so closing this PR and will remake it
Sorry for noise

PR #3382, opened 2024-07-28
Update embedMetadataPNG@morphic.js fix long script

DarDoro:
Fix the embedding of large >65kB scripts (XML). 
The spread (...) operator may cause stack overflow (Chrome@Win10). Replaced with the "slice/concat" sequence.
The problem is described on the Snap forum. https://forum.snap.berkeley.edu/t/some-smart-script-pics-arent-working/17230

### DIFF
--- a/src/morphic.js
+++ b/src/morphic.js
@@ -1631,7 +1631,7 @@ function embedMetadataPNG(aCanvas, aString) {
             embedTag
         );
     try {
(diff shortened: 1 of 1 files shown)


jmoenig:
Thank you, @DarDoro ! I'll pull this for the next patch, which - because we're currently in the throes of SnapShot - might occur only later next month.

PR #3391, opened 2024-08-25
Cat10.0.5

jguille2:
Hi Jens.
This PR is to update Catalan translation with the new features, aligned to current 10.0.5
I have to translate some libraries (pending to other PR) but this PR has all the IDE pending translations. And I want to point you a "little" problem I've seen testing the translation (I hadn't seen that with normal use!)

I saw that the commit [you did updating all locales to the new system](https://github.com/jmoenig/Snap/commit/5391780728444fc93f052bce28d230e7b77701c9) deleted all the "\n" from all the strings (including the originals English/coding strings). So, a lot of strings are not translated now.
Fortunately, most of them are warnings, help text... which are not translated into many languages ​​(not like with Catalan where I did have to spend a long time)

Catalan locale is fine now. I've been checking original strings with the code to find spaces, newlines... deleted. Maybe the solution is using Catalan locale as the template now. Then, I will check deeper, trying to also detect the strings that are already obsolete (because the code is no longer there but they are still there). Tell me what you think about this... and I can continue working on this.

Thanks!
Joan


jmoenig:
Thank you so much, Joan! I'll pull this later for this week's patch!

jguille2:
Thanks Jens.
I will try to find time to translate the new libraries (Embroidery, Beetleblocks...) as well. But it will be another PR.

And we will talk about the problem with the missing "\n" in locales. You can see the problem by looking at the help text in the settings menu using Deutsch. You have translated some of these strings, but now it's only English.
As I said, I can spend time making these corrections, checking all the locales.
Joan

jmoenig:
Hi Joan, yes I'm aware of the line-break issue, we'll deal with it later. I'm not entirely happy that we currently need to enhance the translation file with specific entries for libraries such as embroidery etc., but again at this time I don't have a better idea. Thanks!

jmoenig:
again thank you, Joan! I've pulled, released and deployed your update.

PR #3406, opened 2024-10-13
merge

YeesterPlus:


jmoenig:
Is this a ... mistake?

ego-lay-atman-bay:
I feel like this should just be closed, because [looking at it](https://astofappblox.github.io/snap.html\), it looks like a customized snap mod (which kind of looks good), which I can tell would not be merged.

PR #3419, opened 2024-11-27
Update lang-ja_HIRA.js

fumiyasu-aizu:
Support expansion of [if] and [is identical to] blocks.


jmoenig:
thank you, will pull for this week's patch!

PR #3423, opened 2024-12-10
Fix "delete block _" block in spanish translation

Bubgamer07:
The delete block _ had no inputs in the spanish translation


jmoenig:
good catch, thanks!

PR #3427, opened 2024-12-28
Cat10.3.5

jguille2:
Hi Jens!

I'm updating Catalan translation with the latest changes... and also translating some pending libraries.

Note1.
In the third commit, translating Beetle lib, I've added to the locale different new strings that are not related to those blocks (the new Beelte view and their elements). I point to this because maybe you want to pick those strings to the Deutsch locale (as our template).

Note2.
Beetle blocks have some menus that I can't translate directly. I know I can do it applying to their options something like  "circle=$_circle" and so, the translation is done. But then I have to fix beetle.js to use this programming strings into its code.
I'll send a proposal in other PR and then, Bernat and you will check it... to see if this is the right way.

Thanks!!

Joan


jguille2:
Hi Jens.
Finally I've done those changes (to translate the options of the menus of the beetle lib) directly here because I see I don't need to change nothing in beetle.js

What I did is in the next picture, using "$_string" to get the translations from the locales, and using the primitive txt_transform('unselect',string) to return the original English word to the beetle.js script.
optionsofBeetleLib

jmoenig:
Hi Joan,
this is fabulous, thank you so much! I'll pull this asap for the next patch (I'm working on a few minor fixes and hope to finalize a little patch sometime next week - or the week after that).

PR #3434, opened 2025-01-19
Create Sci-Snap localization file: SciSnap3Blocks_hy.xml

antrohoosclubs:
Please create a localization file for Armenian translation. Antrohoos Foundation will translate it fully.


jmoenig:
Hmm... we might have to look at this together. Maybe it makes sense to include the Armenian translation in the SciSnap file instead of exchanging the file...

antrohoosclubs:
Sure, let's add the translation to the main .xml file. I've already translated many sections, but it's challenging to work directly in the XML due to its difficult structure.

ego-lay-atman-bay:
Couldn't you add the translations through the gui?

antrohoosclubs:
No, I cannot add the translations through the GUI because I didn’t find the GUI editor available for that purpose.

jmoenig:
Okay, this file seems to conflict with SciSnap's latest changes (a fix from Eckart Modrow for the file reader block). I am going to leave this for now, we can pull it and integrate it in another patch when we figure out what to do about it. Do you need help with finding out how to translate the SciSnap blocks to Armenian? Thanks!

PR #3437, opened 2025-01-23
Updating Greek Translation (Again)

HM100:
Its been time since the Greek translation was updated. Its now the right time to update it.


HM100:
Should any issues arise, I can fix them

jmoenig:
Oooh, awesome, thank you!

PR #3439, opened 2025-03-03
Update lang-hy.js

antrohoosclubs:
Please change the new Armenian translation made by Antrohoos Educational Foundation. A lot of new words and sentences have been translated


jmoenig:
Ugh. - I've first pulled and merged this request, but then discovered that it breaks the Armenian translation, because there appear to be LTR issues in the file, resulting in unclosed strings in JavaScript. @antrohoosclubs could you please check your translation update once more and verify it with something like JSHint.com for syntactical correctness? That would be wonderful, thanks!

jmoenig:
Update: I've investigated into this issue, and it doesn't seem to be LTR related. Instead your file has a bunch of unescaped double-quotation marks inside the translated strings, which appear to be redundant. I'm editing the translation file and will add it later today. I think this will make your translation update fine for tomorrow's release. Next time please use a Lint to ensure your PR's syntactic validity. Thanks!

PR #3444, opened 2025-03-18
Update mqttExtension.js

cymplecy:
Small fix to clear issue when feeding list output to unicode block

Forum post
https://forum.snap.berkeley.edu/t/strange-unicode-base64-inconsistency/19177/2


jmoenig:
Thanks, will pull later this week - or whenever we'll finalize the next patch.

PR #3449, opened 2025-04-18
Base64 unicode fix

cymplecy:
Improve Unicode string handling in Base64 blocks


jmoenig:
Thanks! Will pull asap for a little patch later today

PR #3451, opened 2025-04-20
Error types and try-catch support for them(feature)

YeesterPlus:
https://forum.snap.berkeley.edu/t/please-add-custom-error-types/19318/11
I'll stop with the typed error pull making here
image

when the error name is "Error", the trycatch block passes the error message instead of both its name and message to the catch code.
when the input to the error block isn't a list, it uses "Error" as the name.
both the trycatch block and the error block uses a list which is: the name and, the message.

js funct blocks won't work when off

### DIFF
--- a/src/extensions.js
+++ b/src/extensions.js
@@ -368,6 +368,11 @@ SnapExtensions.primitives.set(
 SnapExtensions.primitives.set(
     'err_error(msg)',
     function (msg) {
+        if (msg instanceof List){
+            var err = new Error(msg.at(2));
+            err.name = msg.at(1);
+            throw err
+        }
         throw new Error(msg, {cause: 'user'});
     }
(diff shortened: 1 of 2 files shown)


YeesterPlus:
image

YeesterPlus:
image

YeesterPlus:
you can use an error with a name of "Why is there a" when there is a type error

YeesterPlus:
when the error name is "Error" the trycatch block passes the error message instead of both its name and message to the catch code

YeesterPlus:
js funct blocks won't work in some cases

jmoenig:
Again, my aim is not to support kids who play with JavaScript.

YeesterPlus:
ok :/

Issue #3452, opened 2025-04-29
Spanish Block Name Incorrect?


maryfries:
![Image](https://github.com/user-attachments/assets/0155c215-1317-4276-b3b1-e6193bb1a56e)

![Image](https://github.com/user-attachments/assets/9ffd224b-9370-4d55-ac96-6b2b209aa702)

@bromagosa Hi and happy spring! :)

bromagosa:
Hi, Mary! :)

Yep. I've been repeatedly saying that the Spanish translation was changed for the worse at some point. Quite a lot of strings that were correct were changed into funny-sounding incorrect ones.

In any case, this one's blatantly wrong. I've just sent a PR with a fix.

PR #3453, opened 2025-04-30
Fix two wrong strings in ES translation

bromagosa:
As reported by Mary in #3452


jmoenig:
Thanks! Will pull asap for a little patch later today

jmoenig:
oh, wait! It's directed to the dev branch. In this case I'll still pull it, but not for a patch :)

jmoenig:
OTOH, would you mind re-issuing it against main? I'm preparing a little patch anyway, so we can add this, thanks!

PR #3455, opened 2025-05-11
transparent broadcasts for MicroBlocks library

bromagosa:
This makes it so regular Snap! broadcasts are forwarded to MicroBlocks when loading the MicroBlocks library.

It also removes the now obsolete `[ broadcast [ ] to MicroBlocks ]` block.

Thoroughly tested in both the Snap! and MicroBlocks sides.


bromagosa:
superseded by newer PR

PR #3456, opened 2025-05-11
Speed up BLE connection

bromagosa:
John found out this is about 10x faster :)


bromagosa:
superseded by newer PR

PR #3462, opened 2025-05-22
Base64 byte decoding improvement

cymplecy:
Decode base64 string direct to byte list when selected


jmoenig:
Thanks! Will pull this into dev later this morning.

PR #3465, opened 2025-06-04
[FIXED] Add basic quadratics solver library

PPPDUD:
This is a pull request to add the `Quadratics` library for Snap!, written entirely using Snap!'s GUI functionality.

I created the aforementioned library to help solve quadratic equations ([learn more](https://en.wikipedia.org/wiki/Quadratic_equation)) in the form `ax^2+bx+c=0`, using `a`, `b`, and `c` as parameters.

In order to allow for the usage of complex numbers in calculations, the library will automatically setup the complex numbers library as needed.

Some of the blocks provided with this library:
image

_Note: I wrote this library in a slightly modified Snap! 10.2.5, but I've found that it works on the latest unmodified builds as well._


cymplecy:
FYI - most unsolicited PRs are rejected (this applies to a larger number of opensource projects)

Much better to have a discussion on the Snap forum first and then see if your code would be welcomed or not :)
https://forum.snap.berkeley.edu

cymplecy:
I'm trying to avoid you wasting your time - do with my suggestion what you will :)

PR #3480, opened 2025-09-06
Fixing noExitWarning flag/configuration feature

jguille2:
Hi Jens!

I saw noExitWarning flag didn't work
Then, I also saw that current JS onbeforeunload  implementation had different changes. Now, changing the warning message is not possible (not without a custom implementation, disabling the browser feature and creating you own "modal").

So, I fix our current operation in a simple way: using a ` window.noExitWarning`  flag,  checking in our "beforeunload" listener if `(this.hasUnsavedEdits() && !window.noExitWarning)` and using the current allowed instructions there.

Joan

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,7 +4,7 @@
 
 ### 2025-09-06
 * blocks: fixed & added "senders / receivers" scanning support for "request" reporters
-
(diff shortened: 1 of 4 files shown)


jmoenig:
Thanks, Joan, will pull later (today or tomorrow)!

PR #3486, opened 2025-10-10
An Accessible High Contrast Mode

cycomachead:
This is not done!! Not sure when I will finish this, but this is one way to think about meeting WCAG contrast requirements. 

* darken the default palette by 40%
* still use white/black text for blocks because of this.
* Create a new 'design' mode which is largely based off flat design
  * This design sets large fonts
* Create a new theme called 'high contrast' which uses the high contrast colors

TODOs:
* Flat mode buttons look ugly. I tried to add a border (we need to remove the gradient)
* The design mode menu checkboxes are broken
* I think 40% is too much. Maybe 30% or 25% darker?
* With darker blocks, should we use a light mode, or possibly a
  * slightly lighter dark mode background?
* We should darken the font (to be closer to #000) in dialogs and other places
* We should increase the base font size in dialogs and related places by a bit
* The green/red boolean toggle will need adjustment

---
A smattering of screenshots:

Screen Shot 2025-10-09 at 19 42 04
Screenshot 2025-10-09 at 7 33 31 PM
Screenshot 2025-10-09 at 7 33 00 PM
Screenshot 2025-10-09 at 7 32 56 PM
Screen Shot 2025-10-09 at 19 27 04
Screen Shot 2025-10-09 at 19 21 18

Some are from: https://multiple-contrast-checker.netlify.app which I used to generate the color tables. we probably want to verify the hex codes match what are the rgb values in Snap! because I used copilot to convert them but they look close enough. 

Project with all C-shaped blocks to test zebra coloring and combos
[Untitled-2.xml](https://github.com/user-attachments/files/22827069/Untitled-2.xml)


Mark4SISB:
I've been looking at this with grayscale turned on on my computer (found in accessibility settings on windows), and noticed a few things when combining flat design and high contrast.

Text boxes don't have clear outlines:
image
Additionally, gray rings don't have outlines like other blocks:
image
The new scrollbar colouration is great... until you have a white menu with a scrollbar or slider:
image
The outline doesn't properly cover the hump of hat blocks:
image
The white outline of the jagged input types needs some work as well:
image
The category selector outlines are clearly rounded on the left side when they shouldn't be:
image
And empty categories have low contrast (although this may be intentional):
image
Visible stepping looks terrible when enabled (even without greyscale - this is a problem in flat design period):
image
Dropdown menus could use some work as well (maybe an outline around the arrow?):
image

That's everything I noticed in flat high contrast. High contrast on its own looks BEAUTIFUL, so much so that I might switch over when the time comes.

As for "large text", however, I wonder whether the issue you were trying to address has been solved by the new magnification setting. It particularly feels weird to me because the setting doesn't make the block labels any larger and actually makes the blocks feel smaller because of the outline.

PR #3491, opened 2025-11-01
Fix combining empty lists with `all`, `any`, `append`, or `combinations` reporters

### LINKED ISSUE #3324: COMBINE doesn't know APPEND's base case (and others)
When we turned a bunch of former library blocks into primitives, some variadic ones should have been added to COMBINE's list of base cases:
help-test script pic
help-test script pic (1)
Don't forget that in the COMBINATIONS the base case result isn't empty:
help-test script pic (2)

COMBINE also lost the base cases of some formerly dyadic blocks that have become variadic:
help-test script pic (3)
help-test script pic (4)
help-test script pic (5)
help-test script pic (6)

These are all the ones I found but I didn't carefully check every block.

rmunn:
Fixes #3324.

This PR fixes the empty-list base case for the `combine` block for use with the `all` (variadic AND), `any` (variadic OR), `append`, or `combinations` (cross product) reporters. In all cases, the base case will be the same as what that block reports when it is called with zero input slots:

- For `all`, the base case is TRUE, because "do all items of this empty list satisfy condition X?" is true for any condition X.
- For `any`, the base case is FALSE, because "do any items of this empty list satisfy condition X?" is false for any condition X.
- For `append`, the base case is the empty list
- For `combinations`, the base case is a list containing one empty list (i.e., `(())` in Lisp notation)


rmunn:
I realize that the lines I added all run over 78 characters, but given how deeply-indented this check is and how long the names are that we're dealing with, this is unavoidable short of refactoring the existing code. I decided that for a quick bugfix, being obviously correct (by making the smallest possible change that fixes the bug) was better than playing code golf to conform to the 78-character limit.

However, if the 78-character limit is an iron rule, then a refactoring should be possible to bring this under 78 characters on each line. First, split the `list.length() === 1` and `list.length() === 0` checks apart rather than having them both under `list.length() < 2`. Then in the `list.length() === 0` check, we could write `const selector = reporter.expression.selector` and save 20 characters on subsequent lines.

If you want me to edit this PR to refactor the code to keep lines under 78 characters, let me know and I'll be happy to do so. I won't do so yet, though, so that you can review the code in its "obvious" state rather than in the less-obvious state it would be left after refactoring.

src/threads.js
```diff
@@ -3843,8 +3843,11 @@ Process.prototype.reportCombine = function (list, reporter) {
                 this.returnValueToParentContext(
                     list.length() ?
                         list.at(1)
-                        : (reporter.expression.selector === 'reportJoinWords' ?
-                            ''
+                        : (reporter.expression.selector === 'reportJoinWords' ? ''
+                            : reporter.expression.selector === 'reportVariadicAnd' ? true
+                            : reporter.expression.selector === 'reportVariadicOr' ? false
+                            : reporter.expression.selector === 'reportConcatenatedLists' ? new List()
+                            : reporter.expression.selector === 'reportCrossproduct' ? new List([new List()])
```
cycomachead: I haven't tried this, but I don't think this is very readable

```suggestion
           initial_combine_value_for_selector(reporter.expression.selector)
```

```js
let initial_combine_value_for_selector = (selector) => {
  switch(selector) {
  case 'reportJoinWords':
    return '';
  case '':
     return ...
  default:
    return 0;
};
```

rmunn: Yes, now that there are five cases extracting them into a switch/case probably makes more sense. I'll work on that in a few days when I get a chance.

rmunn:
About to rebase on top of current `master` branch to resolve merge conflict. If you've pulled this branch, you'll need to pull again after the rebase since commit IDs will be different.

jmoenig:
Thank you! I'll pull this for the next minor release, which might take a few weeks, so please don't think I'm ignoring this while I'm fidgeting with other stuff, I'm not forgetting you!

rmunn:
I'll hold off on fixing any more merge conflicts in snap.html until your work is done, because rebasing and pushing multiple times would make the PR history complicated. If you want me to rebase it again once you're done with the work in `objects.js`, I'll be happy to, just let me know when to do it.

PR #3505, opened 2025-12-19
Add COSTUME FROM URL and SOUND FROM URL blocks

ego-lay-atman-bay:
This pull request adds two new blocks to the Web services access (https) library to load costumes and sounds from a url. This also adds the `snd_load(url)` extension function, which is used by the SOUND FROM URL block.

Some notes:
- Currently `cst_load(url)` doesn't handle svg images, so if a user enters a url to an svg, it will will return it as a bitmap costume. I think I know how to do this, but the way I'm thinking will depend on what we do for the next point.
- When a costume is loaded from an external website that doesn't allow CORS, the image will load fine, and you can mostly use it normally, but you can't get the pixels. The project also won't be able to be saved if one of these "tainted" costumes is being used. Would it be best to just not allow the use of these "tainted" images at all?

Semi-related: I noticed that when trying to drag a costume from a report bubble with a name of `null` (the default name), it throws an error in the console.


ego-lay-atman-bay:
I found out that the way I was doing it before would make all external costumes be tainted, which isn't ideal, so it really is best to handle cors. It also now reports an svg costume if the external resource is an svg image.

PR #3528, opened 2026-04-28
Updating Catalan translation for 12beta260428

jguille2:
Hi Jens!

Just updating Catalan translation with the latest changes.

- One comment
I've updated dates and "HISTORY", trying  to save you routine work... but maybe you'd prefer to manage these changes yourself... Let me know so I can do it right next time.

- One suggestion
Now we have the "Looks" section inside the Settings Menu. We can move the "Performer mode" selection to this section.

Thanks... and see you really soon :)

Joan


jmoenig:
this is all wonderful, Joan, thank you so much! I'll pull this right now.

The "Looks" menu only has settings that can be included in templates and projects, not (yet) temporary settings like turbo mode and performer mode. You're right that performer mode is also a "look". Maybe there should also be a way to mark a project to always directly open in performer mode? Let's talk about this next week!

jguille2:
Thanks Jens.

Yes, my suggestion was  just considering the setting "type".

But now I see a more important sorting consideration. Regardless of how the "performer mode" setting is (whether it remains temporary or can be saved in the project), it would be fine to separate the three types of settings in the "Settings menu": those that are saved with the IDE (they do not go with the project but are maintained when opening another one, like language), those that are saved with the project and the totally temporary ones.

This distribution would have an important meaning. Inside these three blocks (IDE, project, temp) we can consider (as a second step) a sorting by topic.

Example of this. The "single palette" setting is stored inside projects. And yes, maybe it is also a "look"... but more important than this (the topic), is that I want to explain (and to know) that it is a project setting.

If you want, we can "play" with this the next week :)

PR #3536, opened 2026-05-20
New Keyboard Menu

ego-lay-atman-bay:
I never really liked how in scratch and snap, the menu for selecting keys was just a huge dropdown list, so I decided to redesign it to be a full keyboard layout.

https://github.com/user-attachments/assets/3c804f24-8a24-4f7c-a2ca-780e443fb456

When you open it, you can select all the keys you could before, but with a better ui. When you hold shift, there are also many keys that change to their shifted counterparts, so for example, 4 would become $ when holding shift. I also added a really nice quality of life feature that just detects whatever key you press and automatically selects that one, so you don't even have to use the ui (it also allows for selecting keys not in the menu), but pressing esc will close the menu instead of selecting it. I also made sure that everything is localized, or at least uses the localized method. The only issue right now is that there's no way to toggle shift with the ui.

Just to be clear, when holding shift, it doesn't make all the letters uppercase because snap doesn't detect the key when it's uppercase (possibly a bug).

Since this new menu is very large, I decided to make it fall back to the old menu on smaller screens (if the menu width is larger than the world width).


jmoenig:
Oh wow, this is pretty, and pretty cool!
I'd love to come back to this after Snap! v12 has been released and after we've ironed out its expected baby glitches. So, please give me some time (I'll be on vacation after the launch until mid-/end July.
Does the virtual keyboard layout work with different translation settings, too? Otherwise we'd have to think about that.
Thank you!

codingisfun2831t:
Now I'm not related to Snap!'s development at all; but maybe making each button like.. smaller? Would be better? So like more like the piano menu. Like the old System Software's keycaps:
Screen Shot 2026-06-02 at 12 43 30

ego-lay-atman-bay:
Sorry I didn't get to this sooner, I have no idea how I missed it

> Does the virtual keyboard layout work with different translation settings, too? Otherwise we'd have to think about that.

The layout itself doesn't change, but that could be added. It does translate words like "space" if that's what you were asking.

> Now I'm not related to Snap!'s development at all; but maybe making each button like.. smaller? Would be better? So like more like the piano menu. Like the old System Software's keycaps: Screen Shot 2026-06-02 at 12 43 30

Oh, that looks very nice, and I think that would be better.

codingisfun2831t:
Would I be able to try to make a proof-of-concept of that more compact UI? Should just be some size/style changing.

ego-lay-atman-bay:
Yeah, go ahead

codingisfun2831t:
I was trying to and it looks well, not the best:
image

I'm going to see if I can kind of do a bit of a hardcodey method (just call a addKey or such a bunch of times, should work out fine)

codingisfun2831t:
@ego-lay-atman-bay Hows this:
image

codingisfun2831t:
I got an invitation - do you want me to commit it there (I already kind of made a whole branch on something else for a PR...)

ego-lay-atman-bay:
Yeah, I was thinking it might be best to contribute to the same PR

codingisfun2831t:
Commited! Hoping I didn't do anything wrong..

Mark4SISB:
Just took a look at this. I like the new compact design, but ideally the borders between keys would be half-width, or the keys would be closer together, so that the borders are equal in thickness whether on the edges or between keys.

codingisfun2831t:
> Just took a look at this. I like the new compact design, but ideally the borders between keys would be half-width, or the keys would be closer together, so that the borders are equal in thickness whether on the edges or between keys.

Think I shouldve fixed this, i made each key subtract one from x after placing. Have yet to test this though, would you be able to pull down the new code and test it?

Mark4SISB:
The borders look much better!

But... umm... heh heh. The right border and up arrow key are messed up:

image

codingisfun2831t:
Should be fully fixed now, i just needed to update the spacing code to also respect the borders

Mark4SISB:
Still not quite.

image

ego-lay-atman-bay:
One issue I see is when translated to other languages. Some words may go outside the key bounds. The space key should be fine because it's so wide, but any, backspace, tab, and enter won't. Backspace and enter should be fine to just increase the width to compensate, and any can just be moved to the bottom row to also be able to increase, but tab may be an issue. I'm thinking we can just remove tab because I think it's a more niche key to want, and it's also still possible to select it just by pressing tab.

Mark4SISB:
I can conceive of some kind of autoscaling to maintain the rest of the dimensions, but removing tab wouldn't be the worst thing in the world by any stretch. I really like being able to just type any key and make it go straight in though.

PR #3538, opened 2026-05-23
Final Chinese Translations Update(for Snap! 12)

Ayist14:
Because of **The Second Mock Junior High School Academic Proficiency Test in Liupanshui(六盘水市第二次初中学业水平适应性考试)** taken in next Monday and Tuesday, I only added translations for 4 libraries , fixed up translations for the new settings' menu, and completed the basic notes for Chinese scratchers in few settings by 2 PRs. Now I must prepare for the test and **Zhongkao** in the last month, so the next PR will be opened after the **Zhongkao(from June 20, 2026 to June 22, 2026)**.


jmoenig:
Excellent, thank you!
I just noticed another activity here and wanted to ask: Is this ready to be pulled now? If so I'd integrate it right now. I also noticed that you changed the dropdown menu items in the pixels library, prefixing them with "$_ " to make them translatable. That's great! But we need to make sure to leave out the blank space character between the "$_" and the rest of the word that's to be translated. Don't worry, I can take care of that once I've pulled it (just in case you're wondering why a translation might not yet "work").
Thank you!

Ayist14:
@jmoenig Sorry for the late reply! Because I always live in my mother’s former workmate’s home (it’s near my school) in school days, I haven't had time to reply to your question and advice these days.

> I just noticed another activity here and wanted to ask: Is this ready to be pulled now? 

No, I think this PR I opened is not complete but almost enough for me, so it’s not pulled now.

>But we need to make sure to leave out the blank space character between the "$_" and the rest of the word that's to be translated. Don't worry, I can take care of that once I've pulled it (just in case you're wondering why a translation might not yet "work").

Thank you for your suggestion! I agree with the change you mentioned.

And I noticed that Snap<em>!</em> 12 was released yesterday. What a nice and convenient version! I hope Snap<em>!</em> can become better and better!

PR #3539, opened 2026-05-23
Add compressed files to the gitignore

cycomachead:
This ignores gzip and brotli compressed files if present in the repo. 

There's currently no need for these files to occur naturally while working on Snap! itself. However, I will deploy a change to the snapCloud server that stores compressed files on the server. 
We already serve pre-compressed assets when loading snap, but we are wasting CPU cycles on constantly recompressing files. This will just keep server checkouts cleaner. :)


cycomachead:
@jmoenig Just FYI going to merge this right now for easier testing on the server, but there's no functional changes here. :)

PR #3542, opened 2026-05-31
added noShare config flag

bromagosa:
When present, it hides share and publish buttons from project list dialog.


jmoenig:
Thank you, Bernat, will pull asap...

PR #3544, opened 2026-05-31
Fix potential undefined reference to window.parent in TuneScope.js

MatrixNeoKozak:
The TuneScope.js initialization code (TS_init.js) defines several functions and objects intended for cross-frame communication via window.parent, but TuneScope.js directly accesses window.parent without checking if the script is running in an iframe. When loaded directly (not in an iframe), window.parent references the current window, but subsequent accesses like window.parent.midiPitches assume a parent context that may not exist. This change adds a guard in TuneScope.js to ensure window.parent is defined and accessible before accessing its properties, preventing runtime errors when the script runs outside of an iframe.

### DIFF
--- a/libraries/TuneScope/TuneScope.js
+++ b/libraries/TuneScope/TuneScope.js
@@ -1,7 +1,9 @@
 SnapExtensions.primitives.set(
     'ts_setinst(name)',
     function (name) {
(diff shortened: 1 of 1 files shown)


jmoenig:
Thanks! will pull later today and deploy with the v12.0.3 patch within the next couple of days.

PR #3545, opened 2026-06-01
forcing fixLayout to skip loadingByURL problems

jguille2:
Hi Jens.

This is the issue we were talking by mail.

The origin is [a commit on February 18](https://github.com/jmoenig/Snap/ commit/519a8b5d17b50bf1aa00f80f84b3923d2b01829b). It seems right, because the `hideSpritePanes` is already fixing the Layout... but you know asynchronous processes they give us surprises :)

I think this fix is not doing nothing wrong... and I've tested and it fix the problem I see in some scenarios (like the Moodle plugin loader).

Joan

### DIFF
--- a/src/gui.js
+++ b/src/gui.js
@@ -7291,7 +7291,8 @@ IDE_Morph.prototype.switchToScene = function (
     scene.applyGlobalSettings();
     this.selectSprite(this.scene.currentSprite, true);
     this.corral.album.updateSelection();
-    this.hideSpritePanes(this.scene.hideSprites); // this.fixLayout();
+    this.hideSpritePanes(this.scene.hideSprites);
+    this.fixLayout();
     this.corral.album.contents.children.forEach(function (morph) {
         if (morph.state) {
             morph.scrollIntoView();


src/gui.js
```diff
@@ -7291,7 +7291,8 @@ IDE_Morph.prototype.switchToScene = function (
     scene.applyGlobalSettings();
     this.selectSprite(this.scene.currentSprite, true);
     this.corral.album.updateSelection();
-    this.hideSpritePanes(this.scene.hideSprites); // this.fixLayout();
+    this.hideSpritePanes(this.scene.hideSprites);
+    this.fixLayout();
```
jmoenig, line 7295: Joan, I don't understand this. hideSpritePanes() already calls fixLayout() (even in both cases, which I'll change right away) as its final statement.

jguille2:
Yes Jens... As I said, I also don't understand some of these asynchronous processes... but I see I need that "fixLayout" because the other processes are deleting the loading project.
I'll mail you with more details and an example to test.

jmoenig:
closing this, as we've found another solution. Thank you, @jguille2 for all your work and testing!

PR #3547, opened 2026-06-05
Point the Reference Manual link at the new website

cycomachead:
Just points to the new website. :)
The manual is still in place and will continue to have links to the PDF that work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 a visual, blocks based programming language
 inspired by Scratch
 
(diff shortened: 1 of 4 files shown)


cycomachead:
@jmoenig Updated this to point to dev.

jmoenig:
Thank you, @cycomachead , will pull asap and deploy with the v12.0.3 patch within the next days!

PR #3554, opened 2026-07-13
Skip cloud session init at startup when config.noCloud is set

gigamonkey:
Maybe not a huge issue but I was getting errors in the console on my embedded Snap. Claude helped me diagnose what was going on and suggested this patch which I've tested on my site and seems to work.


jmoenig:
this looks straightforward and fine to me. Will pull for the next patch - once I'm done with stuff in the dev branch. Thanks, Peter!

PR #3555, opened 2026-07-13
Restore palette after getting project XML.

gigamonkey:
Another one Claude helped me debug. In my embedded Snap I have it set up to autosave and after autosaving the palette search box no longer worked. This patch seems to fix that. There may be a better place to do this so that both paths that need this restoration get it but this is the simple change that fixes the immediate bug.


jmoenig:
This also looks perfectly fine to me, thank you!

PR #3563, opened 2026-08-15
Add 'clean unhatted' scripts-pane menu item

gigamonkey:
This is a feature I've wanted for a long time: like 'clean up', but it first deletes all loose blocks, i.e. blocks not attached to some hat. Deletions go through the normal drop-record path so each one is undoable, and through IDE_Morph.removeBlock so any processes running those scripts are stopped. I'm not tied to the name. `clean loose` might be more clear in some sense? And maybe easier to translate into other languages?

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -9224,6 +9224,11 @@ ScriptsMorph.prototype.userMenu = function () {
     }
 
     menu.addItem('clean up', 'cleanUp', 'arrange scripts\nvertically');
+    menu.addItem(
+        'clean unhatted',
+        'cleanUnhatted',
+        'arrange scripts vertically,\nremoving all scripts\nnot attached to a hat block'
+    );
     menu.addItem('add comment', 'addComment');
     menu.addItem(
         'scripts pic...',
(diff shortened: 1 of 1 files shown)


gigamonkey:
@jmoenig Do you want PRs against `master` or `dev`? `docs/CONTRIBUTING.md` says `master` but I feel like I saw you ask someone to submit against `dev` and so that's what I've been doing.

jmoenig:
Hi @gigamonkey , thank you both for the PR and for notifying me about updating the contribution guide! Your're right that nowadays I prefer PRs issued against dev, because that lets me work on longer termed features while leaving the main branch ready to fix urgent bugs that come up in schools etc. I've just changed the doc accordingly.

I like your feature, and I especially like that you've taken great care to really integrate it well into the environment. I'm sorry that Snap's codebase is somehow not standard web programming stuff, but this weird homebrew GUI. That has historical reasons, because when I started to work on Snap way back when the DOM wasn't really usable with JS for things like this. Back then I was discussing these issues with my colleagues Dan Ingalls, John Maloney and Evelyn Eastmond and decided to try a minimal implementation of our Morphic system from Self / Squeak. Alas, it's not "minimal" anymore, but at least it's enabled me to single-handedly keep developing and supporting Snap! for over 15 years now.

I'm a little worried that - even though your feature nicely supports undo - students in schools might accidentally loose code and panic by checking out your "clean-unhatted" feature. To my great frustration Snap! is nowadays used more in 5th to 7th grade - especially over here in Germany - and not so much in College, so those "power-user" features that all college students and - well - power users want are exactly the ones that many school teachers want us to disable for their kids.

Would it be okay for me to pull this, and then to make it a "hidden" feature? It would be such that the menu entry only shows up when you hold down the shift-key when popping up the context menu, like the entry for clearing the undrop-queue. Would you be okay with that?

gigamonkey:
That sounds fine to me. I"m an Emacs user; extra modifier keys can't scare me! ;-)

PR #3564, opened 2026-08-17
Slight improvement to click-to-edit UI

gigamonkey:
It's always frustrated me that there's no way to shift-click and type a block name to add a new block at the bottom of a set of blocks. This PR fixes that.

Shift-clicking a command block's lower half now places the blinking keyboard-entry cursor below the block (between it and any next block) instead of always above it. Shift-clicking the scripting pane's background close beneath the end of a script (within the drag-snap threshold, horizontally within the bottom block) attaches the cursor to that script's end instead of creating a free-floating one.

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -6080,13 +6080,13 @@ BlockMorph.prototype.pickUp = function (wrrld) {
 
 // BlockMorph events
 
-BlockMorph.prototype.mouseClickLeft = function () {
+BlockMorph.prototype.mouseClickLeft = function (pos) {
     var top = this.topBlock(),
         receiver = top.scriptTarget(),
         shiftClicked = this.world().currentKey === 16,
         stage;
     if (shiftClicked && !this.isTemplate) {
-        return this.selectForEdit().focus(); // enable copy-on-edit
+        return this.selectForEdit().focus(pos); // enable copy-on-edit
     }
     if (top instanceof PrototypeHatBlockMorph) {
         return; // top.mouseClickLeft();
@@ -6099,7 +6099,9 @@ BlockMorph.prototype.mouseClickLeft = function () {
     }
 };
 
-BlockMorph.prototype.focus = function () {
+BlockMorph.prototype.focus = function (pos) {
+    // if the click position is given and lies in my lower half,
+    // place the focus below me rather than above
     var scripts = this.parentThatIsA(ScriptsMorph),
         world = this.world(),
         focus;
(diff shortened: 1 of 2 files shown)


cycomachead:
I like this feature in principle but I do not think we can “split” a block like this — the rule is a little too undiscoverable and I think the click targets are too small — especially at 1x zoom level. IMO, it would be reasonable to always place the cursor immediately beneath or around the selected block. -- Michael BallFrom my iPhonehttps://mball.coContinuing Lecturer UC Berkeley EECS | Travel Advisor Fora On Aug 17, 2026, at 9:57 AM, Peter Seibel ***@***.***> wrote:﻿It's always frustrated me that there's no way to shift-click and type a block name to add a new block at the bottom of a set of blocks. This PR fixes that.
Shift-clicking a command block's lower half now places the blinking keyboard-entry cursor below the block (between it and any next block) instead of always above it. Shift-clicking the scripting pane's background close beneath the end of a script (within the drag-snap threshold, horizontally within the bottom block) attaches the cursor to that script's end instead of creating a free-floating one.



You can view, comment on, or merge this pull request online at:
  https://github.com/jmoenig/Snap/pull/3564

Commit Summary

  8e3a29b  keyboard entry: attach focus to script end on shift-click near a block's bottom


File Changes  (1 file)

  
    M
    src/blocks.js
    (51)
  


Patch Links:

  https://github.com/jmoenig/Snap/pull/3564.patch
  https://github.com/jmoenig/Snap/pull/3564.diff

gigamonkey:
I've deployed it on my server and it actually feels very natural. Prior to this PR if you click right at the top of a block it puts the cursor above the block (because it puts the cursor at the top of the block no matter where in the block you click.) But if you move your mouse one pixel up and click now the cursor jumps all the way to the top of the preceding block because you are now in the next block.

What this change does is make it so everywhere you might want to insert a block has a one-block high click target around it: the bottom half of the block above the insertion point and the top half of the block below it. So effectively the click target is the same size as before (one block) but it puts the insertion point in the middle of that area rather than at the top.

gigamonkey:
I've added a few more changes to fix other glitches with the text entry mode:

1. Now you can click inside a C block to add blocks inside. E.g. click inside a Repeat block to start adding blocks to the loop
2. Clicking near the top of a block adds the cursor at the top of the block so you can click-to-add blocks above as well as below existing blocks.
3. After shift-clicking and typing a name you can click in the palette of matching blocks to select where that used to just put you into a kind of dead state where you couldn't do anything and had to shift-click elsewhere to get out.

gigamonkey:
And the last change allows shift-click in a hole to add a block.

PR #3567, opened 2026-08-30
Dedupe custom block names at creation time

gigamonkey:
Currently when you make a block the name doesn't get uniquiefied until the editor is closed. Which means if the project is somehow saved with the editor still open (which can happen on my website where I've embedded Snap!) then the project is saved with multiple blocks with the same name.

This PR changes it so the name is made unique as soon as the block is created. If there's already a `foo` block and you make a new block and name it `foo` it opens in the editor as `foo (2)` which is how it would have been saved later anyway. So this seems strictly better in that it's obvious that you're making a duplicate block as soon as the editor opens.

### DIFF
--- a/src/blocks.js
+++ b/src/blocks.js
@@ -9276,11 +9276,20 @@ ScriptsMorph.prototype.userMenu = function () {
                 () => new BlockDialogMorph(
                     null,
                     definition => {
+                        var spec = definition.spec,
+                            count = 1;
                         if (definition.spec !== '') {
                             if (definition.isGlobal) {
                                 stage.globalBlocks.push(definition);
                             } else {
                                 obj.customBlocks.push(definition);
+                            }
+                            // make sure the spec is unique
+                            while (
+                                obj.doubleDefinitionsFor(definition).length > 0
+                            ) {
+                                count += 1;
+                                definition.spec = spec + ' (' + count + ')';
                             }
                             ide.flushPaletteCache();
                             ide.refreshPalette();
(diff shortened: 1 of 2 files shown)


Mark4SISB:
You can also close the open window by changing a looks configuration such as "flat design". This will produce multiple "valid" same-name blocks. This PR should fix that as well.

jmoenig:
Thank you! This looks like a good idea. I'll try to pull it (and your other PR) in time for the upcoming minor release. However, as that is approaching very quickly (probably tomorrow or Wednesday) and as I'm currently working feverishly on curriculum materials that require the new release I might not get around to pulling it in time. In that case we'll just make another patch release one or two weeks later. So, if I might appear incommunicado over the next couple of days please don't despair and don't think that I'm rejecting these. I'm just ... busy. Thanks!

jmoenig:
Okay, I'm sorry, but this is not going to make it into the release I'm about to make, there's just too much on my slate right now. But - as I mentioned before - we'll get this pulled and published shortly in the coming couple of weeks.

gigamonkey:
No worries on my end. Since I'm hosting my own Snap! I have all my own patches applied. When I upgrade to the new release I can just keep patching anything that hasn't made it in yet. Thanks!

jmoenig:
That's great to hear, Peter, but I really want these patches in production. Right now we're feverishly working on https://snap.schule to get our curriculum stuff for 7th grade in Baden-Württemberg (state in Germany) out by next week. Today I've finally released v12.1, and we've also mirrored it over at snap.schule (the German GDPR conforming site that's used in schools over here). School starts in less than 2 weeks, and they're introducing this new subject, and there's still so much to do! When the dust settles later in September / October we'll get your contributions in, I promise!

PR #3570, opened 2026-09-07
Glide reporters dropped onto a rejecting input slot back to their starting point

gigamonkey:
Students often make the understandable mistake of trying to drag a variable into the first hole in a `set [] to []` block. Snap! correctly won't let that happen. But it *will* let the variable be dropped on top of the hole so it looks like it's in the hole.

It seems useful to give students some feedback that they've done something impossible rather than make it look like the impossible thing worked. Thus this PR changes things so when you try to drop a reporter (such as a variable) in a hole where it doesn't belong, e.g. the static variable dropdown of 'set var to', an upvar template, a prototype's input in the block editor, or a type-checked slot of the wrong type, when you let go it just glides back to where it came from.


jmoenig:
thank you, Peter!
I agree that some visual feedback in that situation would perhaps help, even though I've personally never experienced any of that in classes or workshops that I've been present at. Maybe that's because all the kids already knew Scratch and were familiar with how variables are accessed in Scratch.

I'm not quite sure whether making the dropped reporter slide *all the way* back to from where it was picked up isn't going to annoy these kids even more. Maybe we make it so it does get dropped, but slightly away from the slot? This - maybe - makes it clear that you can't drop the reporter into that slot, but also doesn't "take away" the reporter from you?

Anyway. I'll be working on another quick patch today, and - just so you know - I'm *not* going to pull any PRs for that one. It's not because I don't like the PRs (I do!), but because I urgently need the quick patch to make https://snap.schule ready for the launch of 7 th grade in Baden-Württemberg, Germany, a week from now.

I'll eventually get around to pulling contributions after that, I promise.

gigamonkey:
Okay, I changed it so it just glides along the path towards its original location until it gets to clear space. (There are a few situations where it goes back to where it came from, such as if you drag a block from the right hand hole of a a `set [] to []` into the left; in that case it ends up back where it started and nothing has changed.

